PR #31, opened 2026-07-04
feat(cli,harness): lib-store bundle build pipeline and CLI pull

krajla:
Implements the two lib-store OpenSpec changes (both archived with specs synced into `cli/openspec/specs/lib-store-provisioning` and `harness/openspec/specs/lib-store-build`).

## cli — `add-lib-store-pull`
- `inflexa libs pull|status|list`: arch-aware bundle resolution (full@amd64, core@arm64 with an explained downgrade), manifest fetch (zod-validated, path-safe version segment), sha256-verified parallel downloads into a content-addressed dedup cache, atomic activation (unique staging dir → rename → `current` symlink swap), one rollback version retained.
- Store mutation is serialized machine-wide via the existing shared instance lock (`lib/lock.ts`, key `lib-store`); `prune` sweeps crash debris under that lock. Deliberately simple concurrency: one pull at a time, everything else follows.
- Setup-wizard offer (size shown, consent required, cancel-tolerant; headless setup prints a hint instead of auto-downloading), TTY-only timeout-bounded pre-launch hint, and a coupling guard so the harness only mounts a healthy store (re-checked at sandbox creation in `docker-client.ts`).
- `packages.txt` is byte-identical to the shell assembler's output — a contract test drives the real `lib-store-pack.sh`/`lib-store-assemble.sh` against `libsPull`.
- Mirror support: manifests carry a store-relative `path` per track, so `INFLEXA_LIB_STORE_URL` redirects payloads, not just manifests.

## harness — `add-lib-store-bundles-and-validation`
- `images/lib-store-builder`: per-track fail-loud load tests (Gate 1: import/library()/require()/tool-invocation), `packages.txt` fragments generated from the *loaded* set, per-track tarballs.
- `scripts/lib-store-*`: shared pack/assemble/write-manifest tooling; `scripts/lib-store-validate/` (Gate 2): validates advertised ⊆ loadable inside the real sandbox image via the real `inflexa libs pull`, plus curated anchors (DESeq2, scran, scanpy+numba, squidpy, ms-deisotope, samtools) and R example scripts.
- Workflows: `lib-store.yml` (arch matrix, immutable versioned publish, candidate pointers) and `validate-lib-store.yml` (Gate 2 + promotion to `latest`). No mutable tags; promotion only after a green Gate 2.

## Validation
- cli: typecheck clean, full suite 440 pass / 1 arch-gated skip locally (one unrelated embedding-GGUF test flakes on timeout under full parallel load; passes in isolation). Lint clean on all touched files.
- harness: `tsc` clean, full suite 684 pass / 1 skip / 0 fail.
- **Note:** cli CI is expected to fail at `bun install` (403 on the private `@inflexa-ai/tsprov`) until that package is public — the local runs above stand in for it. Harness CI should be green.
- Shell scripts `bash -n` clean; workflow YAML parses; `validate.py` py_compile clean; R scripts parse.

## Known limitations (accepted, reviewed)
- Gate 2 requires a real built store + S3; not locally testable — validated by construction and dry parsing only.
- `packages.txt` invariant is directional (advertised ⊆ loadable); extra loadable packages are tolerated.
- Stale-lock takeover uses the shared `lib/lock.ts` semantics (dead-pid reclaim; EPERM treated as alive).
- Blob-cache GC is deferred (`TODO(extend)`): prune reclaims versions, not orphaned blobs.


krajla:
@claude review

krajla:
@claude review

vivere-dally (approved):

PR #34, opened 2026-07-06
Add library-validator smoke-test suite

radutanasa:
Self-contained per-package smoke tests under cli/scripts/lib-validator/, one <package>.R or <package>.py per library, following a common contract: a not-installed guard (prints "FAIL: package '<pkg>' is not installed" and exits 1), an isolated per-test harness, and a PASS/exit-0 vs FAIL/exit-1 summary. Covers CRAN, Bioconductor, and GitHub R packages plus the Python data-science / single-cell / omics stack. Missing packages exercise only the guard; installed ones run real API checks (synthetic data, tolerances).

run_all.py discovers every sibling validator, runs them concurrently with the right interpreter (python3 / Rscript), classifies each outcome (PASS / not-installed / FAIL / ERROR), and prints a grouped report. It exits non-zero only when an installed library is actually broken; absent libraries are skipped.

Also ignore Python __pycache__/ bytecode.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] **Sandbox:** the security model is preserved (default no network egress, working-directory-only mounts, resource limits); any loosening is called out and justified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,10 @@ dist-ssr
 # TypeScript incremental build info
 *.tsbuildinfo
 
+# Python bytecode cache (e.g. from cli/scripts/lib-validator/*.py smoke tests)
+__pycache__/
+*.py[cod]
+
 # Guard against tsc emitting alongside source. the harness builds to harness/dist (ignored
 # above); these catch an accidental in-place emit. Scoped to core, NOT blanket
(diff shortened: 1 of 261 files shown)


krajla:
@radutanasa I cherry picked the validators into my branch, closing this one.

Issue #33, opened 2026-07-06
Adopt a client–server architecture: one harness daemon, many chats & runs


vivere-dally:
# Adopt a client–server architecture: one harness daemon, many chats & runs

## Problem

Only one `inflexa` process per machine can execute work. The harness boot acquires a machine-wide advisory lock (`cli/src/modules/harness/runtime.ts` — `RUNTIME_LOCK_KEY = "harness-runtime"`, acquired via `acquireInstanceLock` from `cli/src/lib/lock.ts`), and a second `inflexa profile` / `inflexa run` fails with:

> Another `inflexa` process (pid …) is already running the harness runtime. Only one harness runtime per machine at a time.

Two user-facing consequences:

1. **No concurrent work.** A user with a capable machine cannot run two analyses (or a profile and a run) at the same time — even on completely different analyses. Chat is unaffected today (the TUI/intelligence module never touches the harness; chat goes straight to the proxy), but every *execution* funnels through the one lock.
2. **The owning terminal is hostage.** The DBOS workflow lives inside the CLI process that booted the runtime, so `inflexa profile`/`run` block until terminal status. Ctrl+C does not detach a live run — it *stalls* it: the workflow only resumes when a future boot adopts it via DBOS recovery. A multi-hour run requires a terminal open for multiple hours.

## Why the lock exists (it is not accidental)

Every process launches DBOS with the fixed `executorId: "local"` (`runtime.ts`, launch config). DBOS launch-time recovery reclaims in-flight workflows **by executor id** — so a second concurrent launch under the same id would adopt and re-run the first process's live workflows. The executor id must stay *stable* so crashed runs resume on the next boot. Given both constraints, mutual exclusion of runtimes is the correct call for the embedded model — the lock is a symptom of the topology, not a bug.

The key observation: **DBOS happily runs many workflows concurrently inside one engine** — `executeAnalysis` already dispatches parallel child workflows per step. The single-runtime constraint is purely *process topology*, not engine capacity. Multiple concurrent runs need zero harness-engine changes; they need the runtime to live in one long-lived process that many clients talk to.

## Decision

Adopt a **client–server architecture** with **full scope** as the committed end-state, built now:

- A single **`inflexa serve` daemon** (same binary as the CLI) owns the harness runtime — DBOS engine, sandbox-callback ingress, pg pools, the embedding provider — and, at end-state, all cli state access: SQLite, the chat engine, the provenance recorder.
[...]

PR #45, opened 2026-07-07
feat(cli): TUI chat drives the harness conversation agent (+ embedded skeleton)

vivere-dally:
## What this branch delivers

**The TUI chat — plain `inflexa` — talks to the harness conversation agent at Cortex-managed parity, the sidebar renders live ledger truth, and production builds carry only the product commands.** The full binding direction (`docs/harness_integration_followup/14-tui-chat-direction.md`) is delivered: one chat, no mocks, no dev surfaces in release builds.

Four OpenSpec changes, each archived with full artifacts (proposal/design/deltas/tasks/findings) under `cli/openspec/changes/archive/`:

1. **`embed-conversation-agent`** — the walking skeleton: `assembleCoreRuntime` boot adoption, conversation deps, the temporary REPL, per-analysis locks, harness barrel riders. Live-verified (chat → plan → real run with `cortex_runs.thread_id` stamped).
2. **`tui-harness-chat`** — the product surface: shared turn engine (TUI + REPL), boot-on-open behind an animation with gated input, emit adapter on the harness contracts, plan/run-card blocks, `threadId := sessionId`, profile parity auto-trigger. Live PTY E2E exit 0.
3. **`tui-sidebar-live`** — the sidebar tells the truth: ledger-sourced DATA PROFILE + RUNS sections (all mock fixtures deleted), details dialogs (`ctrl+x d`/`r` + click), active-work-gated polling. Live-verified via tmux screen capture; caught + fixed the trigger→sidebar refresh gap.
4. **`retire-proxy-chat-dev-umbrella`** — one chat, clean tree: proxy engine + `modules/intelligence/` deleted, bus contract shrunk to `prov.*`, `chat`/`profile`/`run` gated behind the `INFLEXA_BUILD_CHANNEL` baked constant (`INFLEXA_DEV=1` escape hatch).

Plus the **test-sandbox data-loss hardening** (`d517eff`): one `assertTestSandbox` choke point + a root `bunfig.toml` that makes root-level `bun test` run zero tests (incident disclosed in the comments).

## Verification
- **578 cli tests / 730 harness tests, 0 fail; typecheck + eslint clean.**
- Every change: adversarial spec-vs-code verify + review lenses (dispositions in the three review comments below), all findings fixed pre-archive.
- Live E2E: TUI boot gate → streamed turns → plan/run cards; sidebar ledger truth via tmux capture; dev-channel truth table.

## Known follow-ups (tracked in `00-progress.md`, not in this PR)
- ~~Release binary builds broken at HEAD~~ — fixed on this branch (`442a304`: deps externalized, build + build:all verified clean).
- ~~`INFLEXA_GIT_COMMIT` never actually baked~~ — fixed on this branch (`40eb417`: explicit production gate + --define; channel-keyed runtime backstop).
- Chat-model boot probe gap; `record-plan-lineage`; #33 daemon (transport swap under the unchanged TUI).

## Not merged on purpose
Left for your review per instruction.


vivere-dally:
## Self-review (3 parallel passes: correctness/lifecycle · conventions/neverthrow · streaming/tests)

Ran three independent review passes over the code diff. Findings + dispositions, now pushed as `6453d07`:

**Fixed**
- **Double-Ctrl+C teardown race** *(correctness, the one real issue)* — the second SIGINT called `void shutdown(130)` from the handler (fire-and-forget), so the aborted turn kept unwinding and the REPL resumed against a runtime being torn down (`appendTurn` racing `pool.end()`). `runTurn` now returns `"continue" | "stop"`; a second interrupt only flags `forceStop`, the turn unwinds with the pool alive, and `runRepl` drains + exits **once, after** the turn.
- **`eventSource` hardening** — guard `Array.isArray(src.callPath)` before `.length` (a malformed source falls through as top-level instead of throwing).
- **Conventions** — added the missing JSDoc on `ChatPrinter.emit` and per-site safety notes on the `readRunCard`/`readPlanCard`-step `as` casts; `finishTurn` now runs on the `prepare_failed`/`not_found` bails for uniform output.
- **Tests** — added unknown-part fallback, `tool-finished` without a matching start, both `formatMs` branches, and run-card copy-on-receive (51 pass, typecheck clean).

**Reviewed clean** (no defect): per-turn `AbortController` (fresh each turn), SIGINT listener add/remove balance (no leak), lock release on every bail path (via the `process.on("exit")` hook), boot-failure cleanup (ingress/pool/lock all released), `appendTurn` consumed on all paths, boot ordering (`sweep` strictly between `initState` and `launch`), and printer rendering fidelity (field names match the harness `PlanCardData`/`RunCardData` contract — plans don't render blank).

**Deliberately not changed** (rationale in the archived change's `design.md` / `findings.md`):
- **No `runTurn` unit test** — the turn loop is budgeted for live-E2E verification (run on analysis A1: boot, streaming, cross-turn recall, and the full plan→execute→`thread_id`-stamp loop), matching the existing `run.ts`/`profile.ts` pattern; the pure logic (printer, `selectThread`) is unit-tested.
- **`templatesDir` now an unconditional boot prerequisite** — intended (one-cohort `assembleCoreRuntime` always builds the conversation agent); mirrors `skillsDir`. The installed-cli packaging question is tracked in the design's open questions.
- **`findings.md` F1** — a dead-but-advertised Claude model passes the boot guard (boot probes the embedder but not the chat model) and fails at first turn. Surfaced live; recorded as a candidate follow-up, out of scope for this focused change.

vivere-dally:
## Review disposition (3 parallel lenses over the TUI chat commit, fixes in `8ed2414`)

Three independent review passes: **lifecycle/races**, **conventions/spec-coherence**, **rendering/tests**. (The skeleton commits had their own 3-lens pass earlier — fixes in `6453d07`.)

**Lifecycle/races: zero critical/major.** The per-turn token guard, boot lifecycle, lock exchange, and profile-parity wiring were traced through concrete interleavings (double-send, abort→send, swap mid-append, load-vs-turn races) and held everywhere.

**Fixed in `8ed2414`:**
- **`loadMessages` windowing** *(the one real bug — two lenses converged)*: past 200 turns the page-aligned last-page fetch collapsed the transcript to `total mod 200` turns (201-turn thread → 1 turn shown). Now fetches the last two turn pages + message-capped slice; the blind test fake (ignored its `page` arg) replaced with a faithful page-slicing one over 200/201/250/400-turn fixtures.
- **Mid-turn interleaving**: streamed prose now seals into a segment when a tool/plan/run part arrives and resumes in a fresh part — live order `[text][tool][text]` matches the reload path (previously all prose merged above the cards).
- **Boot render test** now composes the real mount (flexGrow scrollbox above, painted wrapper, chrome below) and sweeps heights 8–40.
- Convention items: two cast-invariant comments, two changelog-phrased comments rewritten, unpaired `tool-finished` fallback test, `MESSAGE_CAP`↔`loadPage`-clamp coupling documented, `TODO(extend)` on the reload-fidelity gaps, and the stale pre-existing `launchChat` requirement in the chat-wiring spec rewritten to the real launcher surface.

**Declined with rationale:** JSX return-type annotations on block components (CLAUDE.md's explicit carve-out); empty-card gallery exhibit (the empty-coercion path is unit-tested; cards render defensively); boot `failed`-state retry affordance (deliberate terminal state — a failed boot's resources are verified fully released; retry = reopen).

**Known limitations recorded** (in-code TODO(extend) + findings.md): harness reconstruction doesn't thread `is_error` into replayed tool cards (renders "ok" on reload); `data-presentation`/`data-preview` parts render as tagged mentions pending dedicated blocks; `system`-role messages collapse to "assistant" on reload.

Final state: **535 cli tests, 0 fail; typecheck + eslint clean.** Branch left unmerged for review as instructed.

vivere-dally:
## Review disposition #2 — the `tui-sidebar-live` commits (`0023346`…`d517eff`)

Three independent passes over change 2: a 9-hunt adversarial spec-vs-code verify (verdict **SHIP**, 0 critical — findings and dispositions in the archived change's `findings.md`) and two fresh-eyes review lenses (**both SHIP**, zero majors).

**What the passes confirmed:** the sidebar store's staleness/poll machinery is airtight (torn snapshot pairs impossible; poll re-keying on swap safe; the sandbox guard immune to the macOS `/var` symlink edge); conventions, synced specs, and docs all coherent with the landed code; the arming/staleness tests genuinely load-bearing.

**Fixed in `85a3528`** (review survivors, all minor):
- the parity-trigger poke is now **analysis-scoped** — a swap during the in-flight profile check no longer writes the old analysis's snapshots (or its toast) over the new one
- selection-drags released over the clickable sidebar sections no longer pop the details dialogs (leader keys always open)
- pure run-label helpers homed with their siblings in `sidebar_live` (the runs dialog stops importing the JSX layout module)
- the sandbox-guard test now proves **no deletion happens** (sentinel survives the throw), not merely that it throws
- one comment aligned with the point-in-time `props.runs` contract

**`d517eff` — data-loss hardening (incident disclosure):** running `bun test` from the monorepo **root** bypasses `cli/bunfig.toml`'s test sandbox; this destroyed real local dev state twice today (agent.db, `config.json`, the embeddings model dir — the proxy's auth and the prov signing key were never touched; everything was restored). The complete fix: one `assertTestSandbox` choke point every destructive test helper delegates to, plus a root `bunfig.toml` whose `[test]` preload exits before any test runs (empirical note: bun 1.3.8 does not abort on a throwing preload — hence `process.exit(1)`). Verified: root `bun test` executes zero tests; cli 573/0, harness 730/0.

Final state: **573 cli tests, 0 fail; typecheck + eslint clean.** Branch remains unmerged for review as instructed.

vivere-dally:
## Review disposition #3 — `retire-proxy-chat-dev-umbrella` (`642252f`) — **the doc-14 sequence is complete**

Combined adversarial-verify + review pass (7 hunts, all with file:line evidence). Verdict was FIX-FIRST solely on **documentation drift** — `CLAUDE.md`/`CONTEXT.md` still taught the deleted module and the session-scoped event contract to every future agent session — fixed before archive along with the spec-wording items (the real refusal is commander's non-zero unrecognized-argument error, not a literal "unknown command"; the chat-command requirement retitled to its standing dev-channel role).

**What this change is:** a pure subtraction, verified as such — the engine's only importer was the boot-helpers line; the six session bus members had zero consumers; moved code is byte-identical to its old home modulo added invariant comments; the test delta is fully accounted (573 − 8 dead + 4 moved + 7 new = 576 → 578 after the wording pass).

**The dev umbrella:** `chat`/`profile`/`run` register only in the dev channel (`INFLEXA_BUILD_CHANNEL` baked constant; release builds must declare it or fail the baked-var guard; `INFLEXA_DEV=1` is the shipped-binary escape hatch). Truth table verified live at the source runtime: dev → present; release → absent with the release surface matching the spec list exactly; release+override → present.

**Two pre-existing issues surfaced and tracked** (not caused by this change; both reproduced on unmodified HEAD):
1. **Release binary builds are broken at HEAD** — `bun run build` fails on DBOS-SDK telemetry lazy-requires (`winston`, `winston-transport`, otel-proto) unresolvable under `Bun.build`. This blocked the binary-level gate proof (the source-runtime truth table exercises the identical code path).
2. **`INFLEXA_GIT_COMMIT` is never actually baked** — `build.ts` sets it claiming the bakedEnv guard covers it, but the literal-scan finds no in-block dot access; a compiled binary would throw at first `gitCommit` read. Masked today by (1).

Final: **578 cli tests, 0 fail; typecheck + eslint clean.** With this, all three changes of `docs/harness_integration_followup/14-tui-chat-direction.md` are landed: the TUI chat is the one conversation surface, the sidebar tells the truth, and production builds carry only the product commands. Branch remains unmerged for review as instructed.

cli/src/lib/bus.ts
```diff
@@ -28,45 +28,11 @@ export const Bus = new BusEmitter();
...
-            };
-        case "part.delta":
-            return {
-                sessionId: event.sessionId,
-                messageId: event.messageId,
-                partId: event.partId,
-                deltaLength: event.delta.length,
-            };
-        case "session.error":
-            return { sessionId: event.sessionId, error: event.error };
```
vivere-dally, line 69: Why are these removed?

cli/src/lib/env.ts
vivere-dally, line 1: I think the comments in this file can be more succint

cli/src/tui/layout/design_gallery_fixtures.ts
vivere-dally, line 1: Since we did wire the `harness`, why is this file needed anymore?
If it's design-gallery only, can it maybe be colocated there? I'm unsure about this still, reason please

cli/src/modules/harness/chat.ts
vivere-dally, line 1: If `chat.ts` and `chat_printer.ts` are dev only, maybe we can move them under a `dev` sub-directory?
Same for `profile.ts` and whatever else.
That way it would be really easy for the developer to understand: "these functionalities are built for local DX and helpful to agents". Consider this idea

cli/src/modules/proxy/models.ts
vivere-dally, line 1: What is the purpose of this file? How does it wire into the app?

cli/src/tui/app.tsx
```diff
@@ -63,6 +71,64 @@ export function App(props: AppProps) {
...
+    // point-in-time view) and hands the composed lines to `ResultsDialog`, reused verbatim.
+    function openProfile(): void {
+        const snap = profileSnapshot();
+        const name = workspace.analysis?.name;
+        const title = name ? `Data profile ${GLYPHS.emDash} ${name}` : "Data profile";
+        dialogPush(() => <ResultsDialog title={title} lines={profileDetailLines(snap)} emptyText="no profile data" onClose={() => dialogClose()} />);
+    }
+
+    // Open the RUNS details view (design D5/D6). Snapshots the runs as of open; `loadSteps` fetches
+    // the latest run's steps once, over the booted pool.
```
vivere-dally: You're using spec language `(design D5/D6)`, exactly what CLAUDE.md comments guidelines forbid...

cli/src/tui/hooks/conversation.ts
```diff
@@ -83,109 +118,487 @@ function commitStream(): void {
...
+ * the SAME readers the live adapter uses; anything else the harness resolver kept → a visible tagged
+ * mention. The harness resolver already dropped what the UI does not render (reasoning, tool
+ * results). Card parts are FLAT on the reconstructed part, and the readers narrow off any object, so
+ * the part is passed straight through.
+ */
+function cortexToUiMessage(m: CortexMsg, sessionId: string): UIMessage {
+    // TODO(extend): content-fidelity gap, deliberately deferred. Any non-user role collapses onto
+    // "assistant" here because UIMessage.role is a two-value union — a `system` turn loses its framing
+    // and reads as the assistant. Widen UIMessage.role (and MessageBlock) to carry system turns
+    // honestly when the transcript needs to distinguish them.
```
vivere-dally, line 524: Maybe open a issue for this so we don't forget

vivere-dally:
**77ebb0a — `tui-profile-lifecycle` (polish round, user-requested):** the data profile now follows the input set. Drift-aware parity (current enumerated fileId set vs the profile's recorded `inputFileIds`; enumerate is read-only/hash-free, staging only on a decided trigger), live edges (debounced `prov.input_*` bus subscription + running→completed watch), clear-on-empty (harness rider: `clearDataProfile` + nullable status; the start CAS now claims NULL — without that a cleared analysis could never be profiled again), manual re-profile (palette + `r` in the profile dialog via an optional ResultsDialog action affordance), sidebar reordered to SESSION → ANALYSIS → DATA PROFILE → RUNS.

Full lifecycle verified live in one tmux pass (4 profile runs: no-drift silent skip, drift re-profile, forced restart, clear to a fully-NULL row, re-add profiling through the widened claim). harness 738 / cli 623 tests, 0 fail. Specs synced + change archived (`cli/openspec/changes/archive/2026-07-09-tui-profile-lifecycle/`).

Pre-existing finding, out of scope (untouched code): `resolveContext` ignores `--analysis <id|name>` when an anchor holds several analyses, so `inflexa profile --analysis … --status` dies at the ambiguity picker.

cli/src/lib/bus.ts
vivere-dally: These six members (`session.status`, `message.created`, `message.updated`, `part.updated`, `part.delta`, `session.error`) were the proxy chat engine's session-scoped telemetry taps. Their only consumer was rewired in 943b38f (the TUI chat writes the Solid store directly — no bus on the harness path), and their only emitter was deleted in 642252f with `modules/intelligence/`. That left them with no live emitter or consumer, so the union shrank to `prov.*` per its own rule ("only members with a live emitter AND consumer"). Verified: zero references to any removed event name remain in cli/src, including casing variants.

cli/src/lib/env.ts
vivere-dally: Agreed — the guard commentary told the incident story three times over (it lives in full in `test_support/sandbox.ts` and the root bunfig). Tightened on the branch: the import-guard block is now one 19-line comment keeping only this file's own decisions (deny-the-paths-not-each-call, the inertness matrix, throw-vs-Result at module evaluation, the one sanctioned NODE_ENV read) and pointing at sandbox.ts for the history; the `isDevelopment` doc no longer re-tells the channel-axis story the `buildChannel` note owns; the stale `(design decision D2)` spec pointer is gone. Committing on this branch now — push pending your word.

cli/src/tui/layout/design_gallery_fixtures.ts
vivere-dally: Reasoning: it is design-gallery-only. The sole importer across cli/src is `design_gallery.tsx` — this PR deleted the last sidebar consumers (`mockContext`/`mockRuns`/`ContextUsage`), and no test imports it (symbol-level grep over every export, including the `Run`/`RunStep` types). It also sat oddly in `lib/` ("non-domain infrastructure") while importing `@inflexa-ai/harness` types for a purely presentational consumer. So: colocated as `src/tui/layout/design_gallery_fixtures.ts` beside its one caller, matching the single-caller rule. If a second consumer ever appears, that's the moment it moves back out. Committing on this branch now.

cli/src/modules/harness/chat.ts
vivere-dally: Considered — the idea is right, but the cut isn't clean today. `chat.ts` and `run.ts` are pure dev surfaces (their only importer is the dev-gated registry in `cli/index.ts:138-175`), so they're movable as-is. `profile.ts` is NOT dev-only: `describeBootError` + `ensureSandboxImage` feed the product TUI boot path (`app.launch.tsx:14`, `hooks/boot.ts:5`), `seedProfileLedger` feeds the parity trigger (`profile_trigger.ts:17`), and `run.ts`/`chat.ts` lean on five more of its helpers — it needs a command/helper split before a `dev/` home means what it says. Recommendation: small follow-up — move `chat.ts` + `run.ts` into `modules/harness/dev/`, extract `runProfile`/`runProfileStatus` there too, leave the shared boot/profile helpers where the product imports them. Leaving this thread open as the tracker unless you'd rather have an issue filed.

cli/src/modules/proxy/models.ts
vivere-dally: It's product path now, not proxy-chat leftovers. Purpose: the two proxy-endpoint concerns any model caller must settle before streaming — `readApiKey()` extracts the client key `inflexa setup` baked into the proxy's config.yaml, and `resolveModelId()` resolves a default from the live `/models` list (claude > gpt > gemini > qwen, cached per process, failure never cached). Wiring: `modules/harness/runtime.ts:50` imports both at embedded-harness boot to configure the conversation agent's provider through CLIProxyAPI. Related known gap (tracked in 00-progress): boot probes the embedder but not the chat model, so a dead-but-advertised model passes the guard and fails at first turn.

cli/src/tui/app.tsx
vivere-dally: Right — and this site was the tip of an iceberg: the sweep found ~80 spec-artifact references across ~20 files in this PR's diff (`(design DN)`, `task N.M`, `FIX-N`/`C1`/`S1`/`WN` tags, plus a handful of changelog-phrased comments). All scrubbed on the branch: bare IDs dropped where the rationale was already inline, embedded ones rewritten to carry their substance in place, changelog phrasings made static, and two factual comment bugs fixed in the same pass — `turn.ts`'s `thread_gone` reachability claim (prepareChatTurn refuses only foreign-analysis threads; absent ids are re-created) and three dead `{@link enumerateInputFileIds}` references left from the rename to `enumerateInputSignatures`. Pre-existing occurrences outside this PR's diff (`prov_bridge.ts`, `prov/document.ts`) left for a follow-up sweep. Committing now.

cli/src/tui/hooks/conversation.ts
vivere-dally: Agreed. Issue text is drafted as one tracker covering the three reload-fidelity gaps of the same contract: (1) this TODO — `system` turns collapse to `assistant` (two-value `UIMessage.role`); (2) `is_error` not threaded through harness reconstruction, so an errored tool renders `ok` on reload (`content-to-cortex.ts` sets no `isError`); (3) `data-presentation`/`data-preview` parts render as tagged mentions pending dedicated blocks — plus the recommendation to pin all of it with a live-vs-reload parity test. Filing the issue was blocked by my current permission scope, so it needs your explicit go (or file it from the drafted text) — leaving this thread open until the issue exists.

vivere-dally: Filed: https://github.com/inflexa-ai/inf-cli/issues/55 — one tracker for all three reload-fidelity gaps (this role-collapse TODO, the missing is_error threading, the data-* tagged mentions) plus the live-vs-reload parity test. Resolving.

cli/src/modules/harness/chat.ts
vivere-dally: Filed: https://github.com/inflexa-ai/inf-cli/issues/56 with the proposed cut (move chat.ts + run.ts into modules/harness/dev/, extract the runProfile/runProfileStatus actions, keep the shared boot/profile helpers where the product imports them). Resolving — the issue is the tracker now.

vivere-dally:
## Consolidated review — six parallel deep passes (harness core · cli harness module · TUI chat · TUI sidebar · infra/test-safety · spec-vs-code coherence)

**Verdict: strong branch; three confirmed majors before merge.** The concurrency architecture — generation tokens, seed-CAS claims, the lock exchange, clone-on-receive — held under adversarial tracing everywhere except the findings below. The self-review dispositions in the earlier PR comments were re-verified rather than trusted, and genuinely hold (double-Ctrl+C teardown, loadMessages windowing, lock release on every bail path, thread-history numeric ordering).

### Majors (fixes in progress on this branch)
1. **Cleared data profile can be resurrected as `completed` with a NULL seed** — `harness/src/state/data-profile.ts:127-151`: `completeDataProfile`/`failDataProfile` carry no status guard and are not durable steps. An expiry-zombie (stale profile expired → inputs cleared → the still-live workflow finishes and stamps `completed` over the cleared row) or a DBOS recovery replay recreates exactly the seedless-completed incoherence the seed-CAS work forbids. Fix: CAS both terminal writes on `status='running'`.
2. **A deleted-on-disk input wedges the parity loop** — `cli/src/modules/staging/staging.ts:406-409` skips a stat-missing file as drift, but the re-profile it triggers hard-fails in `stageFile` (`:92-99`) because `resolveInputPath` never checks existence — an endless "Could not start profiling" toast that never names the file. Fix: skip stat-missing sources in `stageInputs` too (existence resolved once in the shared walk).
3. **Tool-first turns render prose above the tool chip live, below it on reload** (executed repro) — `cli/src/tui/hooks/conversation.ts:327-345` pre-mints a text part at `parts[0]`; `flushPendingText` (`:136-138`) only re-points on a non-empty buffer, so when the first event is a tool/card, later deltas and the ok-fallback land above the chip; reload orders `[tool, text]`. Fix: null/splice the empty pre-minted segment when a part interrupts, plus a live-vs-reload parity test.

### Minors (also being fixed)
- Orphaned `ai` + `@ai-sdk/openai-compatible` deps survive the engine deletion (zero importers).
- `assertTestSandbox` trusts the marker's **value**: `INFLEXA_TEST_SANDBOX=$HOME` re-arms the data-loss incident (add an `inflexa-test-` shape check); a shell-pre-exported `NODE_ENV` disables the env.ts backstop (verified on bun 1.3.14) — boot tests then touch real `~/.local/share/inflexa/locks/`.
- Analysis swap releases the old instance lock while a parity drive against it may still be queued → cross-process concurrent-staging window (`profile_parity.ts:248-251` + `workspace.ts:123-124`).
- Sidebar snapshots aren't analysis-scoped (previous analysis renders briefly under the new one and is dialog-snapshottable); a swap also fabricates a false running→completed parity edge.
- `StagedInput.mtimeMs` (required) breaks recovery-compat for pre-deploy in-flight workflows; the trigger validates the seed but not the dispatched manifest; one raw-SQL read bypasses the state layer.
- No in-flight boot memoization (second same-process boot would delete the runtime lock from under the live engine); abort classification can mask a raced provider failure; failed boot keeps the "Booting…" placeholder; a superseded boot-edge transcript load is never retried; "(pid -1)" toast on the reclaim race.
- CLAUDE.md/CONTEXT.md module maps list 4 of 11 modules.

### Also verified
- The sandbox hardening's two layers hold empirically (root refusal + nested-cwd import throw; every destructive test writer traced to a guard; bun preload/`--cwd`/config-swap bypasses probed and refused).
- Live specs match landed code on every sampled high-stakes claim; **the drift was in `00-progress.md`** — the release-build break and the `INFLEXA_GIT_COMMIT` bake gap listed as open are both already fixed on this branch (`442a304`, `40eb417`). Ledger + PR description now corrected.

### Dispositions so far
- All 7 review-comment threads answered and resolved; comment-hygiene commit `4bcb7fd` (spec-artifact reference scrub across ~20 files, env.ts doc tightening, `mock_fixtures.ts` → `tui/layout/design_gallery_fixtures.ts`, two factual comment fixes).
- Trackers: #55 (transcript reload fidelity), #56 (dev-surface split under `modules/harness/dev/`).
- Major + minor fixes are landing as a follow-up commit on this branch.

Issue #41, opened 2026-07-07
Sandbox recovery wedge: a killed run adopted by DBOS re-uses a leaked container and never makes progress


vivere-dally:
## Summary

A host process killed **after a sandbox command finishes but before its completion callback
lands** can leave a run that DBOS recovery *adopts* but that can never make progress. The
recovered `DBOS.recv` waits on a completion the leaked sandbox is POSTing to a now-dead ingress
port; the container is still alive, so the liveness watchdog never synthesises a failure, and the
recv never unblocks. The failure mode is **silent** — a `running` row that never advances.

This is distinct from #28 (data-profile kill/resume verification, closed — that path resumes
cleanly). It was observed live **twice** during change D2's E2E and, until now, lived only in the
research tracker (`docs/harness_integration_followup/01-sandbox-recovery-wedge.md`, whose §5
recommends filing it). This issue supersedes that doc.

## Mechanism

The wedge turns on **where each path creates its sandbox**, and the run path and the profile path
differ:

1. A run is executing a sandbox step. The callback ingress
   (`cli/src/modules/harness/ingress.ts`) binds an **ephemeral loopback port** (`Bun.serve({ port: 0 })`,
   `ingress.ts:101-116`) and bakes `CORTEX_BASE_URL=http://host.docker.internal:{port}` into the
   container at creation (`harness/src/sandbox/docker-client.ts:168-170`). That port dies with the
   host.
2. The container's baked callback target is **immutable for the container's life**: `SubmitExecBody`
   carries no callback URL (`harness/src/sandbox/types.ts:163-170`); sandbox-server loads
   `CORTEX_BASE_URL` once at startup (`images/sandbox-base/server/config.go:24`,
   `main.go:385`) and builds its callback client once (`main.go:402`). The completion POST runs on a
   **background, never-cancelled** context (`context.Background()`, `executor.go:283`) and retries
   connection errors **indefinitely** — `isRetryable` returns true for any transport error
   (`callback.go:123-125`), exponential backoff capped at 30s (`callback.go:28-29,115-119`). So a
   surviving container hammers the dead port forever.
3. A recovery boot adopts the workflow. **On the run path** (`executeAnalysis`), sandbox creation is
   a **checkpointed DBOS step** — `DBOS.runStep(() => deps.sandboxClient.createSandbox(...), { name:
   "sandbox.create" })` (`harness/src/workflows/sandbox-step.ts:406`) — so re-execution replays the
[...]

PR #46, opened 2026-07-08
fix(sandbox): withhold callback credentials from spawned commands, harden the container

radutanasa:
## What & why

This began as a one-line credential fix and grew, through the security audit that produced `paper/EVIDENCE.md`, into the set of sandbox-isolation fixes that audit surfaced. Four related properties, each verified against the real `sandbox-base:latest`.

### 1. The recording credential sat in the observed code's environment

`sandbox-server` passed its full environment to every command it spawned (`cmd.Env = os.Environ()`), so analysis code could read `SANDBOX_CALLBACK_SECRET` from `/proc/self/environ` and POST a **validly-HMAC-signed `/complete` callback** carrying a fabricated exit code, stdout, and provenance frame. That inverts the property the sandbox exists to provide: provenance is supposed to be *observed* by a channel the executing code does not control.

`buildCommand` now spawns children with `sanitizedEnviron()`, stripping `SANDBOX_CALLBACK_SECRET` and `CORTEX_BASE_URL`. `loadServerConfig` reads both once at startup, so the server is unaffected. Verified end-to-end: PID 1 retains the secret, the spawned command sees neither variable, and `PATH` and `LD_PRELOAD` survive — provenance capture is unchanged.

### 2. The Docker backend configured almost no isolation

It set bind mounts and CPU/memory caps and nothing else. Added, each verified against the real image (health, `provtrack.so` interposition, Python, R, and `/tmp` writes all unaffected):

| | Before | After |
|---|---|---|
| Capabilities | Docker default set | `CapDrop: ["ALL"]` |
| Privilege escalation | allowed | `no-new-privileges` |
| User | uid 1000 by image convention | `User: "1000:1000"` asserted |
| `/exec` port | published on `0.0.0.0` | not published at all — see §3 |

The Kubernetes backend already had the equivalent posture; this ports it across.

### 3. Egress is now confined behind one gateway door

`SECURITY.md`, `README.md`, and this PR template all claimed the sandbox has **no network egress**. No `NetworkMode` was set anywhere; `SECURITY.md` even listed *"unexpected network egress"* as an in-scope vulnerability, so by our own criterion the shipped default was one.

The fix is a per-**analysis** `--internal` Docker network plus a per-sandbox **gateway** sidecar — the same image, run as `sandbox-server gateway`, with no bind mounts and **no callback secret** — bridging the internal network and the default bridge: it forwards `/exec` inbound from a loopback-bound port and callbacks outbound to the ingress. The sandbox can now reach nothing but the endpoint that records its results — internet, LAN, host, and other analyses' sandboxes all unreachable — E2E-verified against the real image while Python, R, `/tmp`, and all three provenance layers keep working. Sibling steps of one analysis share a network, matching the read-only analysis-tree mount they already share; different analyses are mutually unreachable.

The same change makes completions **pullable**. A host that dies mid-exec returns on a new ingress port, so a pushed callback can never land and the run hangs in `running` forever (#41). `GET /exec/{execId}` now serves the terminal result — signed fresh at request time, provenance frame included — and `awaitExec` pulls it when the recv topic falls quiet. (A bind-mounted Unix domain socket, floated earlier as the fix, was **empirically disproved**: it cannot cross the Docker Desktop VM boundary, since macOS and the Linux VM share no kernel — see `paper/BLOCKERS.md` B4.)

### 4. The exec endpoints are now authenticated inbound

`POST /exec` and `GET /exec/{execId}` required no authentication; under the gateway a sibling step could drive them, and the GET discloses the command's stdout/stderr. Both now require a per-request HMAC signature over the same construction as the callbacks, run inbound, checked against a 300 s freshness window (`POST /exec` over the request body, the GET over an empty one); unsigned, forged, or stale → `401` before any command runs or output is disclosed.

It is **request-signing, not a bearer**, on purpose: the gateway forwards inbound bytes in cleartext, so a static credential would be exposed to the one transport that must never hold it — whereas a signature it can forward or drop but never mint another. The same secret that stops the gateway forging a *completion* now stops it forging a *submit*, and because the check tests possession of the per-sandbox secret, a sibling holding only its own secret cannot sign for this one. The `POST /exec/{pid}/kill` route is **removed** (nothing drove it; `killAll()` still handles shutdown teardown). Verified with cross-language HMAC vectors (the TS signer and Go verifier agree byte-for-byte, including the empty-body GET), Go handler tests, TS client tests, and an end-to-end run against the built binary.

## Type of change

- [x] Bug fix
- [x] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (harness typecheck + eslint + `bun test`; `gofmt` + `go test`). `cli/` is untouched.
- [ ] This PR is focused on a single logical change. — **No.** It is one *theme* — the sandbox-isolation findings from a single audit — across several commits (credential withholding, container hardening, network confinement + pull recovery, inbound auth), plus a `paper/` commit carrying the manuscript scaffold and the `EVIDENCE.md`/`BLOCKERS.md` register that surfaced them. Happy to split.

- [x] **Methods:** not applicable — no analytical method changed.
- [x] **Sandbox:** the security model is **tightened**, never loosened. The one apparent loosening — dropping the "no network egress" claim — corrects documentation to match reality, and the same PR then implements the confinement that makes the corrected claim true.
- [x] **Provenance:** prior analyses remain reproducible. The PROV document format, hash chain, and signatures are untouched. Nothing in `skills/` or `templates/` reads `CORTEX_BASE_URL`, so withholding it from spawned commands cannot change an analysis result; the pulled result serves the callback's own bytes, so the provenance frame is identical on both paths.

## Reviewer notes

1. **The server-side fixes ship inside the sandbox image**, so they take effect only once the image is rebuilt and republished to GHCR. Harness and image must ship together: the gateway's `CORTEX_BASE_URL` rewrite and the inbound signature check are a wire-protocol change, so a new harness against an old `:latest` (or vice-versa) would break.
2. **Network confinement is a Docker-backend property.** The K8s backend reaches pods directly and still has **no NetworkPolicy**, so its egress is not yet confined (tracked in `EVIDENCE.md` S6). The inbound signature check is backend-independent and applies to both.
3. `ReadonlyRootfs` and `PidsLimit` were deliberately deferred: the former needs a tmpfs for `/tmp`, and the latter counts threads, which bioinformatics tools use heavily, so the ceiling needs measuring rather than guessing.
4. `paper/EVIDENCE.md` and `paper/BLOCKERS.md` still describe issues that are **not yet fixed** (and now record which ones are). Both need review before this repository becomes public.


radutanasa:
@claude

radutanasa:
Superseded by #47, recreated from a tidied branch. Closing this one.

PR #47, opened 2026-07-08
feat(sandbox): host-polls-sandbox transport (default), dropping the gateway sidecar

### LINKED ISSUE #27: Linux Docker Engine: sandbox exec callbacks unreachable (loopback ingress) — needs bridge-gateway bind
## Summary

On native Linux Docker Engine (not Docker Desktop), a data-profile run's sandbox cannot reach the host's exec-callback ingress, so every sandbox callback is lost and each exec times out — the profile fails after the full deadline with a generic timeout. macOS/Windows Docker Desktop is unaffected (its VM forwards `host.docker.internal` to a host loopback listener).

## Root cause

The callback ingress (`cli/src/modules/harness/ingress.ts`, `startExecIngress`) binds `127.0.0.1` and advertises `CORTEX_BASE_URL=http://host.docker.internal:{port}`. On Linux Docker Engine:

- `host.docker.internal` does not resolve inside a container by default.
- Even with a `host.docker.internal:host-gateway` alias, `host-gateway` maps to the **docker bridge gateway IP** (e.g. `172.17.0.1`), not loopback — so a `127.0.0.1`-only listener refuses the connection.

The Go sandbox-server then retries the callback forever until the exec deadline (`DEFAULT_DEADLINE_MS = 300_000` in `harness/src/tasks/data-profile.ts`), and `awaitExec` never sees the done-marker.

## Why not just bind `0.0.0.0`

An earlier fix attempt bound the ingress to `0.0.0.0` (+ `ExtraHosts: host.docker.internal:host-gateway`). An automated security review flagged it (MEDIUM — Network Exposure / Unauthenticated Endpoint): `0.0.0.0` exposes the callback endpoint to the entire LAN, and the route is intentionally secret-less (HMAC is verified downstream in the workflow recv loop, not at the ingress). That change was reverted; the ingress is back to loopback-only.

## Proposed secure fix

Bind the ingress to the **docker bridge gateway address** — reachable from the bridge network and the host, but **not** the external LAN — and advertise that IP directly (no dependency on `host.docker.internal`):

- macOS/Windows (Docker Desktop): keep `127.0.0.1` + `host.docker.internal` (works today).
- Linux Docker Engine: discover the bridge gateway at boot and bind to it; set `CORTEX_BASE_URL=http://{gatewayIP}:{port}`.

Open questions to settle during implementation:

- Gateway discovery must be runtime-aware: `docker network inspect bridge` vs. `podman network inspect podman`, and custom `bip`. The cli already knows the runtime (`activeRuntime`) and can shell out via `capture`.
- Rootless docker/podman use slirp4netns — different networking; may need a distinct path or a clear unsupported error.
- `startExecIngress` / the `BootSeams.startIngress` seam is currently synchronous; bridge discovery is async, so the seam would need to go async (and the runtime boot tests updated).
- If discovery fails, fail the boot with an actionable error rather than silently timing out.

## Pointers

- `cli/src/modules/harness/ingress.ts` — `startExecIngress` (the `TODO(robustness)` on `ExecIngress.cortexBaseUrl` references this issue).
- `harness/src/sandbox/docker-client.ts` — where a Linux `ExtraHosts`/network setting would attach if the name-based approach is chosen instead.
- `harness/src/tasks/data-profile.ts` — the exec deadline the failed callbacks currently run out.

### LINKED ISSUE #41: Sandbox recovery wedge: a killed run adopted by DBOS re-uses a leaked container and never makes progress

### LINKED ISSUE #48: Poll mode: escalate to an isAlive probe after N consecutive unavailable polls
## Context

In poll transport mode (#47) the liveness watchdog's synthetic-complete fast-fail has nothing to unblock: the poll loop has no `DBOS.recv`. A sandbox that dies mid-exec (e.g. an OOM-killed container) is therefore bound by `step.timeout` — the loop keeps polling an unreachable sandbox until the deadline, then times out. Correctness holds (the deadline is the backstop; the watchdog still marks the registry and the reaper cleans up the machine), but failure latency regresses from callback mode's fast-fail to potentially hours on a long-deadline analysis step.

## Proposed follow-up

Make the poll loop liveness-aware, as noted in `harness/openspec/changes/add-poll-transport-mode/design.md`: after N consecutive `unavailable` polls, escalate to a `SandboxClient.isAlive(sandboxRef)` probe; if the machine is observably dead with no completion recorded, fail the exec with the synthetic-failure result instead of waiting out the deadline.

Design constraints:

- Keep the loop free of topic/recv coordination — the probe should be another durable step in the existing attempt sequence, replay-stable like the polls.
- `unavailable` conflates "unreachable" with "unknown execId / non-200"; only sustained unreachability should escalate.
- The distinction between a dead machine and a live-but-slow one stays with `isAlive` (backend inspect), not with poll failures alone.

Refs: #47 (trade-offs section); `harness/src/sandbox/await-exec.ts` (`awaitExecPoll`).

radutanasa:
## Summary

Inverts the sandbox exec protocol so the **host polls the sandbox** for results (the new default) instead of listening for a pushed callback, and **removes the gateway sidecar** the push path required on Docker. A third round on this branch also gives poll mode a **dead-sandbox fast-fail** (#48).

The old protocol made the sandbox *push* its results over an HMAC callback. On Docker that push forced a per-analysis `--internal` network plus a per-sandbox **gateway** container to bridge the callback out — the direct cause of two open defects, both properties of *push*:

- **#27** — on native Linux the sandbox can't reach the host callback ingress, so every exec times out.
- **#41** — a host that restarts mid-exec comes back on a new ingress port, so the pushed callback lands nowhere and the run wedges silently in `running`.

If the host **polls** instead, the sandbox never initiates a connection: there is no host ingress to be unreachable (#27), and the host always dials from its *current* identity (#41). The sandbox then needs no egress, so it can be *isolated* rather than *bridged*, and the gateway's reason to exist disappears.

Callback stays available as an opt-in second mode for deployments that prefer push — real-time progress, or many-sandbox scale with proper host networking. The embedder chooses; the local CLI defaults to poll.

## What changes

- **`SandboxTransport` = `poll` | `callback`**, chosen by the embedder at its composition root and carried to the container as `SANDBOX_TRANSPORT`. Backends stay mode-agnostic; the OSS/CLI default is `poll`. Inbound request signing on the exec endpoints applies in **both** modes.
- **Poll mode (new default).** sandbox-server never dials out (no `callbackClient`, no `CORTEX_BASE_URL`). Progress events accumulate in a bounded per-exec ring; `GET /exec/{execId}?since={cursor}` serves a **signed** `{ status, events[], cursor, result? }`. `awaitExec` becomes a durable poll loop (steps `sandbox.poll-exec-result.${execId}.${n}`, no `DBOS.recv`/topic) that verifies each body against the per-sandbox HMAC. The CLI starts **no** callback ingress.
- **Callback mode (opt-in).** The pre-gateway push path, retained: the sandbox is *permitted* egress and POSTs signed callbacks; the embedder runs the ingress; `awaitExec` uses the `DBOS.recv` loop with `GET /exec/{execId}` as its recovery backstop.
- **The gateway is removed** (`gateway.go`, the `--internal` network + gateway lifecycle in `docker-client.ts`). In poll mode the Docker sandbox is confined by an **in-container egress firewall**: the exec port is published to `127.0.0.1` only, and a root entrypoint holding `CAP_NET_ADMIN` installs `iptables -P OUTPUT DROP` (allowing loopback + established return traffic) then `setpriv`-drops to the uid-1000 workload, which can neither reach the network nor flush the rules. Spike-verified on Docker Desktop: host polling works, egress hard-blocked, uid-1000 can't flush. On K8s, poll-mode confinement is a **cluster-side NetworkPolicy expectation — the harness ships none**; without one, K8s egress is unconfined.

The security core from the earlier hardening work is **kept**: `sanitizedEnviron` (callback secret withheld from spawned commands), inbound request signing on `POST /exec` + `GET /exec/{execId}` (a request signature, not a bearer — so any cleartext hop can drop a request but never mint one, and a sibling sandbox holding only its own secret can't drive this one's `/exec`), container hardening (uid 1000, `CapDrop: ALL`, `no-new-privileges`), and the signed `GET /exec/{execId}` — which *becomes* the poll primitive.

## Review hardening (second round, 25d120d)

An adversarial review of this PR produced a hardening round, landed here:

- **Poll loop.** Events are deduped client-side by `seq > cursor` — the signature covers the body, not the request's `since`, so a replayed signed snapshot verifies against any poll and must not re-emit delivered events. A seq gap above the cursor (events shed by the bounded ring before delivery) surfaces as an advisory warning instead of silent loss. The loop always polls once more upon crossing the deadline before `ExecTimeoutError`, matching the callback loop's deadline pull. The cadence is now two-phase — 1.5s for the exec's first ~minute, then 10s — bounding DBOS step-row growth on hours-long execs (~720 rows/hour instead of ~4,800), attempt-derived so it replays identically.
- **Fail-closed confinement.** sandbox-server refuses to start when `SANDBOX_EGRESS_FIREWALL=1` but euid is 0: an image that bypasses the entrypoint's firewall+privilege-drop chain (e.g. an overridden `ENTRYPOINT`) now fails loudly at create time instead of running as root with `NET_ADMIN` and unconfined egress while looking healthy. The entrypoint mirrors the egress deny with `ip6tables` when the container has an IPv6 stack, so a dual-stack bridge is not a hole through the IPv4 firewall.
- **Docs de-contradicted.** The retired gateway/internal-network topology is purged from the PR template, `CONTRIBUTING.md`, and `inbound_auth.go`; `SECURITY.md` now states network reachability per transport mode (poll: nothing outbound; callback: bridge-reachable, the HMAC is the boundary).

## Dead-sandbox fast-fail in poll mode (third round, 6c9bd5d — closes #48)

The trade-off originally accepted here — a dead sandbox in poll mode rides `step.timeout` because the watchdog's synthetic-complete has no recv to unblock — is now fixed on this branch rather than deferred:

- **`awaitExecPoll` escalates sustained unavailability.** After 4 consecutive `unavailable` polls (an `ok` poll resets the streak), the loop runs a durable backend-inspect probe (`sandbox.probe-liveness.${execId}.${k}`). An observably **dead** machine returns the synthetic-failure result in-loop (`sandbox-oom-killed` when the backend attributes the death to the memory limit, `sandbox-dead` otherwise) instead of waiting out the deadline; **alive** (live-but-slow exec, evicted execId — `unavailable` deliberately conflates these) and **inconclusive** (transient backend API error) resume deadline-bounded polling. Probe scheduling is a pure function of the checkpointed poll outcomes, so replays walk the identical step sequence — no topic, no recv, no cross-workflow race gate.
- **The escalation is a transport-agnostic module** (`sandbox/liveness.ts`): the consecutive-unavailable policy, a never-throwing probe runner, and the synthetic-result constructor hoisted from `watchdog.ts` so both adjudicators build identical failures. This is increment 1 of moving liveness verdicts to the waiter; the callback loop's pull backstop can adopt the same policy next, after which the watchdog's `sendSynthetic` role can retire. `isAlive` joins `AwaitExecOptions` as an injected seam, self-wired by `createSandboxClient` — no composition-root changes.
- Typical detection: ~6s on the fast poll cadence, ~40s on the slow one — at or better than callback mode's ~65s watchdog bound.

## Review hardening (fourth round, 97c47d0)

- **Submit body cap.** sandbox-server buffers at most 16 MiB of a `POST /exec` body (`http.MaxBytesReader`) and rejects larger with `413` before spawning — the body is read before its signature can be verified (the signature covers the bytes), so the cap, not the auth check, bounds an unauthenticated peer's memory cost. Sized generously because `write_file` ships whole files base64-inflated inside the command array. Spec'd in the `sandbox-server` submit requirement.
- **K8s poll-mode env parity.** The K8s backend now omits `CORTEX_BASE_URL` in poll mode, matching Docker — the pod spec itself documents that no callback egress is expected.
- **`PollResponseSchema.truncated`** carries a comment stating it's deliberately unused by the loop (the seq-gap arithmetic supersedes it) and kept on the wire for forward-compat.

## Spec + docs

- OpenSpec changes `add-poll-transport-mode` in **both** `harness/openspec` (deltas to `harness-sandbox-exec`, `sandbox-server`, `docker-sandbox-provider`) and `cli/openspec` (`harness-runtime`), plus `add-poll-liveness-escalation` (`harness-sandbox-exec`).
- **The deltas are synced into the main spec trees** (a3b3441), so `openspec/specs/` matches the implementation at this PR's head — the gateway requirement is removed, the two-transport + poll-loop + liveness-escalation requirements are in, and the stale retry-reuses-one-signature clause in `sandbox-server` is corrected to re-sign-per-attempt. `openspec validate --all --strict` clean in both trees (56 + 63 items). The change directories stay active for the apply/archive lifecycle.
- `SECURITY.md`, `README.md` (sandbox bullet), `harness/CLAUDE.md`, `harness/CONTEXT.md`, `images/sandbox-base/README.md` rewritten to the two-mode + egress-firewall story.

## Verification

- `go test ./...` ok; `gofmt` clean on changed files.
- Harness `tsc` clean; full harness suite **782 pass** (`bun test`), including: poll loop returns on terminal, forwards incremental events once — replayed snapshots don't duplicate, replay-stable step names, forged poll response hard-cancels, ring-shed gaps warn, deadline polls once more then times out, cadence backs off; transport dispatch; callback recv loop unchanged; the liveness escalation's policy/probe/constructor unit tests and all seven loop-level escalation scenarios (dead fast-fail, OOM reason, live-but-slow, streak reset, inconclusive probe, replay-stable probe step names, no-seam fallback).
- CLI `tsc` clean; harness-runtime boot tests — 31 pass (boot binds no ingress in poll mode).
- Cross-language HMAC over the `?since` body uses the same `signCallback` construction proven equivalent earlier; both sides' unit tests sign/verify the poll body.
- **Not yet run:** E2E against a built `sandbox-base` image (needs Docker). The firewall mechanism itself is spike-verified; the escalation is contract-tested.

## Trade-offs

- **The CLI hardwires poll** via a module const; callback is a managed-embedder path.
- Only *observable* death fast-fails: a host↔sandbox network partition with a healthy machine still waits out the deadline by design. Observable death covers stopped, OOM-killed, **and removed** containers — Docker `isAlive` maps an inspect 404 to `dead` (`docker-client.ts`, "404 (never created) → dead" test), matching K8s's missing-pod handling; an earlier revision of this description wrongly listed the 404 case as a gap.

Closes #27. Closes #41. Closes #48.


krajla (changes requested):
I really dislike the idea of a sidecar that intermediates request just so it bridges the callback. I made a comment in the issue linked to this PR in which I suggested the harness owner use a polling system, the sandbox can remain isolated with no egress.

vivere-dally (approved):

PR #70, opened 2026-07-10
feat(cli): record the LLM model identity as a PROV agent

### LINKED ISSUE #68: Record the LLM model identity as a PROV agent
The provenance document records *that* the CLI performed an action, but never *which model reasoned about it*. For a project whose positioning rests on open, inspectable provenance, "we record everything about how this result was made, except which AI produced it" is an awkward gap — and it is one of the few things a reviewer of an AI-assisted analysis would most want to know.

## Current state

`ProvActor` is a three-way union (`cli/src/types/prov.ts:18-27`):

```ts
export type ProvActor =
    | { kind: "user"; email: string }
    | { kind: "anonymous" }
    | { kind: "system"; version: string; commit: string };
```

`systemActor()` (`cli/src/modules/prov/prov.ts:50-52`) stamps the **CLI's** package version and git commit. `recordPreamble` renders it as a `prov:SoftwareAgent` with `inflexa:version` / `inflexa:commit` (`cli/src/modules/prov/document.ts:66-94`), and every step and command activity gets `wasAssociatedWith(activity, thatAgent)`.

So an LLM-driven step — where a model chose the method, wrote the code, and interpreted the result — is attributed to a generic software agent named "inflexa cli", indistinguishable from a step where the CLI moved a file.

The model id **is** available. It is threaded through the harness deps as `comp.model` (`cli/src/modules/harness/run_deps.ts:45-46`, and again at `:76`, `:132`, `:163`, `:189`, `:211-212` for the chat / decision / synthesis seats). It simply never reaches a `prov.*` event or the document builder.

## Proposed

Model the LLM as its own PROV agent, and use `actedOnBehalfOf` to express delegation — this is exactly the relation PROV has for it.

1. **Extend `ProvActor`** with a model kind, or (cleaner) carry the model alongside the system actor so one activity can be associated with both agents:

   ```ts
   | { kind: "model"; provider: string; model: string; }
   ```

2. **Emit a `prov:SoftwareAgent`** per distinct model, e.g. QName `inflexa:agent-model-{digest(provider,model)}`, with attributes:
   - `prov:type = "inflexa:Model"`
   - `inflexa:provider` — `anthropic` | `openai-compatible` (the two native kinds, `harness/src/providers/ai-sdk.ts:156-165`)
   - `inflexa:model` — the configured model id
   - `inflexa:endpoint` — **only** for `openai-compatible`, and only the host, never a key. For a local endpoint this is `localhost`, which is itself meaningful provenance.

3. **Relate it:**
   - `wasAssociatedWith(activity, modelAgent)` on the step and command activities the model drove.
   - `actedOnBehalfOf(modelAgent, cliAgent)` — or the reverse, depending on how we read delegation. My reading: the CLI is the agent the user directed; the model acted on its behalf.

4. **Do not record**: API keys, base URLs with credentials, or prompt content. The point is *which intelligence entered the process*, not what it saw.

## Design questions to settle

- **Seats.** `run_deps.ts` already distinguishes chat / decision / synthesis models (`:163`, `:211-212`), even though all three resolve to one config id today (D6). Do we record one model agent or three? Recording per-seat now costs nothing and future-proofs the split.
- **Resolution timing.** `harness.model` may be `null` in config, meaning "resolve the default from the proxy's `/models` at boot" (`cli/src/modules/harness/config.ts:16,59`). The **resolved** id must be what lands in the document, not `null`. This is the actual work in this issue.
- **Proxy opacity.** When routing through CLIProxyAPI, the model id we hold is what we asked for. If the proxy silently substitutes, our record is wrong. Worth deciding whether we record the requested id, the id echoed back in the response, or both.
- **Determinism.** Model ids are replay-stable within a run, so this does not disturb the DBOS re-execution dedup story (record ids remain keyed by run/step/digest).

## Why this matters

`docs/WEBSITE_REPOSITIONING.md` (in `atrium`) rules out the claim *"records which model produced each step"* against the current code. It is a claim we should be able to make — arguably the most differentiated provenance claim available to an open, model-agnostic orchestrator, since a vendor whose model is welded to the harness has no reason to record it.

Related: the same document notes that container image digest, environment, random seeds, and tool versions are also uncaptured, which is why "full reproducibility" is currently an overclaim. Those are separate issues, but this one is the cheapest and the most visible.

radutanasa:
Fixes #68.

The signed provenance document now records **which model reasoned about each model-driven activity** — the gap the issue calls out: "we record everything about how this result was made, except which AI produced it."

## What this does

- **`ProvModelId`** (`cli/src/types/prov.ts`) — the vendor-qualified **`{provider}/{model}`** name (the convention model ecosystems use, e.g. `anthropic/claude-opus-4-8`), enforced as a template-literal type with an **open** provider vocabulary — no closed union to keep in step with any provider list. The model part is always the boot-resolved id (never a config `null`). No keys, credentialed URLs, or prompt content are ever recorded.
- **Events** — `prov.step_completed` and `prov.command_executed` carry a **required** `model` field (a forgotten wiring is a compile error, never a silent gap), following the established rides-the-event pattern so the recorder never infers across events.
- **Document builders** — one `prov:SoftwareAgent` (typed `inflexa:Model`, labelled with the qualified name) per distinct name under a deterministic QName; `wasAssociatedWith(step|command, modelAgent)` under the same id template as the cli-agent associations (so both coexist and DBOS re-emission dedups under `unified()`); `actedOnBehalfOf(modelAgent, cliAgent)` for the delegation — the CLI is the agent the user directed, the model acted on its behalf.
- **Wiring** — the composition carries the two facts separately (`model`: the bare resolved id, which is the API model param; `modelProvider`: the vendor slug), and the bridge emitters compose the qualified name — no redundant combined field to drift. Until provider+model become user-specified config, the boot derives the slug from the model family via the same table the proxy's default-model ranking uses (`claude`→`anthropic`, `gpt`→`openai`, `gemini`→`google`, `qwen`→`qwen`), recording `unknown/{id}` for an unrecognized family rather than guessing silently — a `TODO(extend)` that retires when the provider config change lands.

## Design questions from the issue, as settled

- **Seats** — one identity per composition today (all seats resolve to one config id, D6); each emitter takes its own value at construction, so a future seat split threads through with no event-shape change.
- **Resolution timing** — the boot-resolved id is what lands in the document.
- **Proxy opacity** — we record the **requested** id; response-echo capture needs harness-side surface and is a noted non-goal.
- **Determinism** — model ids are replay-stable within a run; the association/delegation ids fold the agent digest exactly as the existing cli-agent relations do, so a recovery boot that resolves a different default records both agents honestly (same accepted semantics as a cli upgrade mid-run).

## Review rounds

1. An 8-angle code review ran on the diff pre-PR; fixes folded in (deterministic-identity hardening, no dual source of truth for the model in the composition, doc corrections).
2. PR review (@vivere-dally), round one: the original `ProvModelRef` union over the harness's native provider kinds ("anthropic" | "openai-compatible") and its endpoint host were removed — a closed protocol-kind vocabulary with marginal audit value; the endpoint was populated by no production wiring.
3. PR review, round two: adopted the `{provider}/{model}` convention (4aef978) — provider+model will become user-specified config, making the provider a configured fact; the interim family derivation bridges until then.

## Process

OpenSpec change `record-model-agent` (proposal/design/delta specs/tasks) created, implemented, verified, synced into `prov-run-events` + `prov-harness-bridge` main specs, and archived — all in this PR, per repo convention; the artifacts track the review-driven design as it evolved.

## Verification

- `bun run typecheck` ✓, `bun run lint` ✓, `bun test`: **784 pass / 0 fail** (78 files)
- Builder tests: dual association (cli + model agent) on steps and commands, one-agent-per-name across activities, delegation dedup under re-emission, qualified-name identity, PROV-JSON round-trip; bridge/wiring tests assert the composed name rides every emitted event; `modelProvider` family-derivation unit tests.

Notes: the cli typecheck baseline requires a fresh `harness/dist` (`cd harness && bun run build`) — the stale-dist errors on main predate this branch. The `lint (cli)` / `test (cli)` CI failures are a registry-auth 403 fetching `@inflexa-ai/tsprov` from npm.pkg.github.com in Actions — pre-existing (fails on main's own pushes too), unrelated to this change.


cli/src/types/prov.ts
```diff
@@ -26,6 +26,36 @@ export type ProvActor =
...
+          provider: "openai-compatible";
+          model: string;
+          /**
+           * The endpoint HOST only — itself meaningful provenance (`localhost` says "local model");
+           * never a key or a credentialed URL. REQUIRED: the harness's openai-compatible config
+           * always carries a `baseURL`, and an optional field here would let a host silently omit
+           * the most load-bearing fact of this arm — then splitting the agent's identity (the
+           * endpoint keys the agent QName) when the omission is later fixed.
+           */
+          endpoint: string;
```
vivere-dally: why this separation between Anthropic and everything else?
why should we include the endpoint?

vivere-dally: I don't think `openai-compatible` is a reasonable provider type here. It should be like this:
`anthropic/opus4.8`
`openai/gpt5.5` or whatever
`deepseek/whatever`.

That's how Mastra for example accepts models, and Bifrost. It's usually a typescript type of `string/string` where `{provider}/{model}`

radutanasa: Both dropped in 6bb121b. The separation mirrored the harness's two native provider kinds (it came from the issue text), but it was a closed vocabulary the cli would have had to keep in step with the harness for marginal audit value — and the endpoint was populated by no production wiring (the CLI reaches every model through the local proxy, whose host is plumbing, not the model's home). The record is now just the resolved model id, captured verbatim as an opaque string.

radutanasa: One constraint kept me from adopting `{provider}/{model}` outright: the ids the CLI actually holds are bare proxy ids (`claude-…`, `gpt-…`, `gemini-…` from CLIProxyAPI's `/models`), so minting a vendor prefix here would mean inferring it from an id substring — fabricating provenance rather than capturing it. 6bb121b resolves it one level up: `ProvModelId` is an opaque string recorded verbatim with no provider vocabulary of its own, so when model naming adopts the `{provider}/{model}` convention (a model-config concern — Mastra/Bifrost-style, and worth doing harness-first), the fully qualified name lands in the signed document unchanged. There's a test pinning that: "a vendor-qualified {provider}/{model} id is captured verbatim, not parsed".

vivere-dally: > One constraint kept me from adopting `{provider}/{model}` outright: the ids the CLI actually holds are bare proxy ids (`claude-…`, `gpt-…`, `gemini-…` from CLIProxyAPI's `/models`), so minting a vendor prefix here would mean inferring it from an id substring — fabricating provenance rather than capturing it. [6bb121b](https://github.com/inflexa-ai/inf-cli/commit/6bb121b64429bfaba62b898e7f4caf3f6438a676) resolves it one level up: `ProvModelId` is an opaque string recorded verbatim with no provider vocabulary of its own, so when model naming adopts the `{provider}/{model}` convention (a model-config concern — Mastra/Bifrost-style, and worth doing harness-first), the fully qualified name lands in the signed document unchanged. There's a test pinning that: "a vendor-qualified {provider}/{model} id is captured verbatim, not parsed".

I don't love that you dropped that tbh. CLIProxyAPI is wired now just for dev speed. A change will be made where the user will in fact specify the provider, and the model and it will feed them to the harness

radutanasa: Adopted in 4aef978. `ProvModelId` is now the `{provider}/{model}` name, enforced as a template-literal type (`${string}/${string}`), with the provider an open string — so `anthropic/claude-opus-4-8`, `openai/gpt-5`, `deepseek/whatever` all record as-is. Until your provider+model config change lands, the boot derives the slug from the model family using the same table the proxy's default-model ranking already uses (claude→anthropic, gpt→openai, gemini→google, qwen→qwen), and an unrecognized family records `unknown/{id}` rather than a silent guess — that derivation carries a TODO(extend) pointing at this thread and retires the moment config supplies the provider directly. One implementation note: the composition keeps the bare id and the provider slug as two separate fields (the bare id is the model param on every API call) and the provenance emitters compose the qualified name.

vivere-dally:
Don't merge this pls

vivere-dally (dismissed):

vivere-dally (approved):

PR #84, opened 2026-07-13
feat: seed agent loops with structured briefings

krajla:
## What changed

- add typed, pure briefing definitions and persist standing conversation briefings as a pinned, token-budget-exempt thread prefix
- seed conversations and planner runs with data-profile, prior-runs, and prior-plan briefings, with briefing-card metadata for the host
- hand each dependent sandbox step the persisted summaries and canonical artifact paths of its upstream dependencies through checkpointed, unpersisted `step-handoff` briefings
- scaffold the CLI companion OpenSpec change for briefing-card rendering
- sync the completed briefing deltas into the harness specs and archive the completed conversation-briefings, prior-runs, step-handoff, and poll-liveness changes

## Why

Agents need concise, trusted runtime context without rediscovering data profiles, previous runs, plans, or upstream step outputs. A shared briefing abstraction gives each loop an explicit ordered context prefix while preserving the different durability rules of conversation history, planner invocations, and sandbox workflows.

## Impact

- first conversation turns retain standing context even when older turns are evicted by the token window
- planners receive harness-owned context without callers passing duplicate `dataContext` or `priorRuns` inputs
- dependent analysis steps start with upstream interpretations and artifact locations in dependency order; missing summaries degrade to omission and root steps retain prompt-only behavior
- `add-poll-transport-mode` remains active because its Docker E2E and landing tasks are not complete

## Validation

- `bun run build`
- `openspec validate --specs --strict --no-interactive` (56 specs passed)
- `bun test src/prompts/briefings/step-handoff.test.ts src/workflows/sandbox-step.test.ts src/workflows/execute-analysis.test.ts src/schemas/render-step-prompt.test.ts` (47 passed)
- full `bun test`: 755 passed before the sandboxed run failed to find a container runtime; an elevated retry reached the runtime but timed out during PostgreSQL test setup, so the database/DBOS portion could not be completed locally


krajla:
This approach over-complicates the briefing system

PR #104, opened 2026-07-14
feat: provision reference data from its upstream publishers

krajla:
## Summary

Reference data is downloaded **directly from the third party that publishes it**. This project hosts, mirrors, and redistributes nothing, so there is no distribution endpoint to configure and the licence you accept is the upstream's.

- define the host-neutral reference-data catalog, install plan, receipt, and dynamic sandbox discovery contracts in the harness
- add CLI `refs path/list/download/verify`, setup integration, and read-only `/mnt/refs` sandbox mounts
- catalog 9 datasets — NCBI gene identifiers (human/mouse/rat), Reactome, WikiPathways, CollecTRI, GTEx v8, CellTypist — each pointing at its publisher's own URL
- keep user-added files discoverable without adopting them, and explain how to contribute supported datasets

## Integrity: two classes, because four upstreams can't promise immutable bytes

The class is a property of the upstream, not a preference of ours, and `refs list` says which one you're getting.

| class | datasets | guarantee |
|-|-|-|
| **pinned** | wikipathways-human, collectri-human, gtex-v8, celltypist-immune | the catalog carries the size and SHA-256; a download that doesn't match fails and activates nothing |
| **unpinned** | ncbi-gene-{human,mouse,rat}, reactome-pathways, reactome-mappings | trust-on-first-use: the installer records the bytes it actually received, and `verify` proves they haven't changed *since install* |

The unpinned class exists because those upstreams rebuild the same URL in place: NCBI regenerates `gene_info` continuously (its `Last-Modified` moved during this PR's own testing), and Reactome overwrites `current` each release and **deletes** its predecessors — `/download/95` and `/download/96` are already `404`, so there isn't even an archival URL to pin instead. A checked-in digest there wouldn't be a guarantee; it would be a scheduled outage. The weaker guarantee is labelled as weaker rather than hidden behind a checksum that happens to be stale.

Every artifact is a file an upstream actually serves. The `.parquet` conversions in the first draft of this PR were derived locally, which meant no third party could serve them and their digests attested only to the machine that produced them.

`reactome-mappings` is split out and **not recommended by default**: upstream serves those identifier-mapping tables as ~700 MB of uncompressed TSV, two orders of magnitude more than the GMT most workflows actually want.

## The sandbox image assumes nothing

`ENV CELLTYPIST_FOLDER=/mnt/refs/celltypist_models` is removed. It pointed at a layout the ref store doesn't produce — its own comment claimed "the ref-store builder mirrors that layout", describing a convention that no longer exists — so CellTypist found nothing there, fell back to downloading the model, and died against the egress firewall instead of reporting that nothing was provisioned.

No library-specific reference path is compiled into the image now: the store is optional and its layout is catalog-owned and versioned, so a baked path either dangles or pins the image to one catalog version. Agents call `list_available_refs`, get absolute paths back, and pass them explicitly — the contract Azimuth already required — exporting a library's env var themselves when one is needed. The sandbox prompt now states that the store may be absent and that a missing reference is a normal state to report, not to route around.

## Fixes

- **Same-size corruption was unrepairable and falsely reported.** The already-installed check compared sizes, so a corrupted file of unchanged length was skipped and reported as `Installed (0 B downloaded)`. It now compares digests and re-downloads to heal. `--force` re-fetches an intact install, which is also how an `unpinned` dataset is refreshed once its upstream moves on.
- **Resume is pinned-only.** Appending to a partial whose mutable upstream has since replaced the file would splice two files into one that verifies against nothing.
- **Redirects can't downgrade.** `fetch` follows redirects, and an `unpinned` artifact has no digest to catch a hostile substitution, so bytes served from a non-https location are refused.
- **Receipt reads are contained to their dataset** (`managed/` and `user/` are siblings; the schema already refuses traversal, this is the second lock).
- **One corrupt receipt no longer hides the others** — a shared `try` abandoned the scan loop, silently dropping every receipt sorting after the bad one.
- Activation no longer strands a staging directory per attempt.

## Validation

Exercised end-to-end against the **real upstreams**, not a stub:

- `refs download` from `data.wikipathways.org`, `ftp.ncbi.nlm.nih.gov`, and `reactome.org`; pinned and unpinned both install and `verify` clean. Reactome's five chunked, `Content-Length`-less files download and verify.
- Catalog digests confirmed byte-exact against the publishers (WikiPathways GMT, both CellTypist models).
- Same-size corruption is detected and healed; `--force` refreshes; resume issues a real `206`.
- In a sandbox container with the real bind mount, `list_available_refs` discovers the datasets with catalog metadata joined from the receipt, writes fail with `Read-only file system`, and the data is usable (47,819 rat gene rows; 987 WikiPathways pathways).
- harness: `lint`, `typecheck`, tests, OpenSpec strict 62/62. cli: `lint`, `typecheck`, tests, OpenSpec strict 71/71.

## Follow-up (not in this PR)

The published sandbox images still carry the old baked `CELLTYPIST_FOLDER`; they must be rebuilt and republished for the Dockerfile change to take effect. CI does not rebuild the image, so it won't catch this.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -65,3 +65,16 @@ Analyses run inside a **sandbox image** that bakes the R / Python / conda / Node
 - **No local store** — the packages ship inside the pulled image, so there is no `~/.local/share/inflexa/libs` tree, no `/mnt/libs` bind mount, and no architecture-forcing. `harness.sandboxImage` (in `config.json`) records the pulled image tag; set it to a custom `FROM`-extended image to run your own.
 - **Extend it** — `FROM ghcr.io/inflexa-ai/sandbox-python-r` then `RUN pip install …` / `install.packages(…)` lands in the store automatically (the image exports `PIP_TARGET`/`R_LIBS_USER`/`INFLEXA_LIB_ROOT`); run `inflexa-libs-refresh` afterward so the additions show up in `list_available_packages`. See [`images/README.md`](../images/README.md).
 - **Managed deployments** still mount per-track tarballs read-only (cold-start friendly); those tarballs are extracted from these same images by the build pipeline and are infra-managed, not a CLI concern.
+
+## Reference data
+
+`inflexa refs path` prints the public reference store (normally `~/.local/share/inflexa/refs`). When that directory exists, it is mounted read-only in sandboxes at `/mnt/refs`; sandboxed analyses remain offline and cannot download into it themselves.
+
+| Command | Does |
+|-|-|
+| `inflexa refs list` | List the harness catalog with versions, sizes, source/license links, and local state |
+| `inflexa refs download [ids...]` | Download, checksum, and atomically activate selected catalog datasets |
+| `inflexa refs verify [ids...]` | Hash active managed files and report missing or modified content |
+| `inflexa refs path` | Print the host path without creating it |
+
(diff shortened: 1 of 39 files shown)


krajla:
@claude review

krajla:
@claude review

PR #108, opened 2026-07-15
Seed agents with authoritative context; fix tool-use contracts; enable caching

krajla:
## What this is

A pass over everything that becomes model context in the harness — system prompts, tool definitions, the step/seed briefing, working memory, and prompt caching. The goal was to slim context usage, enable caching, and make the agent less prone to tool-use mistakes, without losing performance.

## The honest headline on context size

Moving per-tool manuals out of the conversation system prompt and into tool descriptions is a **maintainability and correctness** win, not a token win: every registered description and schema ships on every request regardless of where the words live. The conversation system prompt dropped ~53% (47.5k → 22.5k chars) but the tool surface absorbed it. The real reductions came from two places that genuinely remove bytes from every request — splitting the 12.6k report-brief schema behind a just-in-time trigger (−11k), and enabling caching (repeated input at 0.1x). The largest structural win is on the **sandbox** side, where step system prompts are now byte-identical and cacheable across steps.

## Correctness — tool-use bugs fixed

Each of these was a prompt instructing a call the schema rejects, or a contract the model never received:

- **`submit_synthesis` / `submit_plan`** carried `z.unknown()` schemas — the model got *no* output contract. `submit_synthesis` has five required fields named nowhere, so it failed its first call on essentially every run. Both now carry their real Zod schemas; `validate_plan` stays permissive on purpose (a validator must accept a malformed candidate to report what's wrong).
- **`submit_profile`** was instructed to emit range-shaped `metrics` against a flat-scalar schema — the whole profile submission failed on cheminformatics data.
- Prompt/schema drift removed: `workspace_search({mode:"vector"})`, `list_files({maxDepth:3})`, `get_bioactivity` missing its required `type`, `iterate_report`'s "skip encoding", six hyphenated tool ids, a ghost data-profile file the prompt told agents to read.
- **`update_working_memory`** now validates `runId` against `cortex_runs` (a typo used to mint a phantom `### Run` bucket that rendered forever).

## Seed / briefing — authoritative context instead of rediscovery

- Step seeds are composed **at dispatch inside a durable step**, not frozen at plan-submission, so a step can finally carry what its completed dependencies produced. `ExecuteAnalysisInput.promptByStepId` (strings) → `planStepById` (data).
- The data profiler's full output now **persists** (organism + taxon id + confidence, experimental design, QC concerns, per-file dimensions) instead of being collapsed to `{summary, files:[{path,description}]}`. New `inspect_data_profile` tool reads it from the DB, paged. No profile file on disk.
- `generate_plan` loads the profile **server-side** instead of relying on the conversation agent to retype a profile it never saw; `analystNotes` is a separate field for facts only the user knows.
- Sandbox system prompts no longer interpolate per-step paths (paths ride in the seed), so they're byte-identical across steps → cacheable. The unconditional "orient-first" discovery mandates are gone — the seed carries what they compensated for.

## Working memory, SOUL, caching

- Working memory gains write-time caps and `revise`/`retire` on findings (it could previously only grow); empty sections no longer render.
- SOUL reaches all ten LLM call sites instead of two — the run synthesizer, report builder, and clinical-consequence annotator author user-facing scientific claims and carried no non-fabrication guardrail.
- Prompt caching enabled behind a host-agnostic `promptCache` policy, with cache-token usage read from the vendor-neutral `usage.inputTokenDetails` (works for Anthropic *and* OpenAI-compatible providers). Note: the default CLI path (CLIProxy/Claude Max OAuth) disables caching — the new metrics reveal this.

## Tool consolidation

17 one-endpoint bio lookups → 5 action-parameterized tools (`chembl`, `pubchem`, `comptox`, `opentargets`, `pubmed`), consolidating by data source. 46 → 35 conversation-agent tools. Source-level caveats (e.g. the EPA API-key rule, previously repeated 4×) stated once.

## Caching-stability fixes

- **`loadRecent`** evicted the oldest turn every turn once over budget, shifting the message prefix and busting the whole history cache. It now evicts in whole blocks, so the prefix holds still between block boundaries.
- **`iterate_report` → `plan_report` + `submit_report`**: the 12.6k brief schema no longer rides the always-on surface. `plan_report` returns it just-in-time as a tool result; `submit_report` validates the opaque brief inside `execute` (the `validate_plan` trade). Explained in a comment on the tool.

## Specs

11 OpenSpec specs + `CLAUDE.md` corrected where the changes falsified them (path substitution, echart-layout obligations moved to the renderer, working-memory render, data-profile shape, caching policy, hyphenated ids).

## Verification

Harness: `tsc` clean, **1215 pass / 0 fail**. CLI: `tsc` clean, **228 pass / 0 fail**. Rebased onto latest `main` (36 commits, including the reference-data-from-upstream feature — the `list_available_refs` rewrite was reconciled to keep both main's live-scan and this branch's bounding). All commits DCO signed-off.

## Follow-ups (not in this PR)

- Consolidation and the prompt→description migration are context-neutral by design; the remaining conversation-agent reductions would come from further schema slimming.
- Caching wants a production read to confirm it engages on non-OAuth paths.
- `runAgent`'s wrap-up tool-set swap remains a caching defeater (documented at its site).


radutanasa (approved):

krajla:
@claude review

PR #118, opened 2026-07-15
feat(cli): bundle skills/templates in the binary; ship linux-arm64

krajla:
## Why

The harness reads `skills/` and `templates/` as plain directory trees off disk (`skillsDir`/`templatesDir`). In a compiled binary those defaulted to `null` (`import.meta.dir` is meaningless there), so `runtime.ts` rejected boot with `skills_dir_missing`/`templates_dir_missing` — **an installed OSS user couldn't run an analysis** without hand-editing `config.json` to point at a git checkout.

This bundles the content **into the binary** so one `curl | bash` yields a fully working, offline install — no separate download step (unlike sandbox images).

## What changed

**Bundled content (embedder-side only — the harness still just consumes two dir paths):**
- Build packs `skills/`+`templates/` into a dependency-free `content.pack`, embedded in every target, with a deterministic `INFLEXA_CONTENT_HASH` baked in.
- On first run (release builds), boot extracts to `<dataDir>/inflexa/content/<hash>/{skills,templates}` (a peer of `refs/`/`models/`, not config) via an atomic temp→`rename`, then `skillsDir`/`templatesDir` resolve there.
- **Updates ride the install:** a new binary bakes a new content hash → fresh extract on first run → stale dirs pruned. No separate step, no version-compat matrix.
- Dev is unchanged: repo-root trees via `import.meta.dir`; the embedded-archive machinery is reached only through a release-gated dynamic import, so a dev checkout never needs `content.pack`.
- New `content_materialize_failed` boot error fails visibly (never a fake/empty dir).

**Also: linux-arm64 as a first-class release target** — build matrix, `install.sh` (dropped the refusal), npm platform package (`assemble.ts` + `release-npm.yml`), and the Homebrew formula (`render.sh` + `on_linux`/`on_arm`).

## Verification

- Built a release binary and confirmed `content.pack` is embedded (grep). Booted with an empty data dir + no config override → extracted all 128 files (117 skills + 11 templates), and boot proceeded **past** the skills/templates gate (proving it saw a populated tree). Warm re-run didn't re-extract; a planted stale hash dir was pruned while a `.tmp-*` was spared.
- `content-pack.ts` unit-tested (pack/unpack/hash/path-escape/truncation), 8/8. Full suite **1074 pass / 0 fail**; typecheck + lint clean.
- Homebrew formula renders with the linux-arm64 block, no leftover placeholders.

## Spec

OpenSpec change `add-bundled-content-assets` (archived) — new `content-assets` capability + `harness-runtime` skills/templates default updated.


krajla:
@claude review

vivere-dally (approved):

PR #129, opened 2026-07-16
feat: sandbox execution on Podman via a configurable Docker-API engine connection

### LINKED ISSUE #123: harness: sandboxes hard-require the Docker socket — podman-compatible engine connection
## Current state (verified against source)

The CLI treats Docker and Podman as interchangeable for everything it runs itself — the container-runtime spec pins `docker | podman` at first need, and the proxy/Postgres compose stack runs on whichever is pinned. Sandboxes do not follow:

- `harness/src/sandbox/create-sandbox.ts` — `SandboxBackendConfig.backend` is `"docker" | "k8s"`; there is no podman-aware path.
- `harness/src/sandbox/docker-client.ts:203` — `const docker = config.docker ?? new Docker()`: a `dockerode` client with default connection (`/var/run/docker.sock`; `docker-modem` honors `DOCKER_HOST` if set, but nothing in the product sets or documents it). The `docker?:` config slot is documented "Injected for tests" and is not reachable from `CreateSandboxClientConfig`.
- `cli/src/modules/harness/runtime.ts` — the composition root hard-codes `env: { backend: "docker", namespace: "" }` (now carries a `TODO(extend)` pointing here).

Two failure modes on a pinned-podman machine:

1. **Podman only (no Docker):** every sandbox step fails at container create — dockerode dials a Docker socket that does not exist, even though the proxy, Postgres, and image pulls all work fine on podman.
2. **Both installed, podman pinned (split-brain):** the CLI's `ensureSandboxImage` pre-pulls the sandbox image via the *pinned* runtime (`cli/src/modules/harness/profile.ts:156`), so the image lands in **podman's** store — while the harness creates containers against **Docker**, which 404s on the image (docker-client never pulls; it relies on the embedder's pre-pull). The user did everything right and still can't run a step.

## Why this is a harness capability, not a CLI flag

Per the monorepo boundary rule, the harness owns its concepts and the embedder only supplies values at the composition root. The CLI cannot fix this by itself: the connection target of the sandbox engine is not part of the harness's configuration surface at all today.

## Proposed shape

Podman ships a Docker-compatible REST API (`podman system service` on Linux; the `podman machine` socket on macOS), and the backend is already a dockerode wrapper — so the smallest sound design is to treat the backend as **"docker-API engine"** rather than "Docker the product":

**1. Connection seam (harness).** Add an explicit engine-connection option to `CreateSandboxClientConfig` — e.g. `engineSocketPath?: string` (or a full `DockerOptions` slice) — threaded into `createDockerSandboxOps` in place of the bare `new Docker()`. Default unchanged (Docker socket), so existing embedders and managed deployments are untouched. `backend` stays `"docker" | "k8s"`; podman is a *connection*, not a third backend, unless divergence (below) forces per-engine behavior.

**2. Compatibility validation (harness).** The docker-client makes assumptions worth checking against podman's compat API before declaring support:
- **Poll-mode egress firewall** — containers are created as root with `CAP_NET_ADMIN` and the entrypoint installs `iptables -P OUTPUT DROP` before dropping to uid 1000. Rootful podman (the macOS machine default) should honor this; **rootless** podman networking (pasta/slirp4netns) is the main risk and may need poll mode gated or the firewall strategy adapted.
- **Port publish to `127.0.0.1`** — supported by podman, but worth an explicit test through the machine VM on macOS.
- **Reaper/adoption label filters** — `listContainers({ filters: { label: [...] } })` and name-conflict adoption must behave identically under the compat API.
- **Bind mounts + SELinux** — `HostConfig.Binds` without `:z` relabeling breaks on SELinux-enforcing hosts (Fedora et al.); the CLI already models this divergence for its own mounts (per-runtime `:z` in the container-runtime spec), and the harness mount plan would need the same knob if first-class podman-on-Linux is in scope.
- **`platform` forcing** — the compat API accepts it; verify the arch-mismatch failure mode still fails loudly at create.

**3. Composition-root wiring (cli, after 1 ships).** When the pinned runtime is podman, resolve the machine's Docker-compatible socket (`podman machine inspect --format '{{ .ConnectionInfo.PodmanSocket.Path }}'` on macOS; `podman info --format '{{ .Host.RemoteSocket.Path }}'` on Linux, starting `podman system service` if needed) and pass it as the engine connection. Pinned docker keeps today's default. This also dissolves the split-brain: image pull and container create then target the same engine by construction.

**4. Failure UX until then.** With the connection seam in place but podman-on-Linux-rootless unvalidated, the honest interim is a launch-gate error naming the constraint ("sandbox steps need the Docker engine / a podman machine socket") instead of a dockerode ECONNREFUSED surfacing mid-run.

## Non-goals

- A third `backend` id, new pull logic in the harness (pre-pull stays embedder-owned), or compose changes — the CLI side already runs fine on podman.

vivere-dally:
## What & why

Sandboxes hard-required the Docker socket: the harness constructed its `dockerode` client with default options, so on a podman-pinned machine every sandbox step failed at container create even though the proxy, Postgres, and the image pre-pull all ran on podman — and with both engines installed, the image landed in podman's store while containers were created against Docker (split-brain). Podman serves the same Docker REST API on a different socket, so this PR treats the backend as a *docker-API engine* rather than "Docker the product": podman becomes a **connection**, not a third backend.

Closes #123

**Harness** (`feat(harness)`, spec: `docker-sandbox-provider`):
- `engineSocketPath?: string` on `CreateSandboxClientConfig`/`DockerClientConfig`, threaded to the single `dockerode` construction site. Unset is byte-identical to today (`DOCKER_HOST`, then the default socket).
- Recovery adoption (`createOrAdopt`) no longer keys on HTTP 409: any create failure inspects the checkpointed name and runs the existing owner-guard on a standing container. Podman's compat API answers duplicate-name creates with **500**, and adoption must not care; no standing container returns the original create error.
- `stepTreeAccess?: "world-writable"`: chmods the pre-created step tree (step dir + subdirs, explicit `chmod`, replay-safe) so the uid-1000 workload can write through engines whose bind mounts preserve host ownership honestly (podman machine's virtiofs); Docker Desktop masks the mismatch and keeps default modes.

**CLI** (`feat(cli)`, specs: `container-runtime`, `harness-runtime`):
- `resolveEngineSocket` on the runtime-descriptor surface: docker → no socket; podman → the machine's compat socket (macOS) or `podman info`'s REST socket gated on existence (Linux). Resolved every boot, never persisted (the macOS socket lives under `$TMPDIR` and moves across machine restarts).
- Boot resolves the pin + socket among the pre-flight gates and wires `createSandboxClient` accordingly; `stepTreeAccess` is passed only under a podman pin; a docker pin produces byte-identical config to before. Failed resolution is a dedicated boot error naming the remediation (`podman machine start` / `systemctl --user enable --now podman.socket`) instead of a dockerode `ECONNREFUSED` mid-run.

**Verification**: harness suite 1228 pass / 0 fail — run through podman itself (`DOCKER_HOST` → compat socket, `TESTCONTAINERS_RYUK_DISABLED=true`); cli suite 1157 pass / 0 fail; live smoke on a rootful podman machine with the current `sandbox-python:latest`: sandbox created healthy through the resolved socket, uid-1000 artifact write landed on the host through the world-writable step tree, teardown clean. Design rationale (including the rejected `engine: "docker" | "podman"` enum and the rejected 409/500 status matching) is recorded in the archived changes' `design.md`.

Out of scope (recorded in the change docs): rootless podman on Linux (pasta/slirp4netns) is unvalidated — reachable via the seam, gated by the boot error until validated; SELinux `:z` relabeling for the harness mount plan on enforcing Linux hosts.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Documentation updated if user-facing behavior changed (OpenSpec specs synced in both subsystems; no command/flag surface changed).

- [x] **Methods:** not applicable — no analytical method change.
- [x] **Sandbox:** the security model is preserved: same unprivileged uid-1000 workload, `CapDrop ALL` + the four entrypoint-only CapAdds, `no-new-privileges`, in-container egress firewall in poll mode, signature-authenticated exec endpoints, read-only analysis tree, resource limits. **One deliberate loosening, called out:** under a podman pin the per-step write tree (`runs/{runId}/{stepId}` + subdirs only) is created world-writable, because podman machine's virtiofs preserves host uid/modes and the uid-1000 workload could otherwise not write its own output directory. Scope is the step output tree only (created to receive sandbox writes); the read-only mounts are untouched; docker-pinned machines keep today's modes.
- [x] **Provenance:** prior analyses remain reproducible — no change to provenance signing, artifact registration, or the exec protocol.

### DIFF
--- a/harness/src/sandbox/create-sandbox.ts
+++ b/harness/src/sandbox/create-sandbox.ts
@@ -81,16 +81,18 @@ export interface CreateSandboxClientConfig {
      */
     engineSocketPath?: string;
     /**
-     * Docker: how the pre-created step write tree is made writable by the
-     * uid-1000 sandbox workload. `world-writable` chmods the step dir and its
-     * subdirs so a workload write lands through engines whose bind mounts
-     * preserve host ownership honestly — podman machine's virtiofs presents the
-     * embedder user's real uid and modes, which the uid-1000 workload cannot
-     * write. Docker Desktop's file-sharing layer masks that mismatch, so it
-     * stays unset there. A single-valued union (not a boolean) so the call site
-     * reads as policy and future strategies extend it without a breaking change.
+     * Docker: the engine's bind mounts expose honest host file ownership to
+     * the container — podman machine's virtiofs and native Linux binds present
+     * the embedder user's real uid and modes, which the uid-1000 sandbox
+     * workload cannot write; Docker Desktop's file-sharing layer masks that
+     * mismatch, so it stays unset there. When set, the client loosens each
+     * step's pre-created write tree so workload writes land. The field names
+     * the engine fact rather than the remediation, so how the harness
+     * compensates can evolve without changing the embedder surface; a
+     * single-valued union (not a boolean) so future ownership classes extend
+     * it without a breaking change.
      */
-    stepTreeAccess?: "world-writable";
+    engineBindOwnership?: "host-preserved";
     /** K8s: PVC claim backing the session PVC the workspace roots live under. */
     sessionPvc?: string;
     /**
@@ -237,7 +239,16 @@ export function createSandboxClient(config: CreateSandboxClientConfig): SandboxC
 
     return {
         createSandbox: async (meta, identity) => {
-            await precreateStepTree({ resolveWorkspaceRoot: config.resolveWorkspaceRoot, stepTreeAccess: config.stepTreeAccess }, meta);
+            // The config states an engine fact (binds preserve host ownership);
+            // which remediation that fact demands — today, a world-writable step
+            // write tree — is harness-owned and chosen here, not by the embedder.
+            await precreateStepTree(
+                {
+                    resolveWorkspaceRoot: config.resolveWorkspaceRoot,
+                    stepTreeAccess: config.engineBindOwnership === "host-preserved" ? "world-writable" : undefined,
+                },
+                meta,
+            );
             // Every caller must declare resources — a sandbox with no cpu/memory
             // request is a semantic error, not something to paper over with a
             // default (a DBOS replay of a pre-resources workflow input lands here).
--- a/harness/src/sandbox/create-sandbox.ts
+++ b/harness/src/sandbox/create-sandbox.ts
@@ -12,7 +12,7 @@
  * `awaitExec` are backend-agnostic — they only need the SandboxRef.
  */
 
-import { mkdir } from "node:fs/promises";
+import { chmod, mkdir } from "node:fs/promises";
 import { join } from "node:path";
 
 import type { V1Toleration } from "@kubernetes/client-node";
@@ -74,6 +74,23 @@ export interface CreateSandboxClientConfig {
     refStorePath?: string;
     /** Docker: force the container platform (e.g. `linux/amd64`) so the sandbox matches the mounted lib store's arch. */
     platform?: string;
+    /**
+     * Docker: unix socket of the Docker-API engine to dial — a Docker daemon
+     * socket or a podman Docker-compat socket. Unset preserves dockerode's
+     * default resolution (`DOCKER_HOST`, then the default Docker socket).
+     */
+    engineSocketPath?: string;
+    /**
+     * Docker: how the pre-created step write tree is made writable by the
+     * uid-1000 sandbox workload. `world-writable` chmods the step dir and its
+     * subdirs so a workload write lands through engines whose bind mounts
+     * preserve host ownership honestly — podman machine's virtiofs presents the
+     * embedder user's real uid and modes, which the uid-1000 workload cannot
+     * write. Docker Desktop's file-sharing layer masks that mismatch, so it
+     * stays unset there. A single-valued union (not a boolean) so the call site
+     * reads as policy and future strategies extend it without a breaking change.
+     */
+    stepTreeAccess?: "world-writable";
     /** K8s: PVC claim backing the session PVC the workspace roots live under. */
     sessionPvc?: string;
     /**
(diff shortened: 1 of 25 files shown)


vivere-dally:
@claude review

harness/src/sandbox/create-sandbox.ts
```diff
@@ -74,6 +74,23 @@ export interface CreateSandboxClientConfig {
...
+     * Docker: how the pre-created step write tree is made writable by the
+     * uid-1000 sandbox workload. `world-writable` chmods the step dir and its
+     * subdirs so a workload write lands through engines whose bind mounts
+     * preserve host ownership honestly — podman machine's virtiofs presents the
+     * embedder user's real uid and modes, which the uid-1000 workload cannot
+     * write. Docker Desktop's file-sharing layer masks that mismatch, so it
+     * stays unset there. A single-valued union (not a boolean) so the call site
+     * reads as policy and future strategies extend it without a breaking change.
+     */
+    stepTreeAccess?: "world-writable";
```
krajla: How does this actually work with steps running in parallel, won't there be a race for permissions setting?
Can this be avoided by changing anything in the sandbox base?

vivere-dally: Good questions — one at a time:

**Race with parallel steps:** there isn't one, by construction. `precreateStepTree` chmods only the step's own tree — `runs/{runId}/{stepId}/` plus its subdirs — and step ids are unique, so parallel steps touch disjoint paths. The shared parents (`runs/{runId}`, the analysis root) are only ever `mkdir(recursive: true)` (idempotent, concurrency-safe) and never chmodded. Each step's chmod also runs before its own container is created, so no workload exists yet that could observe the tree mid-change.

**Avoiding it in the sandbox base:** we can't, by construction. The confined entrypoint runs with `CapDrop: ALL` plus only NET_ADMIN/SETUID/SETGID/SETPCAP — no CAP_FOWNER / CAP_CHOWN / CAP_DAC_OVERRIDE — so even the pre-drop container root can't chmod the mounted tree (we hit exactly this while probing: `apk` breaks in the confined container for lack of CHOWN/FOWNER). And that's only the shallow reason: honest bind ownership is precisely the property that the container side has no authority over host-owned files — through rootless podman's userns there's an ownership wall regardless of caps. The host process is the file owner, so host-side is the only actor guaranteed able to loosen the modes. (Podman's idiomatic escape, `--userns=keep-id`, is engine-specific run config, not an image change, and would run the workload as the host uid — breaking the image's uid-1000 identity and the setpriv drop.)

**On the config terminology (from our chat):** agreed — the field was mechanism-shaped: it made the embedder dictate what the harness does ("make the step tree world-writable") instead of stating what the embedder knows. Renamed to `engineBindOwnership: "host-preserved"`: the embedder declares an engine fact (binds expose honest host ownership — true of podman machine's virtiofs and native Linux binds; Docker Desktop masks it), and the per-step-tree chmod becomes a harness-internal remediation, free to change without touching the config surface. That also covers the "it makes an assumption" point — the assumption now lives inside the harness where it can evolve.

We also evaluated auto-detecting instead of any field (podman self-identifies on `/version` as Libpod), but engine identity isn't the real fact — Linux Docker preserves ownership too, and rootless userns remapping breaks any uid-comparison shortcut — and a misdetection would surface as EACCES deep inside a durable workflow instead of an explicit boot-time decision. The embedder already knows this authoritatively (it pinned the runtime and resolved its socket), so it states it.

krajla (dismissed):

PR #131, opened 2026-07-16
feat(refs): expand catalog, grouped picker, URL-only integrity model

krajla:
## What

Two things, in the reference-data catalog (`harness/`) and the `refs` CLI (`cli/`):

1. **Expand the catalog** from 9 → 25 datasets and make the download surface navigable.
2. **Collapse the integrity model** to a single URL-only, trust-on-first-use scheme.

## Catalog expansion (16 new datasets)

| Group | Added |
|-|-|
| gene-identifiers | `ncbi-gene2ensembl`, `ncbi-gene2refseq`, `uniprot-idmapping-{human,mouse,rat}` |
| pathways | `wikipathways-{mouse,rat}`, `msigdb-hallmark-{human,mouse}` |
| normal-expression | `hpa-proteinatlas` |
| cell-typing | `celltypist-{pan-fetal,covid19}`, `panglaodb-markers`, `azimuth-{pbmc,tonsil}` |
| pharmacogenomics *(new group)* | `pharmcat-grch38-fasta` |

Large or domain-specific downloads (the all-species NCBI mapping tables, the tonsil reference, the PharmCAT genome bundle) are marked opt-in so a default `setup` never silently pulls hundreds of MB.

## Grouped download UX

- The interactive picker uses clack `groupMultiselect`, grouped by category (toggling a group header selects everything under it).
- `refs list` prints category headers instead of one flat wall, and `refs list --urls` reveals the exact upstream URL of every artifact.

## URL-only integrity (trust-on-first-use)

A catalog artifact is now just `{ path, url }` — **no `pinned`/`unpinned` class, no checked-in size or digest.** Rationale:

- The old split (immutable → pinned+digest, mutable → trust-on-first-use) applied scrutiny **unevenly** — half the catalog was already digest-free — and it put on us the burden of downloading+hashing every immutable file and keeping that digest in sync.
- TLS already authenticates the publisher; the **install receipt** (bytes + sha256 observed from the user's *own* download) already lets `refs verify` catch post-install corruption/tampering. The only thing a checked-in digest added was "these match the bytes a maintainer reviewed" — marginal for public, re-downloadable, sandbox-loaded reference data.

Result: **adding a source is now one line — a URL.** Nothing to hash, nothing to keep in sync, no digest that goes stale when a `current` upstream rebuilds. Old receipts carrying the removed `integrity` field still parse (the key is ignored), so existing installs are not invalidated.

Mechanically: the discriminated union / `bytes` / `sha256` / `ReferenceIntegrity` are gone; `store.ts` always fetches fresh and records observed bytes+digest; `verify` compares against the receipt; `referenceDownloadBytes` → `referenceDownloadEstimate`.

## Spec

`harness/openspec/specs/reference-data-catalog/spec.md` rewritten to the uniform model (`openspec validate` passes).

## Validation

- **harness**: `tsc` build clean; reference-data tests 32 pass; spec valid.
- **cli**: `typecheck` clean; `refs` + `cli.test.ts` 40 pass; `lint` clean.
- Smoke: `refs list` / `refs list --urls` render all 6 groups correctly.

## Deferred (follow-ups, now trivial — just URLs)

ProjecTILs (needs a figshare `sourceUrl` host), Ensembl Compara orthologs, Tabula Sapiens v2, the Group-3 query-URL sources (OmniPath/DoRothEA/PROGENy/LINCS). CellMarker2 is blocked (http-only). Curated JSON (safety panel, signatures, markers) is repo-bundled, not a download.

### DIFF
--- a/cli/src/cli/cli.test.ts
+++ b/cli/src/cli/cli.test.ts
@@ -32,15 +32,16 @@ describe("inflexa help & usage (e2e)", () => {
         expect(result.stderr).toContain("Unknown reference dataset");
     });
 
-    test("refs list explains custom content, integrity, and catalog contributions", () => {
+    test("refs list explains custom content, trust-on-first-use integrity, and catalog contributions", () => {
         const result = runCli(["refs", "list"]);
         expect(result.exitCode).toBe(0);
         expect(result.stdout).toContain("Add arbitrary references under");
         expect(result.stdout).toContain("Open a PR");
-        // Nothing is re-hosted: each dataset states the integrity guarantee its own upstream can offer.
-        expect(result.stdout).toContain("Integrity: pinned — verified against the checksums in the catalog");
-        expect(result.stdout).toContain("Integrity: unpinned — upstream is rebuilt in place; verified against what you downloaded");
-        expect(result.stdout).toContain("nothing is mirrored or re-hosted here");
+        // Nothing is mirrored, re-hosted, or checksum-pinned: integrity is trust-on-first-use, checked
+        // against the copy you downloaded rather than a catalog digest.
+        expect(result.stdout).toContain("nothing is mirrored, re-hosted, or checksum-pinned here");
+        expect(result.stdout).toContain("Integrity is trust-on-first-use: `inflexa refs verify` checks each installed file against the copy you downloaded.");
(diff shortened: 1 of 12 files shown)


krajla:
Superseded by the two-PR split so the harness lands and publishes first: #133 (harness, @inflexa-ai/harness 0.3.0) then #134 (cli, pins 0.3.0).

Issue #71, opened 2026-07-10
harness: residual read-side symlink TOCTOU in WorkspaceFilesystem read seam (close with openat2 RESOLVE_BENEATH)


vivere-dally:
## Summary

The deterministic symlink confused-deputy attack on the workspace seams (plant an
escaping `ln -s`, have a host-side operation follow it) is **fixed** on both the read
and write surfaces. A **narrow read-side TOCTOU race remains**: the read seam checks
containment and opens the file in two separate steps, and the read open does not use
`O_NOFOLLOW`. A symlink swapped into the path in the window between check and open is
followed, re-opening the read-exfiltration hole for the duration of the race.

Background and the full defensive writeup:
[`harness/docs/symlink-following-shared-mounts.md`](../blob/main/harness/docs/symlink-following-shared-mounts.md)
(§ "How this is fixed in this codebase" → "Not yet applied").

## What is already fixed (not this issue)

- **Read** — every read tool (`read_file`, `grep`, `list`, `stat`) funnels through
  `confinedResolve` (`harness/src/workspace/filesystem.ts:150`) → `classifyWithinRoot`
  (`harness/src/lib/fs-helpers.ts:103`), which `realpath`s the path and drops an
  `escaped` verdict to `out_of_scope` (`filesystem.ts:160`). A **persistent** escaping
  symlink is rejected every time.
- **Write** — `writeFileWithinRoot` (`harness/src/lib/fs-helpers.ts:126`) `realpath`s
  the parent and opens the leaf with `O_NOFOLLOW` (`fs-helpers.ts:151`), which is
  atomic on the final component.
- Shipped in `c43e5c2` (reads) and `48b7001` (writes).

## The residual bug

The read path validates and uses the path in two separate syscalls:

1. `classifyWithinRoot` `realpath`s a **snapshot** of the path
   (`harness/src/lib/fs-helpers.ts:113`) — correct at that instant.
2. The read then re-opens the **lexical** path (`resolved.absolute`) with a plain
   handle — `fsOpen(absolute, "r")` (`filesystem.ts:268`, `:358`) and
   `createReadStream(absolute)` (`:306`) — **no `O_NOFOLLOW`**.

Between (1) and (2), in-container code that shares the same on-disk tree can replace
the leaf **or an intermediate directory** with a symlink pointing outside the
workspace root. The open in (2) follows it and the host returns off-tree file content
into the model context. This is [CWE-367](https://cwe.mitre.org/data/definitions/367.html),
[...]

PR #132, opened 2026-07-16
fix(harness): close read-side symlink TOCTOU in workspace read seam

krajla:
## Summary

Closes the residual read-side symlink TOCTOU (CWE-367) in the `WorkspaceFilesystem` read seam described in #71.

The seam validated a path's containment with a `realpath` snapshot (`classifyWithinRoot`) and then re-opened the **lexical** path in a separate, plain syscall with **no `O_NOFOLLOW`**. A symlink swapped into the leaf between the check and the open was followed off-tree, leaking host-file content into the model context. (The write seam was already safe — it opens with `O_NOFOLLOW`.)

## What changed

`harness/src/workspace/filesystem.ts`:

- **`openReadNoFollow()`** — opens `O_RDONLY | O_NOFOLLOW`. A symlinked final component fails `ELOOP`, which maps to the `out_of_scope` data variant, so the bytes read always come from a non-symlink leaf. There is no check/use gap: the refusal is in the same syscall that opens.
- All three content readers route through it: `readWithCap` (plain + capped), `readHeadLines` (now streams via `fh.createReadStream({ autoClose: false })`), `readTailLines`. `grep` inherits the guard via `read_file`.
- **Metadata surface hardened to match:** `stat` uses `lstat` and returns `out_of_scope` for a symlink leaf; `list` `lstat`-guards the directory leaf and `lstat`s each entry's size, so a symlink is never followed to its (possibly off-tree) target.

## Scope / residuals

- **Behavior change:** the read tools no longer follow *any* leaf symlink, including in-tree ones that used to be readable. Pure Node cannot express "follow only if it stays in-tree" (that needs `openat2(RESOLVE_BENEATH)`), so the atomic, portable choice is to refuse leaf symlinks outright. **Hard-linked staged inputs are unaffected** (a hard link is not a symlink), and the one production symlink — `report-runner.ts`'s `assets → ../assets` — is an *intermediate directory*, which `O_NOFOLLOW` does not touch, so report previews keep working.
- **Remaining (documented):** the intermediate-directory read race and the `list` readdir race can't be atomically closed in pure Node (no fd-based `openat2`/`readdir`). The persistent case of both stays covered by the upstream `realpath` check. A kernel-atomic close (`openat2 RESOLVE_BENEATH`) or the seccomp `symlink`/`symlinkat` deny remain optional follow-ups, out of scope here by design (host-side, all-environments, no native addon).

## Tests

`harness/src/workspace/filesystem.test.ts` — the former "reads an in-tree symlink" case is flipped: a symlink whose target is **in-tree** (so it passes the `realpath` containment check) is now refused at open across all read modes (plain, `maxBytes`, `headLines`, `tailLines`) plus `stat`, and a symlinked directory is refused by `list`. This is the deterministic equivalent of the "swap a validated path for a symlink between check and open" assertion the issue asks for. The hard-linked-input case still asserts `ok`.

- `tsc -p tsconfig.json` — 0 errors
- `bun test` (workspace + tools/workspace + fs-helpers) — 87 pass, 0 fail
- Prettier + ESLint (incl. `must-use-result`) clean

## Note

The acceptance-criteria doc item (`harness/docs/symlink-following-shared-mounts.md`) is moot — that file isn't in the tree (harness design lives in the OpenSpec specs; there is no `docs/`).

### DIFF
--- a/harness/src/workspace/filesystem.test.ts
+++ b/harness/src/workspace/filesystem.test.ts
@@ -106,15 +106,35 @@ describe("createWorkspaceFilesystem", () => {
         expect((await fs.stat({ session, path: "runs/r1/s1/leak.txt" }))._unsafeUnwrap().kind).toBe("out_of_scope");
     });
 
-    it("still reads an in-tree symlink and a hard-linked input", async () => {
-        // In-tree symlink → allowed (realpath stays in the tree).
+    it("refuses an in-tree symlink leaf (O_NOFOLLOW) but reads a hard-linked input", async () => {
+        // The read open uses O_NOFOLLOW, so it refuses EVERY leaf symlink — even
+        // one whose target stays in-tree and would pass the realpath containment
+        // check. This is the atomic guard that closes the check/open race: a leaf
+        // symlink swapped in after `classifyWithinRoot` validates the path is still
+        // refused at open, returning out_of_scope, never followed content.
         const real = join(sessions, ANALYSIS, "data", "inputs", "real.csv");
-        await writeFile(real, "gene\nBRCA1\n");
+        await writeFile(real, "gene\nBRCA1\nBRCA2\n");
         await symlink(real, join(sessions, ANALYSIS, "data", "inputs", "alias.csv"));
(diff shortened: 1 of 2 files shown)


krajla:
@claude review

vivere-dally (approved):

PR #135, opened 2026-07-16
feat(cli): adopt ecosystem provider env vars at direct-connection setup

krajla:
## Why

A `direct`-mode install almost always runs on a machine already provisioned for Claude Code / the SDKs — it carries `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, and the matching `*_BASE_URL` gateways. Today the CLI ignores all of them: the key must be re-exported as `INFLEXA_MODEL_API_KEY` and the endpoint re-typed at `inflexa setup`. This teaches the direct path to recognize and adopt what's already there.

## What changed

- **`resolveModelApiKey(provider)`** in `lib/env.ts` (kept the sole `process.env` reader): env-only resolution — `INFLEXA_MODEL_API_KEY` first (explicit override), then the provider-conventional var (`ANTHROPIC_API_KEY` for `anthropic`, `OPENAI_API_KEY` otherwise). The key is **never** written to config, telemetry, logs, or provenance.
- **`setup --connection direct` detects `ANTHROPIC_*` / `OPENAI_*`** and offers an editable pre-fill, copying only `{provider, baseURL, protocol}` into `models.connection`. `baseURL` is normalized to the `/v1`-terminated form the wire layer needs — `ANTHROPIC_BASE_URL` is a bare root, `OPENAI_BASE_URL` usually already `/v1`.
- **Non-TTY `setup --connection direct` self-configures** from a detected env with no prompts (anthropic-first tiebreak); keeps the "needs a terminal" failure when nothing is detectable.
- **Boot's direct key-missing error** now names both `INFLEXA_MODEL_API_KEY` and the provider var it tried, plus the config path.
- **`--help`** documents the fallback via `modelConnectionEnvDoc` (presence, never the value).

## Design decisions

Detect-and-adopt **at setup** (not a live boot-time env read — avoids silently hijacking an intended `cliproxy` setup); **derive** the key env from the already-stored provider (no new config field, secret never copied); **snapshot** the non-secret bits into config (self-contained, drives the status surface). Full rationale + alternatives in the change's `design.md`.

## Scope

**Deferred (documented, not implemented):** `ANTHROPIC_AUTH_TOKEN` — Anthropic-wire + Bearer needs a harness custom-auth capability the CLI can't supply alone (workaround: `protocol: openai-compatible`); Bedrock/Vertex — no direct-mode HTTP signer. **CLI-only** — `harness/src` untouched. The existing security posture is preserved exactly.

## Spec

OpenSpec change **`adopt-provider-env-vars`** (strict-valid): `model-connection` delta — MODIFIED the env-only-secret requirement (provider-derived fallback chain) + ADDED the setup-detection/adoption requirement. Proposal, design, and tasks included.

## Testing

- `bun run typecheck` clean, `bun run lint` clean, full suite 1169 pass / 1 skip. (The 1 unrelated failure, `embedding/local-provider.test.ts`, reproduces identically on clean `main` — a WSL2 loopback-proxy quirk, untouched here.)
- **Verified end-to-end against the real CLI** (sandboxed XDG dirs): drove the non-TTY `setup --connection direct` self-configure for both ecosystems and confirmed the written `config.json` carries the normalized connection and **no key material**; exercised the resolver precedence, the asymmetric `/v1` normalization, and presence-only detection against live `process.env`; confirmed the `--help` env doc prints no secret.


krajla:
@claude review

PR #147, opened 2026-07-17
feat(cli): embed bge-small model into the binary for offline local-embeddings setup

vivere-dally:
## What & why

A client running behind a corporate proxy cannot reach huggingface.co, so `inflexa setup --embeddings local` fails at the model download — even though the compiled binary already carries the llama-server *runtime* as an embedded asset. The bge-small GGUF (~36 MB) was the last network dependency of local-embeddings setup in the compiled binary. This PR embeds it, making local-embeddings setup fully offline for release binaries.

- **Build-time embed**: `scripts/build.ts` gains `ensureModelCached()` beside `ensureLlamaArchiveCached()` — the pinned `bge-small-en-v1.5-q8_0.gguf` is fetched once per build into `.llama-cache/`, verified against the vendored `MODEL_SHA256`, and embedded into **every** target via the same define-gated `import(..., { with: { type: "file" } })` mechanism the llama runtime uses (one platform-independent asset; no per-target selection). The stale-cache sweep keeps the model artifact so a pin bump still fails loudly on a stale embed reference.
- **Source-aware acquisition**: `downloadModel()` → `acquireModel()` in `modules/embedding/setup.ts`, mirroring `llama_runtime.ts`: a compiled binary copies its embedded asset (bunfs-safe byte read, zero network); a source checkout downloads from HuggingFace as before. Both byte sources are SHA-256-verified before the atomic `.part` → final rename, so nothing lands at `env.embeddingModelPath` unverified.
- **One integrity authority**: the pin lives in the new leaf `modules/embedding/model_pin.ts`, shared by the build script and runtime acquisition (a leaf module so the build script doesn't evaluate setup's transitive `@inflexa-ai/harness` graph).
- **Install-context-aware copy**: the setup picker and spinners no longer claim the compiled binary "downloads" the model.
- **Runtime-gate reorder**: `inflexa setup --embeddings <mode>` runs the embedding step *before* the container-runtime probe, so an air-gapped host without a ready Docker/Podman can still configure embeddings non-interactively. The interactive flow keeps the embedding question in its spec'd position (after provider auth); the rest of setup still requires a runtime and fails afterward as before. The step runs at most once per invocation.

Considered and rejected: committing the model via Git LFS — `actions/checkout` skips LFS blobs by default so CI would stay cheap, but bandwidth is billed to the repo owner (~28 downloads/month on the free tier at this size), contributors would need `git-lfs`, and it would add a second artifact-distribution mechanism beside the established build-time one.

**Verification**: typecheck + lint clean; 159 pass / 0 fail across the infra + embedding suites (incl. new tests for source-aware acquisition and the gate reorder). A host `bun run build` fetched and hash-verified the model, the smoke test passed, and a byte-level check found a 128-byte needle from the model's middle inside the compiled binary. **Offline end-to-end proof**: the compiled binary ran `setup --embeddings local` under sandboxed XDG dirs behind an unroutable proxy (huggingface.co unreachable) — the embedding step ran ahead of the runtime probe, installed the bundled model (sha256 `ec38e8da…c2f514`, exact match to the pin), verified 384-dim vectors through the loopback sidecar (the provider's NO_PROXY handling held under the proxy), and wrote `embedding.mode = "local"` — zero network. Size impact: ~36 MB per target binary.

Spec deltas are synced into `openspec/specs/local-embeddings/` and the change is archived (`openspec/changes/archive/2026-07-17-embed-embedding-model-in-binary/`), both in this PR.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/cli/.gitignore
+++ b/cli/.gitignore
@@ -1,3 +1,3 @@
-# Pinned llama-server release archives cached by scripts/build.ts for the local-embeddings sidecar
-# runtime. Downloaded on demand (hash-verified against the vendored pin), never committed.
+# Build-time embedded artifacts cached by scripts/build.ts: the pinned llama-server release archives and
+# the bge-small embedding-model GGUF. Downloaded on demand (hash-verified against the vendored pins), never committed.
 .llama-cache/
--- a/cli/openspec/changes/archive/2026-07-17-embed-embedding-model-in-binary/.openspec.yaml
+++ b/cli/openspec/changes/archive/2026-07-17-embed-embedding-model-in-binary/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-17
--- a/cli/openspec/changes/archive/2026-07-17-embed-embedding-model-in-binary/design.md
+++ b/cli/openspec/changes/archive/2026-07-17-embed-embedding-model-in-binary/design.md
@@ -1,0 +1,58 @@
+## Context
+
+Local-embeddings setup has two artifacts to acquire: the `llama-server` sidecar runtime and the bge-small GGUF model. The runtime is already source-aware (`llama_runtime.ts`): compiled binaries materialize it from a build-time embedded asset with no network; source checkouts download the pinned release. The model is not — `downloadModel()` in `setup.ts` always fetches from huggingface.co, which fails for users behind egress-restricted corporate proxies even when they run the official binary. `scripts/build.ts` already owns the whole embed pipeline: an out-of-git `.llama-cache/`, vendored-SHA-256 verification with loud build failure, a stale-artifact sweep, and define-gated `import(..., { with: { type: "file" } })` embedding.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- `inflexa setup --embeddings local` in a compiled binary completes with zero network access.
+- One integrity story: the same vendored `MODEL_SHA256` verifies the build-time fetch, the embedded copy, and the from-source download; nothing lands at `env.embeddingModelPath` unverified.
+- Maximal reuse of the existing llama-runtime embed pattern — no new mechanism, cache, or define where an existing one serves.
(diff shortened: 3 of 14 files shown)


vivere-dally:
@claude review

vivere-dally:
Thanks for the review — both findings addressed:

**1. `local-embeddings/spec.md` stale (medium) — already resolved in this PR.** That review ran against an earlier push; `a93a60f` syncs `cli/openspec/specs/local-embeddings/spec.md`. The "Embedding setup downloads and verifies the model on opt-in" requirement is now source-aware (compiled binary copies the embedded asset with no network; source checkout downloads the pinned revision; both verified against the same SHA-256), with dedicated compiled-binary and source-checkout scenarios mirroring the runtime's, plus a new "Embedding model is a build-time embedded asset" requirement and the preselected-before-gate scenarios. No spec text left describing unconditional download.

**2. `download_failed` misnomer (nit) — fixed.** Renamed `EmbeddingSetupError`'s `download_failed` discriminant to `acquire_failed`, matching the `downloadModel` → `acquireModel` rename, since the variant now also covers the "binary embedded no model" and "embedded asset corrupt" cases — neither a download. Scoped to `EmbeddingSetupError` in `setup.ts` (+ its two test assertions); the independent `download_failed` discriminants in `llama_runtime.ts` and `refs/*` are genuine downloads and untouched. `typecheck`, `lint`, and the embedding + infra suites are green (112 pass / 0 fail).

Also calibrated the bump-procedure docs (`model_pin.ts` + the `build.ts` sweep comment): the stale-cache sweep only removes a superseded model file when the *filename* changes; since `MODEL_ARTIFACT` carries no revision, a revision-only bump keeps the same filename and `ensureModelCached`'s cache re-hash — not the sweep — is what catches a stale model.

radutanasa (approved):

PR #149, opened 2026-07-18
fix(harness): stop a probe submission from latching the run synthesis

krajla:
## What happened

A run synthesizer produced a good synthesis — IPRES/stromal-ECM signature in non-responders, cited PMIDs, a negative AUC result against a permutation null — and shipped an empty placeholder instead. DBOS recorded the run as a clean success, so nothing surfaced as a failure anywhere.

## Root cause

The model emitted **complete, well-formed JSON** inside genuine native tool calls, with XML-style function-call markup attached. Six of nine attempts arrived as a valid synthesis followed by a literal `</parameter>\n</invoke>`; a seventh was the same artifact from the other side, a bare leading `<parameter name="runId">` tag. That trailing 24-character wrapper is the entire defect — it made the argument a *string* rather than an object.

`run-agent.ts` rejects at `inputSchema.safeParse` before `execute` runs, so the model never saw a domain error — only `expected object, received string`. It diagnosed that as a size limit and cut its payload:

| attempt | bytes | findings | conclusions |
|-|-|-|-|
| 2 | 22,190 | 5 | 4,060 ch |
| 4 | 18,345 | 5 | 3,033 ch |
| 7 | 11,582 | 5 | 1,586 ch |
| 8 | 1,795 | 1 | 629 ch |

Attempt 2, wrapper stripped, **passes the entire Zod schema** and fails exactly one semantic check — a `keyReferences` PMID never cited by any finding. A one-iteration fix, with the hint already attached, that the model was never shown.

Attempts 5 and 9 were minimal probes to isolate the failure — correct debugging instinct. Probe 5 was structurally valid, so it latched as the terminal outcome and locked out attempt 9 and everything after.

## Changes

**`repairToolInput`** — issue-guided recovery at the loop's tool-input boundary. Fires only where the schema expected an object or array and found a string; strips a recognized function-call wrapper or ` ```json ` fence, re-parses, accepts only when the parsed type matches what the issue expected. Never mutates the input. **Repair only makes the schema reachable** — the repaired value is validated in full and the tool's own semantic checks still run. Explicitly tested: a repaired-but-invalid payload is still rejected.

**Accurate diagnosis when repair can't help.** The hint is parse-driven: a string that parses in full is diagnosed as double encoding; one that parses only after stripping names the surrounding content with a bounded snippet; one that parses under neither gets no hint rather than a speculative one.

**`validate_synthesis`** — a non-terminal dry-run tool mirroring the planner's `validate_plan`, so probing no longer costs the terminal call. It reports schema *and* semantic problems in a single pass; when a check's inputs can't be salvaged it says so explicitly rather than letting a short list read as a clean bill of health.

**Last-valid-wins on `submit_synthesis`** — a later valid submission supersedes an earlier outcome, so a probe can't seal the run. `report_blocker` keeps its first-wins guard, so a blocker can never clobber a synthesis.

The repair and hint work at the loop boundary benefits **every tool in the harness**, not just synthesis.

## Two decisions that reverse work on main — please review

**1. The post-success terminal guard is removed.** `submit_synthesis` previously rejected any call after a recorded outcome, and `it("rejects a second call after success")` covered it. That guard is what made probe 5 permanent. Resubmission is bounded by `maxIterations`, and last-valid-wins still yields exactly one outcome at the end, so the invariant it protected survives in substance.

**2. `RejectionTelemetry` is removed** (`90e583c`). It counted semantic rejections so a blocker's cause could be read after the fact — zero meaning model misjudgment, repeated rejections meaning a give-up against the checks. `validate_synthesis` breaks that reading: the model now iterates on the dry-run tool and submits once, so the counter reports zero precisely on runs where it fought the checks hardest — the value its own doc defines as misjudgment. A diagnostic that inverts its meaning under normal operation is worse than none.

**The run-ledger feature is untouched.** `synthesisStatus`, `synthesisReason`, `SynthesisStatus`, `setRunSynthesisOutcome`, the `synthesis_status` / `synthesis_reason` columns, and their `inspect_run` exposure all survive exactly as on main. Only the in-memory counter and its never-populated `validationAttempts` wire field are gone.

If the telemetry author disagrees, the counter is straightforward to restore over this — it would just need to count `validate_synthesis` rejections too, since that is where iteration now happens.

## Deliberately not done

No `.min()` constraints. A floor on `findings` would create fabrication pressure on a run that legitimately produced one solid finding, and buys no real guarantee — a plausible stub clears any length floor. Main independently reached the same conclusion in its blocker copy ("submit a synthesis with an empty `findings[]` instead"), which is preserved here.

No enforcement that PMIDs originate from `literature_reviewer`. The prompt calls this a terminal rule but nothing checks it, and it can't be checked cleanly today — the reviewer returns free prose, not structured references. Worth a follow-up.

The `openai-compatible` provider branch passes `maxOutputTokens` through with no per-model fallback, so an unset value omits `max_tokens` and lets the server decide; the `anthropic` branch already falls back to the model maximum. Unrelated to this incident, worth hardening separately.

## Verification

Rebased onto `be1035f`. `tsc` clean. Full `bun test` run on this branch and on detached `origin/main`, comparing **sorted failing-test-name lists** rather than counts — byte-identical, 117 each, zero new failures. All are `Could not find a working container runtime strategy`; note `test.yml` deliberately defers those DB/DBOS suites "until a Postgres service is wired into CI", so they run in neither place.

The emitted JSON Schema for `submit_synthesis` is byte-identical to main's (9,864 chars) — the `KeyReferenceSchema` extraction is emission-neutral, so the model's field-by-field contract is unchanged.


krajla:
Two follow-up commits since this was opened:

**`85d8afb`** — the repair path described in the body above. Issue-guided, never mutates, and explicitly tested not to bypass validation.

**`599091a`** — `validate_synthesis` now reports schema *and* semantic problems in a single pass. `fullyValidate` previously short-circuited on schema failure, so a candidate with a structural problem and a bad PMID reported only the structural one — the model would fix it, re-validate, and only then discover the content issues. That is the same wasted-round-trip pattern this PR exists to eliminate, so it was worth closing.

`semanticCheck` is decomposed into six per-concern functions and recomposed, leaving the schema-valid path unchanged. On schema failure each check's inputs are salvaged independently against their own sub-schemas and whichever checks still have what they need are run. A check whose inputs did not survive is skipped rather than guessed, and a trailing issue names each deferred check and opens `This issue list is INCOMPLETE` — a short list can't be mistaken for a clean bill of health.

`KeyReferenceSchema` was extracted from the inline `keyReferences` element so the salvage path can reuse it. The emitted JSON Schema for `submit_synthesis` is byte-identical (9,864 chars before and after, verified by diffing `z.toJSONSchema` output against the previous revision), so the model's field-by-field contract is unchanged.

One existing assertion was narrowed: it required *every* issue on a malformed candidate to be a schema issue, which encoded the short-circuit itself. It now asserts every field-level issue is a schema issue, plus that the deferral notice is present.

krajla:
@claude review

vivere-dally (approved):

PR #156, opened 2026-07-18
feat(cli): deterministic, accessibility-validated default model election

### LINKED ISSUE #144: cliproxy: default chat model is picked non-deterministically from /v1/models, and some advertised models 404
## What happened?

`pickDefaultModel` (`cli/src/modules/proxy/models.ts:82-89`) returns the **first** id containing `"claude"` in the order `/v1/models` happens to serve them. That order is **not stable across calls**, so the resolved default model changes from launch to launch. Meanwhile the proxy advertises Claude models that a given OAuth account cannot actually use — those answer `404 not_found_error`.

The two facts combine into a real user-facing defect: **`resolveModelId` is also the chat's default model resolver** (`cli/src/modules/harness/runtime.ts:542`, memoized once per process via `cliproxyAutoDefault`). A launch that happens to draw an inaccessible model produces a chat session where **every turn 404s** for any agent without an explicit override. A relaunch usually "fixes" it, which makes this look flaky rather than broken.

This is pre-existing and independent of #142 — but #142's launch-time credential probe made it visible, because the probe issues the same request on every launch and reports what it saw:

```
◇  Provider login not verifiable
▲  Could not verify the provider login (HTTP 404) — continuing; chat will surface any real failure.
```

To be clear, the probe is behaving **correctly** here: a 404 is not a credential verdict, so it warns and proceeds without blocking launch. It is the messenger, not the cause. But it is only as reliable as the model it draws — roughly 1 launch in 6-12 currently cannot verify a perfectly healthy credential.

## Steps to reproduce

1. `inflexa setup --provider claude` and let the proxy come up (image `eceasy/cli-proxy-api:latest`, v7.2.77).
2. Ask the proxy the same question repeatedly, exactly as `resolveModelId` → `pickDefaultModel` does — `GET /v1/models`, take the first id containing `claude`, `POST /v1/messages` with `max_tokens: 1`.

Observed across 6 consecutive simulated launches against one unchanged, healthy credential:

```
launch 1: picked claude-3-5-haiku-20241022    -> HTTP 404
launch 2: picked claude-sonnet-4-6            -> HTTP 200
launch 3: picked claude-sonnet-4-6            -> HTTP 200
launch 4: picked claude-haiku-4-5-20251001    -> HTTP 200
launch 5: picked claude-haiku-4-5-20251001    -> HTTP 200
launch 6: picked claude-sonnet-4-6            -> HTTP 200
```

The 404 body names the cause:

```json
{"type":"error","error":{"type":"not_found_error","message":"model: claude-sonnet-4-20250514"}}
```

Two models confirmed inaccessible on this account while advertised by `/v1/models`: `claude-3-5-haiku-20241022`, `claude-sonnet-4-20250514`.

Corroborating detail from the proxy's own request log — the failure is a fast upstream rejection, not a served completion (404 in 708ms; successful completions take 1.4-7.6s):

```
20:48:04  200 | 1.481s | POST "/v1/messages"
20:51:23  404 |  708ms | POST "/v1/messages"     <-- same process, same model list
20:52:04  200 | 3.111s | POST "/v1/messages"
```

The proxy sources its registry from a GitHub-hosted `models.json` (`model_updater.go`, refreshed at boot and every 3h), so the advertised list reflects **what the proxy knows**, not **what the credential can use**:

```
startup model refresh completed from https://raw.githubusercontent.com/router-for-me/models/refs/heads/main/models.json
```

## Expected behavior

1. The resolved default model is the **same** for the same model list — resolution should not depend on serving order.
2. A model the account cannot use is not silently selected as the chat default.

## Suggested direction

- **Make the pick deterministic** (e.g. sort ids before selecting, or rank explicitly within a family). This does not fix an unusable draw on its own, and that is fine: a consistently broken default is diagnosable, a 1-in-12 flake is not. Deterministic first.
- **Stop treating the advertised list as authoritative** for what the credential can serve. Either pin a default model in config, or validate the pick. This is a design question rather than a one-liner, and worth its own discussion — note `modelMatchesProvider` already guards the *provider* family (`runtime.ts:545`) but not *accessibility*.

## Inflexa version

`1222bb3` (built from source; cli `0.3.3`)

## Operating system

macOS (Apple Silicon), podman 5.8.2, `eceasy/cli-proxy-api:latest` v7.2.77 commit c880371

vivere-dally:
## What & why

In cliproxy mode the default chat model was the first `/v1/models` id containing the preferred family substring, in whatever order the proxy served the list — unstable across launches. Worse, the proxy advertises its GitHub-sourced registry rather than what the OAuth credential can actually serve, so roughly 1 launch in 6–12 drew a model that answers 404: every chat turn failed, and the launch probe reported `Provider login not verifiable (HTTP 404)` against a perfectly healthy credential (#144).

The default is now **elected** instead of picked:

- **Deterministic rank** — candidates order by family preference, then the registry's `created` timestamp descending (id ascending as tiebreak). Same list ⇒ same rank, regardless of serving order. Verified live: the proxy passes `created` through `/v1/models`.
- **Accessibility-validated walk** — claude-family candidates are checked through the proxy's `count_tokens` route (unbilled, sub-second; verified live that it forwards upstream with the real credential). Only a definite `not_found_error` advances the walk; anything inconclusive (timeout, 5xx, bare 404) elects the candidate rather than walking past the best model on a flaky signal. All-404 degrades to the top-ranked candidate so the existing probe warn-and-proceed path reports it — election never adds a way for launch to block.
- **Stale-pin warning** — the launch gate now also checks explicit `models.agents.*` / `harness.model` pins and warns (never blocks, never rewrites config) when a pinned model is no longer served for the account.
- **Setup step** — interactive `inflexa setup` offers a default-model select: a preselected **Auto** row labeled with the elected model (accepting it writes nothing, keeping the default adaptive) plus the accessibility-swept list; an explicit pick pins both agents. Non-TTY setups skip it. No hardcoded model ids anywhere.
- **Picker validation** — the TUI model-switch commit validates the selection (listed or free-text) and rejects an inaccessible id in-dialog instead of persisting a broken pin; `openai-compatible` connections commit as before (no cheap validation route exists).

Election lives inside `resolveModelId`, so the launch probe and harness boot share one election through the existing per-process cache — no new plumbing between them.

Verified live against the local proxy: three fresh processes elected the same (newest served) model; the pre-change non-determinism and the spurious not-verifiable warning are gone. One wire fact discovered and recorded in the design's risk register: after repeated 404s the proxy stops forwarding an inaccessible model and answers `503 auth_unavailable` locally — deliberately treated as inconclusive, since the same shape appears transiently on healthy models during proxy boot.

OpenSpec artifacts (proposal, design with decision rationale, spec deltas, tasks) are included under `cli/openspec/` — one new capability (`default-model-election`) and deltas to `model-connection`, `agent-model-selection`, and `cliproxy-credential-health`.

Closes #144

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/cli/CONTEXT.md
+++ b/cli/CONTEXT.md
@@ -11,7 +11,7 @@ The local-first host for the Inflexa product, and the **embedder** of `@inflexa-
 Code is grouped by feature, not by layer — a module owns its logic, its text command actions, and its logic-local types.
 
 - **`auth/`** — Auth0 device flow + `login` / `logout` / `whoami`. Config seeded from `.env` (`INFLEXA_AUTH0_*`).
-- **`proxy/`** — the CLIProxyAPI model helpers (`models.ts`: client key discovery + default-model ranking); the container lifecycle/provisioning lives in `infra/`.
+- **`proxy/`** — the CLIProxyAPI model helpers (`models.ts`: client key discovery + default-model election — deterministic recency rank walked against the unbilled `count_tokens` accessibility check); the container lifecycle/provisioning lives in `infra/`.
 - **`analysis/`** — analysis lifecycle + the `sessions` command (live launch-identity rows; message history is frozen legacy data).
 - **`anchor/`** — invisible folder-identity markers (`.inflexa/id`) and lazy path reconciliation.
 - **`harness/`** — the harness embedder: boots the harness runtime (DBOS, sandbox, providers) and drives the chat turn, the model-free `run --plan` replay engine, data profiling, and the provenance bridge.
--- a/cli/openspec/changes/archive/2026-07-18-add-default-model-election/.openspec.yaml
+++ b/cli/openspec/changes/archive/2026-07-18-add-default-model-election/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-18
--- a/cli/openspec/changes/archive/2026-07-18-add-default-model-election/design.md
+++ b/cli/openspec/changes/archive/2026-07-18-add-default-model-election/design.md
@@ -1,0 +1,84 @@
+# Design: Default Model Election
+
+## Context
+
+Three verified facts (issue #144 + live verification against the running proxy, v7.2.77) drive this design:
+
+1. `/v1/models` serving order is unstable, and `pickDefaultModel` (`proxy/models.ts`) takes the first family match in that order — the resolved default changes across launches.
+2. The advertised list reflects the proxy's GitHub-sourced registry, not the credential: some ids answer `404 not_found_error` on use. The #142 launch probe observes that 404 and discards it — the poisoned pick stays in `cachedModelId`, which chat boot consumes via the shared cache (`runtime.ts` `resolveDefaultModel` → `seams.resolveModel` → the same memoized `resolveModelId`).
+3. `POST /v1/messages/count_tokens` is routed by the proxy upstream with the real credential: inaccessible models 404 identically to completions, accessible ones return 200, and Anthropic does not bill it. `/v1/models` passes the registry's `created` unix timestamps through. Both verified live.
+
+Fact 3 makes validation free and fast, which is what lets one mechanism serve the probe, setup, and the picker.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- The same model list always elects the same default (deterministic rank).
+- A model the credential cannot serve is never silently elected, pinned, or persisted.
+- Zero added cost on the happy path; extra requests only when a candidate fails.
+- One election primitive consumed by three surfaces: launch probe, setup step, TUI picker.
+
+**Non-Goals:**
+
+- No tier steering (e.g. "prefer sonnet"): a tier keyword list is a hardcoded id-shape list that ages like the registry does (`fable` matches no classic tier). Pure recency is the only rank with zero baked-in model knowledge; the "newest release is premium-tier" wrinkle is accepted — setup/picker pins are the escape hatch.
+- No persistence of the elected winner: an auto-written pin recreates the staleness problem with none of the user's intent behind it. Auto stays adaptive.
+- No picker-list badging/filtering by accessibility in this change: commit-time validation alone closes the "persist a 404ing pin" hole; live badge updates add TUI async complexity for UX sugar. Revisit separately.
+- No change to the per-agent resolution precedence (`models.agents.<agent>` → `harness.model` → mode default), to direct-mode's explicit-model requirement, or to the third-party proxy.
+
+## Decisions
+
+### D1. Election lives inside `resolveModelId` — the probe consumes it, never pokes the cache
+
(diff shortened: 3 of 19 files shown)


vivere-dally:
@claude review

vivere-dally:
Thanks for the thorough review — but this one is a **false positive**, and I want to lay out the evidence since the reasoning is subtle.

The finding hinges on one premise: that CLIProxyAPI's `/v1/models` "advertises the full GitHub-sourced registry regardless of which account is logged in," so claude ids are present essentially always. I checked that against the live proxy, and it doesn't hold.

**The registry is multi-vendor; the endpoint is not.** The GitHub `models.json` does carry 12 provider families:

```
top-level keys: claude, gemini, vertex, gemini-cli, aistudio,
                codex-free, codex-team, codex-plus, codex-pro, kimi, antigravity, xai
```

But the proxy's `/v1/models` is scoped to the **authenticated account's provider**. On an Anthropic-authenticated proxy:

```
GET /v1/models  ->  10 models, families present: {'claude': 10}
```

Zero of the other 11 registry families appear. So for a cliproxy account authenticated as openai/gemini/qwen/iflow, `/v1/models` returns **that account's family**, `rankModelCandidates` finds no claude and correctly selects the account's family, and `electModel` validates within it. The all-claude collapse the finding describes doesn't occur, because the mixed list it assumes as input is never produced.

This is also consistent with #144 itself: the two "advertised-but-inaccessible" ids it named — `claude-3-5-haiku-20241022`, `claude-sonnet-4-20250514` — are **both claude**, i.e. within-family tier gaps on an Anthropic account, not cross-vendor leakage.

**Second, independent backstop.** Even in the rare case a mixed list *could* arise (e.g. two credentials loaded into one proxy), a wrong-family election is not served as a silent 404 — boot's provider-family guard catches it loudly:

```ts
// harness/runtime.ts:619
modelMatchesProvider(connection.provider, m)
    ? ok(m)
    : err({ type: "model_provider_mismatch", provider: connection.provider, model: m })
```

So the worst case is an actionable boot error naming the mismatch and the remedy — not "every chat turn 404s permanently." `rankModelCandidates` has no non-cliproxy callers, so this is the only path.

**What the review did surface** is real and worth fixing: the first-family scan's correctness *depends on* the proxy's provider-scoping, and that assumption was load-bearing but undocumented — which is exactly what makes the code read as a cross-provider bug in isolation. Recorded in `0110463`, with a pointer to the boot-guard backstop. No functional change, since there's no incorrect behavior for any real single-provider setup.

I deliberately did **not** add the suggested mixed-list-plus-non-anthropic-provider regression test: it would encode an input the proxy never emits, and would only pass by forcing a provider-aware ranking refactor that isn't needed for correctness. That said, threading `connection.provider` into the election as defense-in-depth (so ranking anchors to the configured provider rather than leaning on proxy behavior) is a reasonable optional hardening if we later want to stop depending on the endpoint's scoping — happy to do it as a follow-up, but it's an enhancement, not a fix.

Issue #130, opened 2026-07-16
Agent-triggered reference-data install: ask the user, then install the missing catalog datasets in-session


radutanasa:
A plan that needs reference data the user doesn't have installed dead-ends. The planner emits a pathway-enrichment step, the step runs, `list_available_refs` reports no Reactome, and the analysis fails — even though `reactome-pathways` is a catalog dataset that is ~2 MB and one command away. The user has to leave the session, run `inflexa refs download reactome-pathways wikipathways-human`, and start over.

The contract this issue proposes: **the agent should be able to notice that a required reference dataset is absent, ask the user for permission, and install it — without the user leaving the session.**

## Why it fails today

Three independent gaps sit behind the dead-end.

### 1. The planner is blind to the ref store

`plannerPrompt` sees exactly three things (`harness/src/tools/research/generate-plan.ts:66-72`): the agent catalog, the resource policy, and the persisted data profile. `harness/src/prompts/planner.ts` contains **zero** mentions of `/mnt/refs`, the ref store, or reference data. So the planner cannot know Reactome is missing — it will happily emit a step that fails at execution.

This is the root cause of the friction. Prompting mid-run is a fallback; the plan should be built against what's actually installed.

### 2. Refs are invisible to the conversation agent

`list_available_refs` (`harness/src/tools/sandbox/list-available-refs.ts:483`) is **sandbox-only** — registered in `BASE_SANDBOX_TOOLS` (`harness/src/agents/sandbox/shared.ts:88,161`) and requiring a live `sandboxClient` + `step.sandbox` + `workflowId`/`stepId`. It cannot be wired into the conversation agent; it only exists inside a running step, long after planning. The chat layer has no way to answer "is Reactome installed?"

### 3. The harness cannot install anything

The refs seam today is **setup-time and one-directional**. The harness *declares* the catalog (`REFERENCE_DATA_CATALOG`, `harness/src/reference-data/catalog.ts:153`, exported at `harness/src/index.ts:314`; spec: `openspec/specs/reference-data-catalog/spec.md`). The CLI *realizes* the download (`installReferenceDatasets`, `cli/src/modules/refs/store.ts:739`) and surfaces it as out-of-band commands (`cli/src/cli/index.ts:372-402`). Nothing lets the harness ask the embedder to install a dataset mid-run.

## On "a tool that calls `inflexa refs download ids...`"

[...]

PR #157, opened 2026-07-18
fix(harness): re-check ref store at each sandbox create so mid-session installs are visible

### LINKED ISSUE #151: Fix cold-start refs mount gate so mid-session installs are visible
Parent: #130

## Problem

The refs mount is frozen when the chat's harness runtime boots. `existingRefStoreConfig(env.refsDir)` is evaluated once at boot (`cli/src/modules/harness/runtime.ts:714`), and the per-sandbox mount decision reads only that frozen snapshot: `refs: !!config.refStorePath` (`harness/src/sandbox/docker-client.ts:266`).

Consequence for #130: a user with no refs dir opens a chat → boot resolves "no mount" → they later run `inflexa refs download …` (which succeeds and writes to the host store) → **every sandbox for the rest of that session still gets no `/mnt/refs`** until the process restarts. The agent-triggered install would silently fail at the last step.

## Fix

Mirror what the libs store already does four lines up in the same function: `libStoreUsable(config.libStorePath)` re-validates at **every** sandbox create (`harness/src/sandbox/docker-client.ts:255`). Refs should re-check at create-time instead of trusting a boot snapshot — pass the refs path through unconditionally and move the existence check to sandbox creation.

Note: the alternative (pre-create the refs dir at boot so the bind is always offered) conflicts with the no-litter policy, which gates dir creation to deliberate actions (`cli/src/lib/env.ts:146-149`). Re-checking at create-time avoids that.

## Done when

- With no refs dir at chat start, running `inflexa refs download <id>` mid-session causes a newly-created sandbox to see `/mnt/refs` without restarting the chat.
- Independent of every other #130 child — this is a standalone correctness fix (arguably a bug regardless of #130).

vivere-dally:
## What & why

The Docker backend decided the `/mnt/refs` mount from a boot-frozen snapshot
(`refs: !!config.refStorePath`). An embedder that gates the store path on
directory existence at boot — the CLI did exactly this — therefore froze
"no ref store" for the whole process lifetime: a reference store installed
mid-session never reached any sandbox created afterwards without a runtime
restart. This silently broke the last step of agent-triggered reference
installs (#130).

Mirror the sibling lib store's re-check. `refStorePath` is now evaluated at
**every sandbox create** and bound only when it is a real directory at that
moment — never binding a missing source (Docker would auto-create a root-owned
directory at the path, which also bricks a later store install). Symlinks are
rejected: the bind authority must itself be the store, not an indirection. A
missing store is the normal cold state and is skipped silently, deliberately
unlike the lib store's degradation warning.

**Embedder wiring:** the CLI passes `refStorePath` unconditionally and its
boot-time existence gate (`existingRefStoreConfig`) is removed, so a mid-session
install is picked up by the next sandbox.

The re-check ships in harness `0.4.5`; the CLI pin advances to `0.4.5` once it is
published on npm (until then the CLI links the local build).

Closes #151

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Documentation updated if user-facing behavior changed. (`ref-store` spec updated; no end-user doc surface changed.)

- [x] **Sandbox:** the security model is preserved — the ref store stays a
  **read-only** `/mnt/refs` mount; the change only tightens *when* it is bound
  (a real directory at create time, symlinks rejected) and drops the path that
  would let Docker auto-create a root-owned host directory. No change to the
  unprivileged uid-1000 workload, dropped capabilities, `no-new-privileges`,
  network posture, exec-signature auth, read-only analysis tree, or resource
  limits.
- [x] **Provenance:** prior analyses remain reproducible — the mount contract is
  unchanged; only stores that *appear* mid-session become visible, matching the
  behavior already granted to stores present at boot. The read-only mount and
  embedder-side receipt gating bound the incomplete-store-mid-download window,
  which is a pre-existing class, not a new one.


vivere-dally:
@claude review

Issue #154, opened 2026-07-18
Agent tool that shells out to the `inflexa` CLI (approval-gated)


vivere-dally:
Parent: #130

## Goal

The heart of #130's approach: the agent uses the **literal `inflexa` binary** on the user's machine like it uses `grep`. It can run `inflexa --help`, discover the surface, and invoke commands (e.g. `inflexa refs download reactome-pathways --yes`) as a subprocess, reading exit code + captured output. All command logic stays **only** in the CLI command layer — invoked as a black box across the process boundary. No `ReferenceStore` seam, no duplicated tool schema, no re-exposed `installReferenceDatasets`.

The tool is **general**, not refs-scoped: refs is the first use case, but the same tool drives any green-flagged command.

## Harness (generic seam — never learns what `inflexa` is)

Today the conversation-agent tool array is hardcoded and closed (`harness/src/agents/conversation-agent.ts:162-268`; closed union per the `per-agent-tool-allowlist` spec) — there is **no** way for an embedder to inject a tool. Add a generic seam for the embedder to contribute conversation-agent tools. The harness stays ignorant of `inflexa`.

## CLI (the tool)

- A tool whose `execute` spawns `inflexa …` via `Bun.spawn` **with an argv array, never through a shell** (no `sh -c`) — so there is no shell to smuggle a second command through, and the approval prompt shows exactly what will run. Pattern precedent: `cli/src/lib/container.ts:84`. Subprocess stdio is captured, so it never touches the TUI alternate screen.

- **Green-flag which commands the agent may invoke** — a per-command allowlist of agent-invocable commands, excluding TUI-launchers (bare `inflexa`, `new`, `resume`, `config`) and interactive flows (`auth login`, clack prompts). This is a *different axis* from the existing `devCommandsEnabled()` gate, which is registration-time (dev commands don't exist in the prod binary at all); the green-flag is "may the agent drive this command" — new per-command metadata.

- **Approval on every invocation** via the tool-approval primitive (#153): the prompt shows the exact argv; the agent proceeds only on approve; write-ish flags run non-interactively (e.g. `--yes` — the non-interactive refs path already exists at `cli/src/modules/refs/commands.ts:172`). Anything destructive (e.g. `down --delete-data`) therefore cannot run without an explicit approval.

[...]

PR #159, opened 2026-07-18
feat: in-chat tool approval — harness ctx.ask primitive + docked TUI prompt

### LINKED ISSUE #153: In-chat tool-approval primitive + docked approval prompt
Parent: #130

## Goal

The reusable "can I do X?" mechanism, so a tool can pause mid-turn, ask the user in chat, and continue based on the answer — the load-bearing piece behind agent-triggered installs (and reusable by others, e.g. `execute_plan`, whose approval is prose-only today with zero enforcement — `harness/src/tools/execute-plan.ts:97-149`).

Modeled directly on OpenCode's permission service: an in-memory pending map + a deferred the tool awaits; no engine changes needed because the tool's `execute` simply doesn't settle until the user replies.

## Harness (generic capability — never refs-specific)

- Add `ask` to `ToolContext` alongside `emit`/`runStep` (`harness/src/tools/define-tool.ts:44-49`), realized by an embedder seam with a **deny-by-default** shipped realization (the established `Unavailable*` pattern).
- Reply is `once | always | reject(+feedback)`, not a boolean.
- A rejected ask surfaces as model-facing prose ("The user rejected … with the following feedback: X") and **stops the loop** rather than letting the agent flail (OpenCode `session/processor.ts:241`). In our contract, denial is an ok-channel data variant per the harness-tools error contract.

## CLI (the surface)

- A **docked** approval prompt above `chat_bar` (not a modal — a modal covers the transcript the user needs to read to decide; OpenCode docks it and disables the composer while pending). Widget in `src/tui/components/` with primitive props; docking in `src/tui/layout/`; **new surface → design-gallery exhibit** per CLAUDE.md.
- The prompt shows the exact action being approved (for the refs tool: the `inflexa refs download …` command it wants to run, like OpenCode shows `$ command`).
- `always` is in-memory / until-restart to start (OpenCode's V1 default); note it's near-useless for refs specifically (installs are idempotent), so the refs prompt can offer just approve/deny while the type stays three-variant.

## Done when

- A tool can call `ctx.ask(...)`, the chat shows a docked prompt, and approve/deny resolves the call; deny stops the turn with model-visible prose.
- Proven by the deferred round-trip test + the gallery exhibit; first real consumer is the `inflexa` CLI tool child.

## Depends on

Nothing. This is the design-heavy piece — worth getting right first.

vivere-dally:
## What & why

A tool had no way to pause mid-execution for an explicit user decision —
`execute_plan`'s approval is prose-only, and the agent-triggered `inflexa`
shell-out (#154) must not run a subprocess without a real gate. This PR ships
the full approval feature: the harness primitive AND the CLI surface that
renders it.

### Harness — the `ctx.ask` primitive

- **Poll-based, DB single source of truth.** `ctx.ask` persists each ask in a
  new `cortex_asks` ledger (uuidv7, status machine
  `pending → resolved | rejected | aborted | expired`), emits a reconciling
  `data-ask` chat part, and polls at 200ms until terminal. Deliberately **no
  in-memory resolver map**: answering is a pure UPDATE, a crash-orphaned row
  just ends the poll, and a cross-process answer still lands (verified with a
  separate answering process).
- **Three-variant reply** — `once | always | reject(feedback)`. `always`
  records an analysis-scoped standing grant (`cortex_ask_grants`, exact
  command key, survives restarts, never crosses analyses); granted asks
  short-circuit but still write an audit row.
- **Denial hard-stops the turn**: `reject` throws → AI SDK `execution-denied`
  result with the feedback → `finish.reason: "denied"`, siblings complete, no
  wrap-up model call.
- **Deny-by-default seam** (`UnavailableAsk`) for workflow contexts and
  headless embedders; `createAskGateway({pool})` carries `answer(id, reply)`
  (`applied | not_found | already_terminal`), `pending()`, and the boot sweep.

### CLI — the docked approval prompt

- The composition root realizes the Ask seam; each TUI turn binds `ctx.ask`
  with the turn's scope/signal/emit; the REPL stays deny-by-default (write-only
  sink, no mid-turn input path).
- `data-ask` parts become transcript ask cards **reconciled by ask id**, and
  drive a pending-asks FIFO store.
- While an ask pends, a **docked prompt above the chat bar** (never a modal —
  the transcript stays readable while deciding) shows the exact command, holds
  focus (making bare `y/a/n` legal under the keymap's focus-target rule and
  gating the composer), offers approve-once / approve-always / reject with
  optional feedback, stacks concurrent asks, and answers through the gateway —
  stale outcomes advance with a notice instead of wedging the queue.
- Asks are live-turn-only visuals; the ledger is the durable record. The
  design gallery exhibits every prompt state and ask-card status.

Specs are synced and changes archived in both trees:
`harness/openspec/changes/archive/2026-07-19-add-tool-approval-ask/` and
`cli/openspec/changes/archive/2026-07-19-add-tui-ask-approval/`.

Note: the CLI builds against the **local linked harness**; the `0.4.5`
registry pin cannot see the primitive until the next harness release, so the
CI typecheck at the pin is expected to lag (releases handled by the maintainer).

Closes #153

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — harness: 1335 pass; cli: 1349 pass.
- [x] Documentation updated if user-facing behavior changed. (Specs synced in both trees; gallery exhibits added.)


krajla (changes requested):
This use case is quite literally what dbos.recv and send are made for. Pausing a workflow to wait for a message, I would use that instead of polling.
And it keeps the state in db, not memory.

Need to make sure this is easily implemented by harness consumers.

vivere-dally:
@krajla — recv/send was considered explicitly; the design maps onto it one-to-one (`ctx.ask ↔ DBOS.recv`, `answer ↔ DBOS.send`) and the decision trail is in the archived change (`harness/openspec/changes/archive/2026-07-19-add-tool-approval-ask/design.md`, Non-Goals) and the spec (`tool-approval` → "Approval polling never blocks a durable step"). Why it lost, concisely:

**`DBOS.recv` is workflow-body-only, and the chat turn deliberately isn't a workflow.** Chat runs in-process on `passthroughStep` (single replica, live token streaming). "Use recv instead of polling" isn't a local swap in the gateway — it means making every conversation turn a durable workflow, which drags in replay-over-a-nondeterministic-LLM-loop, the Result-cached-as-step-success footgun, and streaming-vs-durable-completion. That's a chat-architecture change, deliberately out of scope here.

**"Keeps the state in db, not memory" — that's what this PR does.** There is no in-memory pending map (see the DESIGN(WHY) at the poll site): `cortex_asks` is the single source of truth, `answer(id, reply)` is one guarded UPDATE, and an answer from a *different process* resolves the suspended tool — verified with a separate answering process in review. The only in-memory thing is the suspended tool continuation, and recv doesn't fix that either unless the whole turn is durable: a per-ask recv-workflow survives a crash only to wait for a turn that died with the process (zombie workflow, misleading "resolved" audit). The poll design records that loss honestly (`pending → expired` at boot).

**Consumer ergonomics actually favor the ledger.** An embedder answers via one exported gateway function over the app pool (or an HTTP route in a hosted deployment) — no DBOS coupling, no workflow id/topic routing, no system-DB access in the answering path.

What your instinct is right about, and where it's planned: poll costs 200ms + idle chatter — invisible at human decision speed, and `LISTEN/NOTIFY` on `cortex_asks` removes it without touching the contract if it ever matters. And recv *is* the right transport for two futures: asks raised from inside durable workflows (today spec'd deny-by-default), and turn-as-workflow — both slot under the same `Ask` seam additively. Happy to open a follow-up issue capturing the recv path if you want it on the roadmap.

vivere-dally:
Completing the picture — there's a middle path we also weighed, and it's the closest to what you're describing: **a per-ask workflow**. The turn stays non-durable; `ctx.ask` launches a tiny DBOS workflow whose only job is `reply = await DBOS.recv(\"answer\", timeout)`, the surface answers with `DBOS.send(askWorkflowId, …)`, and the whole thing is a transport swap under the same `Ask` seam — the ledger, reply contract, and denial throw are untouched.

What it genuinely buys: push latency natively (no poll, no hand-rolled LISTEN/NOTIFY), `recv(timeout)` as built-in expiry, and — in its cleanest shape — the workflow as the ledger's single writer.

Why it still lost for the chat path:

| | latency | crash-resume | answering contract | harness delta |
|--|--|--|--|--|
| A (shipped, poll) | 200ms | no — honest `expired` | 1 guarded UPDATE | — |
| A′ (LISTEN/NOTIFY) | ~0 | no — honest `expired` | 1 guarded UPDATE | ~small |
| C (per-ask workflow) | ~0 | **no — zombies** | DBOS send | new workflow + send/cancel seams |
| B (turn-as-workflow) | ~0 | yes | DBOS send | chat-architecture rewrite |

The decisive row is crash-resume: recv's durability attaches to the *workflow*, not the *turn*. After a restart the ask-workflow faithfully resumes waiting — for a suspended tool that died with the process. It can still be answered, completing into a void, leaving a \"resolved\" record for an action that never ran — worse audit than today's swept `expired`. So the boot sweep doesn't disappear, it becomes orphan-workflow cancellation. C pays most of B's plumbing (a registered ask-workflow, plus send/cancel seams — tools never import DBOS, that's the `RunLauncher` quarantine) while delivering only A′'s benefit. And it's not even a stepping stone to B: a durable turn would recv *inline* and delete the child workflow again.

Where C is exactly right, and where we'd reach for it: **asks raised from inside durable workflows** — a sandbox step that legitimately needs a human decision is already a workflow, recv is legal there, and today that context is spec'd deny-by-default. If/when we open that up, C is the design, and it slots under the shipped seam additively.

krajla (approved):
The advantage I see with this hand rolled system is it allows much easier integration anywhere for question asking, not caring about other dbos requirements. If you prefer this option I consider this advantage good enough to ship this.
I imagine the question asking will be integrated in a lot of places and pay for itself in terms of the maintenance cost.


With recv the cost for a workflow per question and replayability edge case wasn't that bad. And forcing ourselves to make it work might have pushed us to make even the agent turns more resilient, agent turns during execution are resilient but not during conversation.

PR #161, opened 2026-07-19
feat(cli): isolate dev/prod infra stacks and pin container images

vivere-dally:
## What & why

Follow-up to #160, which fixed the launch gate misreading transient CLIProxyAPI states as revoked credentials. That investigation left a deliberate TODO: the two remaining failure generators that would keep dragging the team back to CLIProxy — a **floating proxy image** (`:latest` silently changing the behavior the launch gate is calibrated against) and **dev/prod stacks sharing all mutable state** (port 8317/8432, the compose file, the Postgres data dir, and the provider-credential dir — where two independently-refreshing proxies sharing one rotating OAuth refresh token can trip provider reuse-detection and resurrect the forced re-login symptom #160 removed). This PR addresses that TODO, and additionally lands the embedding-backend choices feature (see below).

- **Full stack isolation by build channel.** Dev builds derive sibling ports (proxy 8318, postgres default 8434, DBOS admin 8435 — production keeps its historical 8317/8432/8433 trio) and sibling paths (`cliproxy-dev/`, `postgres-dev/`, `docker-compose.dev.yml`) from the same baked-channel signal that already namespaces container names. Production paths and ports are byte-identical to their historical values (asserted by tests against the literals) — installed binaries are untouched. Both stacks can run simultaneously under channel defaults; an explicitly customized `postgres.port` applies to both channels by design, and the spec now states that exception. One-time dev cost: sign in once into the dev credential dir; dev Postgres starts fresh.
- **Tag + digest pins for every CLIProxyAPI/Postgres image.** `eceasy/cli-proxy-api:v7.2.90@sha256:6aa1ffb6…` (compose service AND the one-shot OAuth-login container, via one shared constant) and `pgvector/pgvector:0.8.5-pg18@sha256:12a379b4…`. Bumps are deliberate one-line diffs; a republished tag is inert. The pin site documents the bump procedure.
- **Launch-gate calibration re-verified on v7.2.90** (was established on v7.2.77 in #160): client-key-only 401 on `/v1/models`, the empty-list-until-registration boot window (reproduced live: 0 models on first poll after restart, 14 on the next), the `auth_unavailable` 503 cooldown body, and the forwarded `not_found_error` body. No classifier changes were needed.
- **Setup persists only explicit Postgres choices.** An accepted default writes nothing to the channel-shared `config.json` (freezing the accepted port there would override the other channel's sibling default), and both a setup re-run and resolve itself heal a previously frozen one — resolve logs a warning when it discards a persisted reserved port that would have changed the outcome. The `inflexa config` screen applies the same persist-only-explicit rule and rejects reserved ports in the editor.
- **Embedding backend choices.** Setup's picker now offers the built-in bge-small model, the user's own local GGUF (width measured through the sidecar probe and recorded as `embedding.dimensions`), an api-key endpoint, or off; the readiness gate checks the configured model path. `inflexa config` reconfigures the backend through a summary row plus dialog chain, fetching the endpoint's embedding-capable models as a selection (abortable, 10s-bounded, degrading to free-text entry) and refusing to send the api key over cleartext http to a non-loopback host.

Specs: `postgres-provisioning` updated (three requirements) with the change archived under `cli/openspec/changes/archive/2026-07-19-isolate-dev-infra-stack/`; `local-embeddings` updated with the change archived under `cli/openspec/changes/archive/2026-07-20-add-embedding-backend-choices/`.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). Full suite: 1370 pass / 0 fail; typecheck and lint fully clean.
- [x] Tests added or updated for the change (channel-aware ports/paths with prod pinned to historical literals, cross-channel path disjointness, tag+digest compose generation in both modes, persist-only-explicit incl. heal-on-reaccept, api-key model listing incl. the cleartext-transport guard).
- [x] Documentation updated if user-facing behavior changed (the OpenSpec specs are the doc of record).
- [ ] This PR is focused on a single logical change — it contains two related changes (infra-stack isolation and embedding-backend choices), each with its own OpenSpec change of record.


vivere-dally:
@claude review

vivere-dally:
Re: the review findings above — all three addressed as of `ce5e9cf`:

1. **ConfigApp embedding wiring untested** → `cli/src/tui/app_config.render.test.tsx` (new) drives the real `ConfigApp` through the keyboard bus: api-key happy path (fetched model picker → width → persisted block), the free-text fallback on a failed listing, the custom-GGUF chain through the real `FilePicker`, and the dead-flow guard (a superseded probe's late resolve pushes no dialog and changes no notice). Assertions run against the persisted config in the XDG test sandbox.

2. **`finishApiKey` always writes `dimensions`** → fixed in the same commit: omitted when it equals the api-key default 1536, mirroring the local branches' 384 rule. Safe because `resolve.ts` passes the value through and the harness provider falls back to the same constant regardless of model.

3. **Image digests unverifiable from the sandbox** → verified against Docker Hub from a networked environment: the registry manifest HEAD for `eceasy/cli-proxy-api:v7.2.90` returns `sha256:6aa1ffb6…d73d2` and for `pgvector/pgvector:0.8.5-pg18` returns `sha256:12a379b4…647e0` — both as multi-arch OCI index digests, i.e. exactly the values pinned in the compose file.

Full gate at `ce5e9cf`: typecheck clean, lint clean, 1376 pass / 0 fail.

Issue #155, opened 2026-07-18
Planner ref-awareness: plan against installed reference data


vivere-dally:
Parent: #130

## Problem (the root cause behind #130)

The planner is blind to the ref store, so it emits steps that die at execution when the needed dataset isn't installed. It sees only the agent catalog, resource policy, and data profile — nothing about `/mnt/refs` or what's installed.

## Change

Project the installed-ref inventory into the planner so it plans against what's actually present (and can route a missing dataset to the agent-triggered install flow instead of emitting a doomed step).

Constraint: the planner system prompt is built **once per factory** (the resource policy is process-fixed). A **mutable** ref inventory must therefore go in **per-invocation seed content**, not the system prompt.

## Depends on

The machine-readable `inflexa refs` output child (structured install state to consume). Independent of the approval/subprocess path — could land before or after the install tool.

## Done when

The planner, told Reactome is absent, does not silently emit a step that will fail at execution.

krajla:
`list_available_refs` is currently the source of truth for usable refs by the agents, I recommend attaching it and maybe a line in planner agent prompt to take into consideration refs.(No tool name dropping in prompts, we got burned with those making tool changes flaky).

Issue #152, opened 2026-07-18
Machine-readable `inflexa refs` output for programmatic consumers


vivere-dally:
Parent: #130

## Problem

Programmatic consumers need to know what reference data is installed. Today `inflexa refs list` prints human-readable lines only (`cli/src/modules/refs/commands.ts:205` `runRefsList` — `console.log` per dataset), so the only way to learn install state is to scrape prose. The planner (see the planner-awareness child) needs this reliably, and it's cleaner for the agent tool too.

## Change

Add a structured/`--json` output mode to `inflexa refs list` (and optionally `refs verify`), emitting install state per catalog dataset from the existing `inspectReferenceStore` result (`cli/src/modules/refs/store.ts:328`). This enhances the existing command's output — no new download/verify logic, no reimplementation.

## Notes

- The agent, calling the CLI like `grep`, can read plain `refs list` text fine (it reads like any tool output). The structured mode exists for the **planner**, which consumes it host-side at plan time.
- Keep it byte-stable and side-effect-free (inspection never creates the store).

## Done when

- `inflexa refs list --json` emits per-dataset install state (installed / missing / update-available / …) in a documented shape.
- No behavior change to the human output.

PR #163, opened 2026-07-20
feat: agent tool to drive the inflexa CLI, approval-gated

### LINKED ISSUE #154: Agent tool that shells out to the `inflexa` CLI (approval-gated)

vivere-dally:
## What & why

An analysis that needs an uninstalled reference dataset used to dead-end: the agent had no way to run `inflexa refs download …`, so the user had to leave the session. This adds a single `run_inflexa` conversation-agent tool so the agent drives the local `inflexa` CLI the way it drives `grep` — it discovers the surface through `--help`, then invokes commands as a subprocess. **The in-chat approval prompt is the security boundary, not a curated per-command allowlist.**

Built harness-first, then wired from the CLI, per the subsystem boundary rule.

**Harness (host-agnostic — never learns what `inflexa` is):**
- A generic `hostTools` seam on `ConversationAgentDeps` so an embedder can contribute conversation-agent tools; they are dispatched identically to built-ins (same `ToolContext`, including `ask`) and never reach a sandbox agent.
- An optional `AskRequest.grantKey` on the `ctx.ask` primitive: a tool can display the exact command while keying its `always` standing grant on a broader class. Absent, the grant keys on `command`, so it is backward compatible (new `cortex_asks.grant_key` column).

**CLI (owns everything `inflexa`):**
- The commander registry becomes a reusable `buildProgram()` factory; `cli` is derived from it so existing importers (entry point, docs generator) are unchanged.
- A classifier decides — **without executing anything** — whether an argv is introspection, an action, or malformed, by parsing against the real commander tree (a `preAction` sentinel under `exitOverride`, output silenced). Because it is commander's own parse, a `--help` placed after a `--` terminator is an operand, not the help flag — so a masked action cannot slip past as introspection. Help/version auto-allow with no prompt; every action is approval-gated; a parse error returns to the model without spawning.
- The `run_inflexa` tool: `Bun.spawn` with an argv array (no shell), `stdin` ignored, `stdout`/`stderr` captured off the TUI, bounded by a timeout, output truncated. The approval shows the exact argv; the `always` grant keys on the resolved subcommand path (e.g. `inflexa refs download`), so approving one dataset does not re-prompt for the next.

Deny-by-default off the interactive TUI (an unwired `ask` denies), so nothing runs unattended. Builds on its now-merged siblings: the refs mount-gate fix (#151), machine-readable `inflexa refs` output (#152), and the in-chat approval primitive (#153).

Closes #154. (The final shell-out piece of #130; planner ref-awareness, #155, is a separate track.)

> **CI note:** the CLI builds against the linked local harness for `hostTools` / `AskRequest.grantKey`; a typecheck against the published harness pin will lag until the next harness release, exactly as the #153 CLI side did.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). (harness: 1341 pass / 0 fail; cli: 1387 pass / 0 fail.)
- [x] Documentation updated if user-facing behavior changed. (OpenSpec specs synced in both trees: new `host-conversation-tools` and `agent-cli-tool`, plus `tool-approval` / `cli-core` / `harness-runtime` deltas.)


vivere-dally:
@claude review

PR #166, opened 2026-07-20
fix(refs): find reference data by meaning, and read both inventories host-side

krajla:
## Why

Skills told agents to load reference data from a `{category}/processed/{file}` tree **no installer produces**. The catalog installs to `managed/{id}/{version}/{artifact.path}`. Four categories referenced by skills (PROGENy, OmniPath, DoRothEA, LINCS) had no catalog entry at all, and `CATEGORY_LABELS` documented a `/mnt/refs/registry.json` nothing in the repo writes.

The deeper problem was the shape, not the paths: content was encoding *where* things live, which is an installer detail that outlives no environment. Chasing it surfaced the same failure in two tools that read environment inventories.

## What changed

**Reference data is found by meaning, not location.** The catalog gained `format`, `contents`, and `organism` per artifact, so an agent asking for "a TF-target regulon network for mouse" can recognise the file that holds it. `list_available_refs` joins that metadata onto what it observes and searches over it. Skills now state the task; the tool describes itself.

**Both inventory tools read the store host-side.**
- `list_available_refs` shipped a Python scanner into the sandbox to enumerate `/mnt/refs`. Nothing about a directory listing needs a container — the store is a host directory under Docker and a PVC the harness service also mounts under k8s. It now reads through `node:fs` at an injected path and renders results under `/mnt/refs`, so a returned path pastes straight into a container script.
- That made it attachable to `generate_plan`, which has no sandbox at all. A plan can now name the reference data this install actually holds instead of assuming.

**`list_available_packages` was silently broken on every CLI install.** It read `/mnt/libs/current/packages.txt` from the *host* filesystem, but the CLI bakes the store into the image and never creates a `/mnt/libs` bind mount — so the read always failed and every agent was told to *"assume standard bioinformatics packages (numpy, pandas, scanpy, DESeq2, etc.) are available."* It failed soft, so nothing surfaced it. The image build now stamps its load-tested inventory onto the image as an OCI label; the CLI reads it with `docker image inspect` (metadata on an already-pulled image — no container created, no registry client) and caches by image ID.

Per-arch matters here: amd64 carries packages arm64 has no build for, and R can be absent from arm64 entirely, so a single published list would be wrong for half of users. A label is per-arch by construction — the pulled image *is* the host's arch.

**Two behaviour fixes fall out of the same root cause.** The unreadable-inventory note no longer names packages (that state means the set is UNKNOWN, so a roll-call is a guess the agent cannot check). And the default listing no longer truncates: the store is ~270 packages / ~3 KB, but the default limit of 100 rendered a third of it, and a partial listing reads as a complete one.

**Every agent that can provision refs can now see them.** The cli hands the conversation agent `run_inflexa`, which drives `refs download` gated on user approval — so that agent asks for downloads, and it had no way to check what was already installed. It could ask the user to download data they already have, or miss that something was absent. It now carries `list_available_refs`, as do the planner and every sandbox agent.

**Skills and prompts stopped asserting things they do not control.** Sweeps across all skill packs removed invented tool names, corrected format readers (PROGENy/DoRothEA ship as R `.rda`, not CSV — a bug I introduced in my own template and propagated before catching it), and dropped unbacked environment claims. The sandbox prompt now states plainly that there is no egress.

**`mordred` → `mordredcommunity`.** Upstream 1.2.0 calls `np.float`, removed in NumPy **1.24**, so it fails on import under any current NumPy.

The design rule behind all of it is recorded in the root `CLAUDE.md`: content declares *what*, the runtime resolves *how*.

## Verification

- `tsc --noEmit` clean on `harness`.
- Full harness suite diffed against `origin/main`: **131 failures on both sides, identical sets** — all Postgres testcontainer failures (no container runtime available locally). Zero introduced.
- CLI failure set likewise unchanged; lint unchanged at 119 pre-existing errors.
- `shellcheck` clean on the new build script; workflow YAML parses.

**Not verified locally, and worth reviewer attention:**
- **The CLI does not typecheck in isolation** — it is pinned to npm `@inflexa-ai/harness: 0.5.0`, which does not carry the new optional deps. The harness must publish before the CLI half compiles. CLI wiring here is review-verified, not compiler-verified.
- **The `generate_plan` tests need Postgres** and did not run.
- **The image label script and workflow step are unexecuted** — first real proof is a CI run on an `images/**` push.


krajla:
@claude review

krajla:
closes https://github.com/inflexa-ai/inflexa/issues/155

vivere-dally (dismissed):

PR #170, opened 2026-07-20
feat(cli): registration-declared agent command policy for run_inflexa

### LINKED ISSUE #165: run_inflexa: BLOCKED_COMMANDS drifted from courtesy layer to load-bearing curated denylist — redesign command availability
## Context

The design thesis of #163 (archived change `add-inflexa-cli-tool`) is that **the in-chat approval prompt is the security boundary, not a curated per-command allowlist**. `BLOCKED_COMMANDS` (`cli/src/modules/harness/inflexa_tool.ts`) began consistent with that: a courtesy layer over the structurally-unrunnable TUI launchers, whose real backstop is `requireInteractiveTerminal` at every launcher entry (`cli/src/lib/cli.ts`).

During review hardening (`49e5df0`), the infrastructure-lifecycle commands (`up`, `down`, `setup`) were added to the map — for a sound immediate reason (`inflexa down` stops the Postgres the running session is connected to). But that addition changed the map's nature: those commands run fine headless, so for that family **the map is the gate**, with no structural backstop behind it. The map's own doc comment concedes it: *"for this family the map IS the gate: a new lifecycle command must be added here."*

## Problem

- The tool's stated design ("no curated per-command list decides what may run") and its implementation (a curated, load-bearing denylist) now contradict each other.
- Nothing enforces that a new command is triaged into one of the three de-facto buckets: freely-approvable action / TTY-guarded TUI launcher / blocked lifecycle command. A future session-affecting command added to the registry silently becomes approvable through `run_inflexa`.
- Policy lives far from the commands it governs: the registry (`cli/src/cli/index.ts`) declares the commands; agent policy sits in a string-keyed map in a different module, coupled only by the grantKey spelling (`"inflexa refs download"`-style strings). A rename or restructure in the registry silently orphans a map entry.

The review bot proposed a stopgap ([comment](https://github.com/inflexa-ai/inflexa/pull/163#issuecomment-5021238507)): a test that walks `buildProgram()`'s command tree and asserts every leaf is accounted for in one of the three buckets. That turns a forgotten entry into a CI failure, but preserves the split-brain design.

## Direction

Deliberately open — a better design is to be proposed here. Candidate directions noted so far, non-binding:

1. **Declare agent availability at command registration** — each command in the registry carries its own agent-facing class (e.g. available / blocked-with-reason), making the registry the single source; the tool derives its behavior, and exhaustiveness is structural (an unclassified command cannot be registered).
2. **Derive rather than curate where possible** — TUI launchers are already self-identifying via their TTY guard; only genuinely policy-driven blocks (the infra family) would need an explicit declaration.
3. **Interim**: the bot's tree-walk exhaustiveness test, if the redesign doesn't land soon.

## References

- `BLOCKED_COMMANDS`: `cli/src/modules/harness/inflexa_tool.ts`
- Spec requirements: `cli/openspec/specs/agent-cli-tool/spec.md` ("Interactive TUI-launcher commands are blocked from the agent", "Infrastructure lifecycle commands are blocked from the agent")
- Lineage: #130 → #154 → #163

vivere-dally:
## What & why

`BLOCKED_COMMANDS` in `run_inflexa` had drifted from a courtesy layer over structurally-unrunnable TUI launchers into a load-bearing curated denylist for the infrastructure-lifecycle family — contradicting the tool's stated design ("no per-command list decides what may run"), with nothing enforcing that a new command gets triaged and with policy living in a string-keyed map far from the commands it governs.

This PR replaces the map with a **registration-declared agent policy**: every action command in the registry declares one of three kinds at its registration site via `registerAction(command, policy, handler)`:

- **`auto`** — runs prompt-free, but only when every explicitly-set option is in the policy's `safeFlags` allowlist; any other invocation escalates to the approval prompt. Eleven read-only commands (verified against their action implementations) ship as `auto`: `sessions`, `ls`, `project ls`, `auth whoami`, `refs list`, `refs path`, `refs verify`, `sandbox status`, `prov lineage`, `prov verify`, `prov verify-file`.
- **`approval`** — the default: the in-chat `ctx.ask` prompt, grantable per analysis (unchanged behavior).
- **`blocked`** — never runs, with a mandatory model-facing reason (the TUI launchers and `up`/`down`/`setup`, reasons carried over verbatim).

The invariant: **policy is the floor; flags only escalate.** Explicitly-set options are derived from the commander parse (canonical `attributeName` + option-value source), never argv string matching, so short forms and `--opt=val` collapse correctly and a post-`--` flag stays an operand.

Exhaustiveness is enforced in four layers: the required policy parameter (compile error), a registry-scoped ESLint ban on raw `.action(`, a both-channels tree-walk test with pinned `{grantKey → kind (+ safeFlags)}` snapshot tables as the audit surface, and a runtime fail-closed default (an unclassified action returns `blocked`, never silently approvable). The classifier now returns the stamped policy off the resolved `Command` instance, so a rename can never orphan a policy entry. Blocked runs before any grant lookup, so a stale "always" grant cannot resurrect a blocked command.

Specs: new `agent-command-policy` capability; `agent-cli-tool` restated over the three-tier policy (thesis, blocked families, classification, auto tier, fail-closed). Design decisions and the verified read-only audit live in the archived change (`openspec/changes/archive/2026-07-20-add-agent-command-policy/`).

Closes #165

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist


vivere-dally:
@claude review

PR #173, opened 2026-07-20
feat(cli): guard TUI contrast at the rendered span, and fix the blocks that regressed

vivere-dally:
## Why

opentui's text renderable defaults its foreground to opaque white — `_defaultOptions = { fg: RGBA.fromValues(1,1,1,1) }`. So a `<text>` that names no `fg`, and any `<Bold>`/`<Italic>`/`<Underline>`/`<Dim>` with no colored ancestor (those emit attribute-only spans and carry no color of their own), paints `#ffffff`.

That measures **12–18:1 on the five dark themes** but **1.00–1.13:1 on the five light ones** — exactly 1.00:1 on `github-light`, whose background is pure white. The default theme is dark, which is why this shipped and kept shipping.

Eight sites had it, all confirmed by span capture rather than by eye:

| Site | Measured |
|---|---|
| `ask_prompt.tsx` — prompt title | 1.06:1 on the raised panel |
| `openable_card_block.tsx` — card title | 1.00:1 |
| `presentation_block.tsx` — block title | 1.00:1 |
| `design_gallery.tsx` — 5 of 6 "Type & emphasis" specimens | 1.00:1 each |

The gallery one is the sharpest: its captions rendered while the specimen words they describe vanished, so the canonical reference for the emphasis vocabulary displayed nothing on a light theme.

## The actual problem: nothing could see it

This class was already diagnosed in `docs/color_contrast_audit.md`, and its remedy written down. It recurred anyway, because all three test layers miss it by construction:

- **The palette matrix** (`design_system.contrast.test.ts`) measures *declared* tokens against backgrounds. Text that declares no token at all is invisible to it — you cannot enumerate a pair a component never declares.
- **Every render test** asserts through `captureCharFrame()`, which returns characters with no color. An unreadable span satisfies `toContain(...)` identically to a correct one. One card test additionally reused the same string for the card title and a row name, so it passed on the row while the title rendered unpainted.
- **The one span-color test** that existed hand-enumerated two components.

The history makes the point concretely, and in two different ways. `design_gallery.tsx` predates the audit and its specimens were already unpainted **on the audit date** — a surface the inventory was free to read and did not reach, so that miss belongs to the method. The three block titles then arrived *after* the diagnosis: `openable_card_block.tsx` and `presentation_block.tsx` on 2026-07-13, `ask_prompt.tsx` on 2026-07-19 — each reintroducing the identical fallthrough in a repo that already carried this document and its stated remedy.

A written rule plus a palette-level matrix is therefore not containment for this class. That is a measured result, not a worry.

## What this adds

**A guard that measures what reached the screen.** `theme_contrast.render.test.tsx` grows from two hand-picked cases into a 13-block sweep: render on `github-light`, walk `captureSpans()`, resolve each span's background from its own when opaque (falling back to the surface), and fail anything under its floor.

Tiers derive from span *content*, not from per-component declarations — the shape that caused this regression in the first place. A span composed solely of decorative glyphs is held to 3:1; everything else to 4.5:1, and anything unrecognized falls to the **stricter** side so the guard fails loud. The decorative set is held as `GLYPHS` **keys**, so a renamed glyph breaks the build rather than quietly relaxing a threshold. WCAG math moved to `src/test_support/contrast.ts`, shared with the palette matrix so the two cannot drift apart.

It was built **before** the fixes and confirmed red on exactly the three broken titles, so we know it catches the defect rather than merely passing.

**Authoring rules** in `cli/CLAUDE.md` stating the invariant with its mechanism: both valid fix shapes, the banned bare-literal shape, the 4.5:1 / 3:1 tiers, the light-theme obligation, and that a character-frame assertion proves emission — not legibility.

## Two redesigns

**The approval prompt** puts its marker in the shared 2-cell gutter with content hanging at that indent, so the glyph marks the block rather than decorating the title row, and the prompt aligns with the transcript above it:

```
 ⚠ Fetch external dataset
   curl -O https://ftp.ncbi.nlm.nih.gov/geo/GSE78220.txt.gz
   downloads ~40 MB into the analysis sandbox before profiling
   y approve · a always · n reject  +2 more
```

The transcript ask card also drops `◐` for the same `⚠`. One pending ask was wearing two different markers depending on where you looked at it; `◐` means system-busy everywhere else, but a pending ask is not the system working — it is the system waiting on the user.

**The openable card's marker column** stopped encoding content kind. Four glyphs, each colliding with a status meaning elsewhere (`●` is also the plan-card marker, multi-select selected, the active radio dot, chat `ready` and three sidebar run states; `▸` was the tool-running marker *and* the folder affordance inside this same card), with the title `●` and a chart row `◐` landing as two near-identical adjacent circles. It now answers one question — does this open, or is it broken:

```
Differential expression
  ↗ Volcano plot
  ~/proj/.inflexa/analyses/rna/presentations/pres-9f21a3.html

Figures
  ↗ volcano.png · condition A vs B
  ↗ heatmap.png · top 50 DE genes
  ↗ Open containing folder

Referenced file
  ✗ de-summary.csv
  ✗ Report preview v2 failed · render timed out after 60s
```

Kind is left to the filename, which carries it better than a single cell can. `OpenableIcon` and `iconForPath` are therefore deleted rather than left vestigial — `icon` never crossed a harness wire (it was derived CLI-locally from the file extension), and parts deserialize through an unchecked `JSON.parse(...) as Part`, verified on the read path, so rows still holding the key read back unchanged.

## Verification

| Check | Result |
|---|---|
| Full suite (`cli/`) | 1493 pass, 0 fail, 1 skip |
| Contrast guard | 13 pass / 0 fail |
| Lint | clean |
| Typecheck | 10 pre-existing harness-pin errors in `modules/harness`/`modules/refs`, unchanged by this branch |

Spec deltas are merged into `theme-system`, `tui-text-emphasis`, `tui-stream-blocks`, `tui-ask-approval`, and `tui-test-coverage`; the change is archived.

## Reviewer notes

- `DiffBlock` coverage was added after the red baseline was recorded, and confirmed non-vacuous (`renderUntil` returns on timeout too, so a probe verified real diff-body spans were measured).
- Three whole-frame negative assertions were tightened — `not.toContain(...)` over an entire frame would pass identically if the element vanished.
- The decorative tier includes a queued step's hollow circle. That is `fgSubtle` ornament the palette matrix already blesses at 3:1; measuring it at the text floor would contradict a deliberate tier decision and push toward collapsing `fgSubtle` into `fgMuted`, which `design_system.ts` explicitly rejects.


vivere-dally:
@claude review

krajla (approved):

PR #179, opened 2026-07-21
feat(cli): make the run step window admit what it hides, and let the …

vivere-dally:
…reader scroll it

The sidebar RUNS progress embed capped its step list at a fixed number of rows and dropped the remainder silently, so a seven-step run rendered `1/7` above six rows with nothing to say a seventh existed. The reader could not tell whether the missing step was hidden, unreported, or never planned — the same defect the runs picker already guards against by stating its 100-run cap out loud.

The window now brackets itself with markers naming the hidden count on each side, and those markers are the control that resolves them: a click slides the window one step that way, so the row admitting what is missing is the thing that reveals it, and the two counts double as the position indicator. Windowing is skipped entirely when it would not pay for itself — hiding one step to spend a row on its marker is the same height carrying strictly less information, and a plan landing a step or two over the cap is the common case, not the exception.

The window centres on the earliest RUNNING step rather than the first not-done one. Steps run in parallel waves, so several are in flight at once, and anchoring on the first not-done step would follow a finished sibling forward and scroll work that is still live off screen. It returns there whenever the set of running steps changes — the set, not merely the anchor, because a sibling completing changes the work while leaving the anchor exactly where it was.

Scroll position survives a ledger poll. The rail re-reads every few seconds and re-renders with a freshly minted props object, so the release triggers are memoised to fire on a real change rather than on every read; `on()` alone would fire on each one, since it never diffs what its dependency returned.

The markers opt out of text selection and contain their own clicks. They are buttons rather than prose, and they sit inside the RUNS section, whose mouse-up opens the runs picker — an un-contained release would scroll the window and pop a dialog over it on the same press. Containment is scoped to gestures that begin on a marker, so a selection drag merely crossing the rail still reaches the root's copy handler.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/cli/src/lib/design_system.ts
+++ b/cli/src/lib/design_system.ts
@@ -740,6 +740,15 @@ export const size = {
     composerMin: 1,
     /** Palette rows visible before it scrolls. */
     paletteRows: 12,
+    /**
+     * Run-step rows the rail's progress embed shows before it windows the list behind its scrollable
+     * elision markers. ODD by requirement, not by taste: the window centres by placing the running step
+     * at `floor(n / 2)`, which is the exact middle row only for an odd count. Lives here as a token
+     * because the sidebar, the gallery exhibit, and the tests that pin the window's arithmetic must all
+     * name the SAME cap — held by hand in four places it has already drifted, and a stale copy silently
+     * measures a window the product never renders.
+     */
+    railStepRows: 7,
 } as const;
 
 /**
--- a/cli/src/tui/components/run_block.render.test.tsx
+++ b/cli/src/tui/components/run_block.render.test.tsx
@@ -3,10 +3,14 @@ import { createSignal } from "solid-js";
 import { testRender } from "@opentui/solid";
 import { renderFrame } from "../../test_support/tui.ts";
 
+import { size } from "../../lib/design_system.ts";
 import { RunBlock, type RunStepView } from "./run_block.tsx";
 
-/** The rail's window size — the sidebar RUNS embed's `maxSteps`, mirrored so the cases track it. */
-const RAIL_WINDOW = 7;
+/**
+ * The rail's window size, taken from the design-system token the sidebar itself passes rather than
+ * repeated here — the arithmetic these cases assert is only meaningful against the cap that ships.
+ */
+const RAIL_WINDOW = size.railStepRows;
 
 /** A run whose first `done` steps are complete, the next is running, and the rest are queued. */
 function steps(total: number, done: number): RunStepView[] {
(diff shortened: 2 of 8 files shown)


vivere-dally:
@claude review

vivere-dally:
@claude Both findings addressed in e58f3a4 — please review again.

### 1. `pressedMarker` stuck flag — fixed, and it was more reachable than flagged

You framed the stranding gesture as the button being released somewhere the component never sees a mouse-up for, and rated it low real-world impact. There's a mundane trigger that makes it routine: **pressing a marker that scrolls the window to the end of the list unmounts that very marker**, so the `<Show>` tears down the element mid-gesture and no mouse-up ever arrives to clear the flag.

The consequence is also worse than "wrongly swallowed". Copy-on-select fires from the app root's mouse-up (`app.tsx:367` `copySelection`), so the swallowed release costs the user a **clipboard copy**, with nothing on screen indicating it didn't happen.

Rather than harden the flag, I removed the state entirely. The release now decides from the live selection:

```ts
function releaseMarker(e: { stopPropagation(): void }): void {
    if (renderer.getSelection()?.getSelectedText()) return;
    e.stopPropagation();
}
```

That's sound because of two things already true in the app, which I verified rather than assumed:

- `copySelection` (`app.tsx:367`) returns early on an empty selection, so swallowing a selection-less release loses nothing.
- `openRunsFromSidebar` (`app.tsx:199`) already ignores activations that carry a selection, so letting that release bubble cannot open the picker.

So: release with a selection → bubbles, copy works, picker stays shut. Release without one → contained, picker stays shut. Nothing is remembered between mouse-down and mouse-up, so the whole stuck-state class is gone rather than narrowed. It also matches the guard `openProfileFromSidebar`/`openRunsFromSidebar` already apply to the same gesture.

Verified against the counterfactual rather than a green suite: reverting to the flag-based version fails the new test in `run_block.render.test.tsx` with `Expected: 2, Received: 1` — the selection-carrying release swallowed, exactly your scenario. The test walks the window to its end (the press that unmounts its own marker), then drags a selection onto the surviving marker and asserts the release reaches the ancestor.

### 2. `maxSteps={6}` in the contrast test — fixed as a token, not a number

Correct catch. I didn't just change the 6 to a 7: the value had already drifted 6 → 5 → 7 during this PR while being hand-copied into four places, so the number was the wrong unit of fix.

It's now `size.railStepRows` in `design_system.ts`, read by `sidebar.tsx`, the gallery exhibit, the contrast guard, and `RAIL_WINDOW` in the test suite. It sits beside `paletteRows`, which bounds the palette the same way, and its doc records the odd-number requirement (centring places the running step at `floor(n / 2)`, the exact middle only for odd counts). A stale copy can no longer measure a window the product doesn't render.

### 3. The commands you couldn't run

Ran all three on this branch:

- `bun test` — **1522 pass, 1 skip, 0 fail** (full `cli` suite, not just the TUI, since `design_system.ts` is widely imported)
- `bun run lint` — clean
- `bun run typecheck` — 10 errors, all pre-existing `refStorePath`/`format` harness type-drift in `modules/harness/` and `modules/refs/`. Identical count with these changes stashed; none in `tui/` or `lib/`.

One note for the re-review: your trace of the windowing math was done against `maxSteps` values that are now a single token, so please re-check the arithmetic in `run_block.tsx` against `size.railStepRows` rather than any literal.

krajla (approved):

PR #182, opened 2026-07-21
feat(ci): derive CITATION.cff's version from cli/package.json

vivere-dally:
CITATION.cff restated the product version by hand and had drifted five releases behind (0.1.0 against a cli/package.json at 0.4.2), shipping a wrong citation in every release since. Nothing caught it, because a stale citation is not a build failure.

cli/package.json is already the single source of truth for the product version — its version field is what mints the v* tag and the GitHub release — so the citation metadata now follows it:

- .github/citation/sync.sh rewrites version/date-released, or verifies them with --check. It refuses to run if the top-level keys ever move, so a restructured file fails loudly instead of syncing nothing.
- version-guard.yml checks it on every PR. It already runs as a required check and already reads the version field, so the job name is unchanged and no required-check reconfiguration is needed. Checked unconditionally rather than only on bump PRs, since the drift can equally arrive from an edit to CITATION.cff itself.
- release-prep.yml syncs inside the bump commit. Zenodo reads the file at the tag's ref, so same-commit placement keeps the tagged tree correct.

release.yml is deliberately untouched: it cuts the tag at $GITHUB_SHA, so syncing there would either tag a stale file or require pushing an unreviewed commit to protected main.

Also drops the per-version DOI entry. Zenodo mints a version DOI only after the tag is published, so a value written into a file that ships inside that tag could only ever name the previous release — and Zenodo ignores the identifiers key regardless. The concept DOI resolves to the latest version and needs no maintenance.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/.github/citation/sync.sh
+++ b/.github/citation/sync.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# Keeps CITATION.cff's version in step with cli/package.json's.
+#
+# Usage: sync.sh [--check]
+#   (no args)  rewrite CITATION.cff in place
+#   --check    exit non-zero if it is out of step, changing nothing
+#
+# cli/package.json is the single source of truth for the product version: its
+# version field is what mints the v* tag and the GitHub release
+# (.github/workflows/release.yml), so the citation metadata follows it rather
+# than being maintained by hand.
+#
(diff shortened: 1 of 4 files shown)


krajla (approved):
👍🏻

PR #183, opened 2026-07-21
fix(harness,sandbox): gate lineage edges on producing-step completion

vivere-dally:
## The failure

Step `T2S2` (immune-focused GSEA) in run `19110b58` died with `errorClass: lineage_attestation` — surfaced to the user as "Step results could not be finalized" — **after its analysis had completed successfully**. It had written every output, every figure, and a well-formed summary.

The log names the culprit:

```
stepId: T2S2   path: .../runs/<runId>/T5S1/output/_ct_for_r_BRAF.csv
source: "upstream"   refStepId: "T5S1"   throwSite: "input-enoent"
```

`T2S2` is the enrichment agent. `_ct_for_r_BRAF.csv` is a scratch file belonging to **`T5S1`**, the statistical-modeling step, which was running concurrently and deleted it before `T2S2` reconciled. A step was killed by bookkeeping about a file it never used.

It is symmetric: `T5S1` was simultaneously failing to attest `T2S2/output/gseapy_hallmark`. Two steps each claimed the other's outputs as inputs, which is not a DAG.

**The crash is the lucky case.** In the same run, `T4S1` recorded three `T2S2` files (`logs/run_gsea.log`, `output/wikipathways_symbols.gmt`, `scripts/run_gsea.py`) as its own inputs while `T2S2` was still running. Those happened to survive to reconcile, hashed cleanly, and were **registered as real lineage edges**. A survival-analysis step did not read a GSEA log. Silent provenance corruption is the worse half of this defect, and nothing surfaces it.

## Why it happened

Two premises were wrong.

1. `classifyReadPath` turned **any** same-run path outside the step's own directory into a mandatory, content-attested `upstream` input. Its justification: *"a read of a same-run step outside `dependsOn` is still a valid upstream input."*
2. Reconcile deferred input hashing to teardown *"because inputs are immutable for the step — the analysis tree is mounted read-only."*

Read-only bounds what **this** step may write. Every other step has its own directory mounted read-write and mutates it freely, scratch files included. The tree is shared mutable state for exactly the paths premise 1 marked attestable.

Concurrency is normal, not exceptional: the CLI's machine budget is a user-chosen fraction of the host, and the scheduler starts every dependency-satisfied step that fits.

## The fix

**An input edge is admissible if and only if the producing step's status was `completed` at the moment the reading exec was submitted** — whichever run it belongs to.

Two steps running concurrently therefore have no lineage relationship in either direction. That falls out of the same predicate rather than needing a rule of its own.

`completed` is the only admissible status. `failed`, `canceled`, and `skipped` are terminal but never finalized their outputs — which is precisely what a `lineage_attestation` failure means.

Three decisions worth a reviewer's attention:

**The snapshot is taken at submit, not at reconcile.** Reconcile runs after sandbox teardown, by which time a sibling that was mid-flight *while the file was read* may have reached `completed` and the edge would wrongly pass. A reconcile-time check would not have caught the `T4S1` case at all.

**The snapshot is durably checkpointed.** `execute_command` is `executionMode: "workflow"`, so its body runs unwrapped in the DBOS workflow body — an unwrapped query re-executes on replay and returns a *larger* completed set, giving the same run different lineage on recovery. It goes through `ctx.runStep`.

**Failure fails closed, loudly, and durably.** A query failure resolves to an explicit "snapshot unavailable" *value* rather than throwing, because provenance must never fail an exec. All sibling reads for that exec become inadmissible, and the degradation is logged and counted. The degraded outcome is itself checkpointed — otherwise a replay could succeed where the original failed.

Two alternatives were considered and rejected. **Gating on `dependsOn`**: it is a scheduling input, and a step may legitimately read a sibling it never declared, so this would drop real edges. **Gating `prior`-run reads on the run being terminal**: `cortex_runs` counts `partial`, `failed`, and `canceled` as terminal, so it would admit a failed step's unfinalized outputs — the very thing the sibling rule exists to reject.

## Supporting changes

- **Seed the collector with `dependsOn`.** The spec always required it; it was never passed, leaving the declared-dependency branch unreachable. This is a durable DBOS workflow-input shape change, so the field is optional and its absence degrades fail-closed for in-flight recovery.
- **Scope inotify to the step's own run tree** (second commit). Dropping `IN_OPEN` left inotify reporting only mutations, and neither `data/` nor a completed sibling can emit those — so watching them registered watches that structurally cannot fire. `data/` was also the only unbounded term in the walk.
- **Decouple `PROVENANCE_DATA_PREFIXES` from the watch dirs**, so the process-local hooks keep full read coverage while inotify narrows. The ordering matters: narrowing the watch dirs first would have silently killed `data/` input lineage.
- **Stop treating a bare `IN_OPEN` as a read.** It fires on opens for writing and on processes unrelated to the command, and inotify cannot attribute an event to a process.
- **Bind the OTel instruments lazily.** `lib/metrics.ts` created its counters at module load, before any `MeterProvider` is registered, and the metrics API has no proxy upgrade — so all three counters, including the two pre-existing ones, were permanently no-op. Without this the spec's "SHALL increment `lineageEdgeRejected`" is unsatisfiable in production.
- **Watch budget.** The cap is now configurable via `PROVENANCE_MAX_INOTIFY_WATCHES`, and a kernel `ENOSPC` is reported distinctly from a cap refusal instead of being silently swallowed.
- **Corrected the false premise** in the `exec-provenance-lineage` and `artifact-manifest` spec prose, which deltas cannot reach.

## What this does not fix

Lineage edges already registered before this change — `T4S1`'s three fabricated links to `T2S2` in run `19110b58` among them — are indistinguishable from real edges by inspection. A read-only audit is scoped as explicit, non-gating follow-up (tasks §9). Nothing mutates registered provenance without a decision.

## Accepted limitation

Reads by processes the hooks cannot instrument — a statically-linked binary, which `LD_PRELOAD` does not apply to — are now unobserved by any layer. This is deliberate. Restoring `IN_OPEN` to recover them would also manufacture reads from other containers' opens of *completed* siblings, which the gate admits, so the gate does not contain the damage. An incomplete lineage graph is recoverable by re-running; a fabricated one is not.

## Verification

`tsc` clean · 1397 pass / 1 skip / 0 fail across 135 files · `go build`, `go vet`, `go test` pass including `GOOS=linux` for the linux-tagged watcher · `openspec validate --strict` valid.

New coverage worth noting: a mutation-checked DBOS replay test proving a replayed exec classifies against the originally snapshotted set (replacing `ctx.runStep` with a plain call makes it fail), and an OTel test that registers a provider *after* module import — the regression guard for the no-op counter class of bug.

## Review guide

| What | Where |
|-|-|
| The admissibility rule | `harness/src/provenance/collector.ts` — `classifyReadPath` |
| Replay-safe snapshot | `harness/src/tools/workspace/execute-command.ts` |
| Where edges are refused | `harness/src/provenance/exec-frame.ts` |
| Full rationale, alternatives, risks | `harness/openspec/changes/gate-lineage-on-step-completion/design.md` |
| Normative requirements | `harness/openspec/changes/gate-lineage-on-step-completion/specs/` |

The sandbox image needs a rebuild and version bump before the Go-side changes take effect; the harness-side gate is independently sufficient for the reported failure.

### DIFF
--- a/harness/openspec/changes/gate-lineage-on-step-completion/.openspec.yaml
+++ b/harness/openspec/changes/gate-lineage-on-step-completion/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-21
--- a/harness/openspec/changes/gate-lineage-on-step-completion/design.md
+++ b/harness/openspec/changes/gate-lineage-on-step-completion/design.md
@@ -1,0 +1,288 @@
+## Context
+
+Lineage inputs are captured at runtime by the sandbox provenance layers, classified by
+`classifyReadPath`, accumulated in a step-scoped `ProvenanceCollector`, and content-attested at
+reconcile. An input that cannot be hashed at reconcile is treated as drift and fails the step
+(`reconcile-manifest.ts:187-191`), which is the correct behaviour for a genuine input that
+vanished — it prevents registering a hashless lineage edge.
+
+Two premises in the current design make that fail-fast fire on files the step never consumed:
+
+1. `classifyReadPath` branch 4 turns **any** path under `runs/{ownRunId}/` that is not the
+   step's own directory into `{source: "upstream"}`, scraping the step id out of the path. The
+   spec justifies it with "`dependsOn` drives only topo-sort ordering, not read authorization:
+   a read of a same-run step outside `dependsOn` is still a valid upstream input."
+2. Reconcile defers input hashing to teardown "because inputs are immutable for the step — the
+   analysis tree is mounted read-only."
+
+Premise 2 is false whenever a sibling runs concurrently. The read-only mount constrains *this*
+step's writes; every other step has its own directory mounted read-write and mutates it freely,
+including creating and deleting scratch files. Premise 1 then converts that churn into
+mandatory attestation targets.
+
+Concurrency is normal, not exceptional: the CLI's machine budget is a user-chosen percentage of
+the host (`cli/src/modules/infra/setup.ts:729-748`) and the scheduler starts every
+dependency-satisfied step that fits (`execute-analysis-scheduler.ts:206-208`). The capture scope
+makes it unavoidable: `mount-plan.ts:140` sets `PROVENANCE_WATCH_DIRS` to the entire analysis
+tree, so every step observes every sibling's directory.
+
+The constraint that shapes the implementation is that `execute_command` is
+`executionMode: "workflow"` (`execute-command.ts:103`) — its body runs unwrapped in the DBOS
+workflow body, where a raw database read is non-deterministic across replay.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Make sibling lineage edges admissible only when the producing step actually finished, so a
+  concurrently running step can never appear in another step's lineage.
+- Stop fabricating edges silently. The `T4S1` case — three `T2S2` files registered as real
+  inputs because they happened to survive — must become impossible, not merely rarer.
+- Preserve fail-fast for genuine drift. An admissible input that vanished is still a hard
+  failure; only inadmissible edges are dropped.
+- Keep classification pure and unit-testable. No database access inside `classifyReadPath`.
+- Keep lineage deterministic under DBOS replay: the same run must produce the same edges.
+
+**Non-Goals:**
+
+- Reconstructing correct lineage for already-registered runs. Detecting fabricated historical
+  edges is a separate audit, tracked in tasks, not attempted inline.
+- Changing the `data`, `artifacts`, or `prior` classification branches' behaviour.
+- Making provenance capture exhaustive. Provenance stays best-effort and must never fail an
+  exec; this change only narrows what it is allowed to *assert*.
+- Replacing the capture layers. Narrowing the watch scope and fixing `IN_OPEN` reduces noise at
+  the source but is not what makes the rule correct.
+
+## Decisions
+
+### Gate on step completion, not on `dependsOn`
+
+`dependsOn` is a scheduling input: it says what must finish before a step may *start*, and the
+current spec explicitly declines to treat it as read authorization. That reasoning is sound — a
+step may legitimately read a sibling it did not declare — so tightening the gate to "must be in
+`dependsOn`" would break real lineage.
+
+Completion is the causally correct predicate instead. If Y has not completed, Y's directory
+contains work in progress, and nothing X observed there is a stable artifact X can be said to
+have consumed. This subsumes the parallel case without a separate rule: a step running
+concurrently with X is by definition not completed, so it cannot contribute an edge.
+
+`completed` is the only admissible status. `failed`, `canceled`, and `skipped` are terminal but
+their outputs were never finalized — that is exactly what a `lineage_attestation` failure means
+— so treating them as admissible would let one step's unfinalized output become another's
+attested input.
+
+`dependsOn` is still threaded into the collector, as `exec-provenance-lineage` already requires
+and the implementation never honoured (`sandbox-step.ts:380-383` passes only `stepId`/`runId`,
+leaving the declared-dependency branch dead code). It no longer gates admissibility; it earns
+its place by letting diagnostics distinguish a declared edge from an observed one.
+
+### Snapshot admissibility before the exec, not at reconcile
+
+Reconcile is the wrong moment. It runs after sandbox teardown, so a sibling that was mid-flight
+*while the file was read* may have completed by then, and the edge would pass — registering
+precisely the fabricated lineage this change exists to prevent. The `T4S1` case would still slip
+through a reconcile-time check.
+
+The snapshot is therefore taken **before the exec is submitted**, and submit time — not start
+time — is the normative predicate. Submission precedes start, so a set observed at submit is a
+subset of what is completed at start; the rule is deliberately stricter than the causal minimum,
+and the strictness is the point. Completion is monotonic, so a step observed `completed` at
+submit time was necessarily completed before any read that exec performs. The converse is deliberately conservative: a sibling that completes *during* the exec
+is treated as inadmissible for that exec, because we cannot tell whether the read landed before
+or after its final write. Dropping a marginal edge is strictly better than registering a racy
+one, and the drop is logged so the choice is visible rather than silent.
+
+The snapshot is per-exec, not per-step, so a long-running step naturally picks up siblings that
+completed since its previous command.
+
+Alternative considered: timestamp comparison, using the frame's per-read times against each
+sibling's `completed_at`. Rejected — it buys precision we cannot trust, since the read timestamps
+come from capture layers whose clocks and ordering are not guaranteed to agree with the
+database's, and it would make the rule depend on the least reliable field in the pipeline.
+
+### Take the snapshot inside `ctx.runStep`
+
+`execute_command` declares `executionMode: "workflow"`, so its body is *not* wrapped in a
+durable step; it runs directly in the DBOS workflow body. A bare `queryStepsByRun` there would
+re-execute on every replay and return whatever the table says at replay time — a strictly larger
+completed-set than the original execution saw. The same run would then produce different lineage
+on recovery, which breaks both determinism and the audit value of the record.
+
+The snapshot must therefore go through `ctx.runStep`, which checkpoints the result so replay
+returns the original answer. The house rule from `lib/result.ts` applies at that boundary: a
+`Result` error crossing a DBOS step must throw (`unwrapOrThrow`), or the step is durably cached
+as a success and replayed as one forever.
+
+### Inadmissible reads are dropped, not fatal
+
+An inadmissible read is noise, not drift — the step observed activity it did not consume. The
+codebase already has the precedent and the mechanism: out-of-tree reads (`reconcile-manifest.ts:169-181`)
+and directory reads (`196-206`) both call `collector.dropInput` and continue. Inadmissible
+sibling reads join them.
+
+The drop happens at classification, before the ref enters the collector, so the inadmissible
+path never becomes an attestation target in the first place. `classifyReadPath` stays pure: it
+receives the completed-step set as an argument and returns an explicit "not an admissible edge"
+outcome, which `feedExecFrame` honours by not calling `trackInputAccess`.
+
+### Observability is part of the contract
+
+Every rejection logs through the injected `Logger` seam with the ref path, the scraped step id,
+and that step's observed status, and increments a new `lineageEdgeRejected` counter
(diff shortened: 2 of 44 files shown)


vivere-dally:
@claude review

vivere-dally:
@claude review

vivere-dally:
# Review: PR #183 — gate lineage edges on producing-step completion

- **PR**: https://github.com/inflexa-ai/inflexa/pull/183
- **Branch**: `bugfix/gate-lineage-on-step-completion` → `main`
- **Head reviewed**: `73cfc4c816ee88f73dcdd89fa97be23ec8d0554e`
- **Size**: 40 files, +3901 / −141
- **Reviewed**: 2026-07-21

---

## Verdict

The core fix is correct and genuinely well-tested. **One issue should block merge**: the second commit's watch-dir narrowing is not safe to ship ahead of the sandbox image, and it silently destroys the `data/` lineage that works today.

The PR's own verification claims were checked independently and are accurate:

| Check | Result |
|-|-|
| `bunx tsc --noEmit` (harness) | clean |
| `bun test` (harness, full) | 1397 pass / 1 skip / 0 fail across 135 files |
| `go build ./...`, `go vet ./...`, `GOOS=linux go vet ./...` | clean |
| `go test ./...`, `go test -race ./...` | clean |

The central admissibility rule holds up. The premise it rests on was confirmed: in `sandbox-step.ts`, reconcile and artifact registration are steps (4–8), teardown is (9), and `status = 'completed'` is written at (10) — so `completed` genuinely implies finalized outputs. Prefix matching uses trailing slashes correctly, so `S1` cannot swallow `S10`. The spec prose corrections in `harness/openspec/specs/` match the code.

---

## Blocking

### 1. New harness + old sandbox image drops every `data/` and cross-step input read

`harness/src/sandbox/mount-plan.ts:169,177` now sends a narrowed `PROVENANCE_WATCH_DIRS` plus a new `PROVENANCE_DATA_PREFIXES=/{analysisId}`.

The old server never reads that second variable — at `main:images/sandbox-base/server/provenance.go:199` it *derives* the hooks' prefix from `watchDirs` instead:

```go
env = append(env, "PROVENANCE_DATA_PREFIXES="+strings.Join(pt.watchDirs, ":"))
```

And the derived value wins. `main:images/sandbox-base/server/executor.go:489` sets `cmd.Env = sanitizedEnviron()` (carrying the container's `/an-1`), then `main:images/sandbox-base/server/executor.go:172` appends the tracker's derived value. Go's `os/exec` keeps the last duplicate for a given key.

**Consequence**: on a cached image the hooks' prefix filter collapses to the step's own directory, and every `data/` read — the primary input class — produces no lineage edge at all.

**Nothing catches it**: `grep -rn "serverVersion\|protocolVersion" harness/src/sandbox/*.ts` returns nothing — there is no version handshake. Task 6.9 explicitly defers the image rebuild, so harness and image ship on different cadences.

This contradicts the PR body's closing claim that "the harness-side gate is independently sufficient." That is true of the *gate* (commit `f428ca8`), which fixes the reported crash on any image. It is not true of the mount-plan narrowing in commit `73cfc4c`, which is a capture-cost optimization that hard-depends on the new server.

**Suggested fix**: hold the `mount-plan.ts` change until the image ships, or gate it on a server version reported in the frame.

---

## Should fix before merge

### 2. `inotify_init1` failure still reports a clean capture

`images/sandbox-base/server/provenance_inotify_linux.go:55-59`:

```go
fd, err := unix.InotifyInit1(unix.IN_CLOEXEC | unix.IN_NONBLOCK)
if err != nil {
    log.Printf("[provenance] inotify init failed: %v", err)
    return
}
```

The error is logged inside the container and the function returns with `exhausted=false, failedWatches=0`. `budget()` at `provenance_inotify_linux.go:117-119` then returns nil, `watchBudget` is omitted from the frame, and the harness logs nothing.

This is precisely the swallow task 10.5 set out to close, one call earlier. It is also the likelier exhaustion: one inotify instance is created per exec against a default `max_user_instances` of 128. `EMFILE`/`ENFILE` take the same path.

### 3. `failedWatches` counts any errno, not `ENOSPC`

`images/sandbox-base/server/provenance_inotify_linux.go:83-99`. The comment there asserts the kernel-ceiling meaning, and `harness/src/tools/workspace/execute-command.ts:256-264` turns the count into "raise `fs.inotify.max_user_watches` on the node."

But the walk runs over the step's *writable* tree, so a temp directory removed between `readdir` and `add_watch` yields `ENOENT`, counts as a failed watch, and sends the operator to the wrong sysctl on an otherwise healthy exec. `EACCES` behaves the same.

Needs `errors.Is(err, unix.ENOSPC)` or a carried errno. The cap-vs-registration-failure split the PR claims is real; the ENOSPC-vs-anything-else split is not.

### 4. The inserted durable step shifts DBOS function IDs

`harness/src/tools/workspace/execute-command.ts:229` adds a `ctx.runStep` immediately before `sandbox.submit-exec`.

DBOS keys checkpoints by `(workflowID, functionID)` — a positional counter (`node_modules/@dbos-inc/dbos-sdk/dist/src/system_database.d.ts:248`) — and `#runAndRecordResult` in `system_database.js` throws `DBOSUnexpectedStepError` when the recorded name at that ID differs from the expected one.

Any in-flight `sandbox-step` workflow that already ran one `execute_command` therefore cannot be resumed or forked onto this code. Auto-recovery is app-version-gated (`getPendingWorkflows(executorID, appVersion)`), so those workflows strand as PENDING rather than corrupt — loud, not silent.

Worth noting because the PR body presents in-flight recovery as handled via the optional `dependsOn` field; that covers the input-shape half only. A drain note in the release would close it.

### 5. The new Go tests never run anywhere

There is no `setup-go` or `go test` in `.github/workflows/` (14 workflow files, none Go). And `images/sandbox-base/server/provenance_inotify_linux_test.go` is `//go:build linux`, so it is excluded on darwin — `go test ./...` there executes 84 subtests, none from that file.

All 122 new lines asserting the cap, the budget signal, and the mutation-only mask are currently unexecuted by both CI and local runs.

---

## Worth fixing, non-blocking

### 6. `PROVENANCE_WATCH_DIRS=""` is ambiguous across the boundary

The harness emits `""` for a read-only sandbox meaning "watch nothing" (`harness/src/sandbox/mount-plan.ts:169`), while `images/sandbox-base/server/provenance.go:420-422` reads `""` as unset and substitutes `/data`:

```go
raw := os.Getenv("PROVENANCE_WATCH_DIRS")
if raw == "" { raw = "/data" }
```

Benign only because nothing creates `/data` in the image, so the existence filter drops it. `os.LookupEnv` expresses the intent. No Go test pins the `"" → []` case.

### 7. Snapshot checkpoint cost scales as analysis size × exec count

`queryCompletedStepsByAnalysis` (`harness/src/state/step-executions.ts:224`) returns every completed `(runId, stepId)` pair in the analysis, and the whole array is JSON-checkpointed into `dbos.operation_outputs` on **every** `execute_command`.

For an analysis with ~1000 completed pairs and a step running 200 commands, that is several MB of system-DB writes for one step. The snapshot must stay per-exec (submit-time is the normative predicate), but the checkpoint payload could be compacted.

### 8. The metrics fix moves the bug rather than removing it

`harness/src/lib/metrics.ts:20-46` memoizes whatever `getMeter` returned, noop included — so a first record before provider registration caches a noop meter permanently, with only a comment at `metrics.ts:91-92` enforcing the ordering.

Separately, the OSS CLI registers no `MeterProvider` at all (no `setGlobalMeterProvider` anywhere in `cli/src`), so `lineageEdgeRejected` stays no-op there and the spec's "SHALL increment" is satisfied only for a managed embedder.

The underlying premise checks out and the fix is warranted: installed `@opentelemetry/api` is 1.9.1, whose `build/src/` ships `ProxyTracer`/`ProxyTracerProvider` but no proxy meter of any kind — `getMeter` is a bare passthrough to `NOOP_METER_PROVIDER`. All three counters really were permanently no-op.

### 9. Tasks §8.5 and §8.6 are unchecked and uncovered

No end-to-end reproduction of the original two-concurrent-steps shape, and nothing tests the boundary the design resolves conservatively — refused on this exec, admitted on the next. The unit and replay coverage is otherwise strong.

### 10. Minor

- `PROVENANCE_MAX_INOTIFY_WATCHES` has no producer in the harness — `buildMountPlan` doesn't emit it and there is no config field. The only route is `SandboxStepInput.extraEnv`, which is undocumented as a deployment knob.
- `normalizePathList` (`images/sandbox-base/server/provenance.go:451-464`) appends `/` but never `filepath.Clean`s, while `recordOp` cleans the incoming path before comparing. Unreachable from `buildMountPlan` today (`assertSafeId`), but one `filepath.Clean` closes it.
- `trackInputAccess`'s 3-arg overload is production-dead — only tests call it — and always refuses producing-step reads. A future caller would silently lose lineage.
- Refusals log one `warn` per path in `feedExecFrame`. A command reading many files under a running sibling could flood; a per-frame aggregate would be quieter.
- `TestInotifyStart_AppliesTheConfiguredCap` (`provenance_inotify_linux_test.go:93-95`) runs a live `readLoop` against a nil tracker; a create/delete under `root` during the test body would nil-deref in a non-test goroutine.

---

## What's genuinely good

**The replay test is a real regression guard, not a shape test.** `harness/src/tools/workspace/execute-command.replay.test.ts` drives the tool inside a registered DBOS workflow, self-cancels, marks a second sibling completed between attempts, resumes, and asserts the late sibling never becomes an edge. It also reads `dbos.operation_outputs` directly to confirm the degraded outcome was itself checkpointed as `"ok":false`.

**The metrics regression guard was empirically confirmed** — reverting `metrics.ts` to eager binding while keeping the exported API identical produces 4/4 failures.

**The `feedExecFrame` tests assert exact metric attributes and log fields**, so they could not pass against the old code.

**The fail-closed reasoning is consistent throughout** — `undefined` snapshot treated as "no observation" rather than "empty set", refusal as a value rather than a throw, and the refusal happening before `trackInputAccess` so an inadmissible ref is unrepresentable rather than filtered later.

**Read coverage for `data/` is genuinely preserved on a matched image**: `provenance.go:267-269` emits the mount root, `recordOp`'s admissibility bound moved from `watchDirs` to `dataPrefixes`, and since pre-PR `watchDirs` *was* the mount root, the boundary is unchanged.

**No type-safety escape hatches** were introduced in production code — zero `as`/`any`/`@ts-expect-error` in the added lines of `metrics.ts`, `mount-plan.ts`, `types.ts`.

**No data races** in the Go watcher: the five new fields are written only on the executor goroutine, `w.wds` is fully populated before `go w.readLoop()`, `pt.mu` still covers every `pt.ops` access, and `-race` passes.

---

## Next step

Decide on item 1 — either drop `mount-plan.ts` from this PR or add the server-version gate. Everything else can land as follow-ups.

vivere-dally:
Will be handled in a different PR

Issue #187, opened 2026-07-21
Lineage admissibility: enforce at the mount, and collapse the three evaluation clocks (follow-up to #183)


vivere-dally:
> Follow-up to #183. **This is not a request to change or block that PR** — the rule it introduces is
> correct and should land. This issue argues that the rule is enforced one layer higher than it needs to
> be, that #183 already computes everything the cheaper enforcement point requires, and that the same
> predicate is currently evaluated on three different clocks.

## Summary

#183 makes an input edge admissible **iff the producing step was `completed` at the moment the reading
exec was submitted**. That predicate is causally correct and the replay reasoning behind it is sound.

The reservation is *placement*. The hazard — a concurrently-running sibling's mutable directory being
readable — remains in the container's namespace, and every layer beneath the mount re-derives which parts
of that namespace were safe to look at. #183 narrows what the **watcher may observe** to the immutable
set. The proposal here is to narrow what the **container may open** to the same set, using the same
enumeration the PR already builds.

Five recommendations, ranked. **R1 is a follow-up, not a blocker** — it changes container topology and
deserves its own blast radius. R2–R5 are small enough to ride with or immediately after #183.

| | Recommendation | Size | Risk | When |
|-|-|-|-|-|
| **R1** | Make the mount the enforcement point | L | Medium (topology) | Follow-up |
| **R2** | One completed-step snapshot per step, not per exec | S | Low | With/after #183 |
| **R3** | Label edges `declared` vs `observed` | S | Low | With/after #183 |
| **R4** | Close the `data/` branch (ungated + fail-fast) | S | Low | With/after #183 |
| **R5** | Reconsider fail-fast on a vanished *admissible* input | S | Policy call | Needs decision |

---

## 1. Ground truth: the invariant #183 rests on holds

The whole design depends on `completed ⇒ the step's tree will never change again`. This was verified
directly rather than assumed, because if it were false the entire gate would be unsound.

In `harness/src/workflows/sandbox-step.ts`, the post-step sequence is:

1. `walkStepArtifacts` — build the draft manifest
2. `generateStepFileMetadata` (DBOS step)
3. **`generateStepSummaryAndWrite`** — *writes `output/summary.md`* — `sandbox-step.ts:734`
4. `reconcileAndRegisterStepArtifacts` + `artifactRegistry.sync`
5. `vectorIndexStepOutputs`
6. `sandboxClient.teardown`
[...]

PR #188, opened 2026-07-21
fix(harness): refuse lineage edges to undeclared same-run siblings

vivere-dally:
## What & why

A step could be killed by bookkeeping about a file it never read.

In run `19110b58`, step `T2S2` completed its analysis — every output, figure and summary written — then failed with `errorClass: lineage_attestation` because `runs/<runId>/T5S1/output/_ct_for_r_BRAF.csv` was absent at reconcile. That file was a scratch file belonging to `T5S1`, a step running concurrently in a separate container, which created and deleted it. `T2S2` had no dependency on `T5S1` and could not have consumed it. The failure was symmetric — `T5S1` was simultaneously failing to attest `T2S2/output/gseapy_hallmark`.

**The crash is the lucky half.** In the same run, `T4S1` recorded three of `T2S2`'s files (`logs/run_gsea.log`, `output/wikipathways_symbols.gmt`, `scripts/run_gsea.py`) as its own inputs while `T2S2` was still running. They survived to reconcile, hashed cleanly, and were **registered as real lineage edges**. A survival-analysis step did not read a GSEA log. Nothing goes red; the provenance record is simply wrong.

Two premises combined to cause it. `classifyReadPath` turned *any* same-run path outside the step's own directory into a mandatory, content-attested `upstream` input — justified by *"a read of a same-run step outside `dependsOn` is still a valid upstream input."* And reconcile deferred input hashing to teardown *"because inputs are immutable for the step — the analysis tree is mounted read-only."* Read-only bounds what **this** step writes; every sibling has its own directory mounted read-write over the same host inodes and mutates it freely. Concurrency is normal, not exceptional — the scheduler starts every dependency-satisfied step that fits the machine budget.

### The rule

**A same-run sibling contributes a lineage edge only when it is a declared `dependsOn`.** Anything else under the step's own run is refused: no `InputRef`, no attestation target, no registered edge, and a log line naming the producing step so the drop is never silent.

Every branch under `runs/{ownRunId}/` decides off **one segment** — the first one after the run id, compared by equality. Not a `runs/{ownRunId}/{id}/` prefix: a trailing separator is not guaranteed to be present, because an `opendir` of a step directory is reported by the capture layers as the bare name. Under per-branch prefixes each such directory read misses its own branch and lands on the refusal, so a step refuses an edge to *itself* on every listing of its own tree, and does the same for a dependency it did declare. Segment equality also keeps an undeclared `qc2` from borrowing the verdict of a declared `qc`.

### Why declaration is sufficient here, with no runtime completion query

`scheduleReady` (`execute-analysis-scheduler.ts`) admits a step only when `step.depends_on.every((d) => completedSet.has(d))` — so every declared dependency is `completed` before the reading step starts, therefore before it submits any exec.

Completion in turn implies a frozen tree. In `sandbox-step.ts` the post-step order is: walk artifacts → file metadata → **`generateStepSummaryAndWrite` (the last write, `output/summary.md`)** → reconcile + register → sync → index → teardown → **`mark-complete` sets `status: "completed"`**. The status flips after the final write.

Composing the two: a declared dependency's directory is immutable for the reading step's entire lifetime, so its edge can be asserted unconditionally. That is what keeps this PR to 5 source files instead of a per-exec database read, a durable DBOS checkpoint to keep replay deterministic, a snapshot-unavailable failure arm, and metrics plumbing.

### Seeding `dependsOn` is load-bearing, not a tidy-up

`sandbox-step.ts` constructed `new ProvenanceCollector({ stepId, runId })` — `dependsOn` was never passed, so the declared-dependency branch was unreachable. The `exec-provenance-lineage` spec already required the seeding.

The two halves are not independent: refusing the sibling branch **without** seeding `dependsOn` would send every same-run read — declared dependencies included — into the refusal and delete all legitimate same-run lineage. The seeding is what makes the refusal safe. It is therefore projected from the parent's keyed plan snapshot (`planStepById`) and **thrown for** when the step is absent, never defaulted: an empty list is indistinguishable from a step that genuinely declared nothing, so a silent default would delete every edge the step was entitled to and leave no record that it did.

`SandboxStepInput` is DBOS workflow input, so the field is added **optional** and treated as an empty list when absent. A workflow recovered under the old shape under-captures every same-run **cross-step** edge — declared dependencies included, not only siblings — rather than admitting unstable ones. Fail-closed, consistent with the direction of the change. The blast radius is bounded: children already dispatched when the deploy lands keep their old-shape input for the rest of their lives; children dispatched afterward carry the field.

### Trade-off

An undeclared sibling that had **genuinely completed** no longer yields an edge. This is deliberate under-capture: an incomplete lineage graph is recoverable by re-running, a fabricated one is not. Each refused path is logged once per step, so the loss is visible rather than silent — and once per step rather than once per read, because a refusal reports one fact that does not change when the same file is read again, while reads dedup within a frame but not across the many execs one step issues.

**Reconcile's fail-fast is retained on purpose.** Once phantom sibling edges stop being asserted, an `input-enoent` means a *declared* dependency's file vanished — genuine drift that should fail loudly. Softening it here would re-hide the silent half.

## Scope

Harness-only. **No sandbox image rebuild, no Go changes, no database schema or query, no new DBOS step, no mount or capture-layer changes.** This narrows only what the harness is willing to *assert* from a frame, not what the container observes — so it is deployable on the current image.

Deliberately out of scope, tracked in #187: gating undeclared-but-completed siblings on a runtime completion snapshot; gating `prior`-run reads; narrowing the inotify watch scope and dropping bare `IN_OPEN`; the ungated `data/` branch; reconcile's fail-fast policy; declared-vs-observed edge labelling; narrowing the container mount itself.

Also deferred there: admitting a step's **transitive ancestors**. They are provably frozen by the same induction this PR relies on — a step is admitted only once its dependencies are `completed`, and `completed` is add-only — so no snapshot is needed to justify the edge. It is deferred anyway because it is a design question, not a mechanical gap: if `C` reads `A`'s output without declaring `A`, the plan's DAG misstates the data flow, and refusing the edge is what makes that visible. Whether the right answer is to attest it or to fix the planner belongs with the rest of the classification policy.

Related: #187 (structural follow-up and residual risks) · #183 (broader alternative — its per-exec snapshot is precisely what #187 argues should collapse to one snapshot per step).

## Verification

`tsc` clean · **1376 pass / 1 skip / 0 fail** across 134 files · `openspec validate --specs --strict` **62/62** · all changed `src/` files lint clean and Prettier-formatted.

Three caveats, stated so neither a red result nor a green one is misread:

- **`bun test` needs a container runtime.** Without a reachable Docker socket, the Postgres/DBOS testcontainer tests fail on startup alone — none of them in files this PR touches. The numbers above are from a run with `DOCKER_HOST` pointed at a live podman socket.
- **`bun run lint` fails on two pre-existing errors** in files this PR does not touch (`src/providers/ai-sdk.test.ts:346`, `src/tools/sandbox/list-available-refs.ts:394`, both last modified by df0591e on `main`), plus a config error on `scripts/smoke.mjs`. Linting this PR's own files reports clean.
- **A green `lint (harness)` check does not mean ESLint ran.** `.github/workflows/lint.yml` states that `harness/` "has no ESLint config, so its lint is the tsc typecheck", and the job body runs only `bunx tsc -p tsconfig.json --noEmit`. That premise is stale — `harness/eslint.config.js` exists and `harness/package.json` declares `"lint": "eslint ."`, resolving 97 rules including `neverthrow/must-use-result` and `no-console`. So the ESLint pass reported above is local-only; CI does not gate it for `harness/`. Pre-existing, and out of scope here.

### Coverage worth noting

`classifyReadPath` had **zero** test coverage before this — the existing collector tests only exercised `recordFileToolWrite`, which is how a five-branch classifier changed meaning without a single test objecting. It now has a test per branch, including a guard for the case where a declared dependency's path is *also* matched by the sibling branch; if that ordering ever inverts, every legitimate same-run edge silently becomes a refusal. Bare directory names and prefix-colliding step ids are covered as their own cases.

The seeding is tested across the join, not just on either side of it. `createLineageCollector` is split out of the step body so a test can drive **workflow input → collector → classification** in one assertion; previously the projection was tested in the parent and the classification in the collector, and deleting the one line that connects them left every layer's own tests green while all same-run lineage vanished.

Two reconcile-level tests carry the real regression guarantee:

1. A vanished undeclared sibling read no longer fails the step — the exact `T2S2` shape, driven through `reconcileManifestWithDisk`, the function that actually threw in production.
2. **The same read, if tracked, is still terminal** — forces the ref in via an explicit context and asserts `rejects.toThrow(/cannot attest input/)`. This exists so the first test cannot pass vacuously: together they prove the fail-fast path is live and that the *refusal* is what avoids it, rather than a weakened guard.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — with the caveats documented under **Verification**.
- [x] Documentation updated if user-facing behavior changed — the `explicit-input-classification` and `exec-provenance-lineage` specs are updated, and the false immutability premise is corrected in the `exec-provenance-lineage` and `artifact-manifest` Purpose prose (requirement deltas cannot reach Purpose text).

- [x] **Methods:** no analytical method changes — this touches lineage bookkeeping only. No tool or library versions change.
- [x] **Sandbox:** the security model is untouched. No change to the workload uid, capability set, `no-new-privileges`, egress firewall or NetworkPolicy, exec-endpoint authentication, mount layout, resource limits, or spawned-command environment. Nothing is loosened; no sandbox or image file is modified.
- [x] **Provenance:** prior analyses remain reproducible, and the expected variance **is** documented: an undeclared same-run sibling read that previously produced an `upstream` edge now produces none. Declared-dependency, `data`, `artifacts` and `prior` edges are unchanged. The change only removes assertions that were unsound — a concurrently-running sibling has no ordering guarantee, so an edge to it was never reproducible in the first place. Historical fabricated edges already registered (for example `T4S1`'s three links to `T2S2` in run `19110b58`) are **not** rewritten by this PR; detecting them is scoped as a read-only audit in #187, since nothing should mutate registered provenance without a decision.


vivere-dally:
@claude review

vivere-dally:
@claude review

krajla (approved):

PR #194, opened 2026-07-21
feat(harness): retry transient provider failures with capped backoff

vivere-dally:
## What & why

A provider that fails to respond (connection refused/reset, DNS failure, 429/5xx) exhausts the AI SDK's hardcoded 2 retries in ~6 seconds and surfaces a failure — a brief provider outage or local-proxy restart breaks an agent turn or a durable workflow step immediately. The SDK's `generateText`/`streamText` expose only a retry *count*: the 2s initial delay and ×2 factor are hardcoded with no delay cap, so raising the count alone produces unbounded tail waits (~68 min cumulative at 10). Retry also cannot live at the DBOS layer: provider failures return as `err()` Results, which a durable step caches as *success* — step retries never see them. The provider seam is the only layer that can own this.

This PR wraps each wire attempt in `harness/src/providers/ai-sdk.ts` in a harness-owned retry envelope built on `retryWithExponentialBackoff` from `@ai-sdk/provider-utils` (promoted from transitive to direct dependency):

- **Policy**: 10 retries, 2s initial delay, ×2 backoff, every individual delay capped at 30s with full jitter — worst case ≈ 3.5 minutes. A server-named `Retry-After`/`retry-after-ms` is honored exactly when it parses within the cap.
- **Retryability** is decided by the harness's own `classifyProviderError`: 429/5xx/connection-level failures retry; `auth` (401), `budget` (402), `tenant-blocked` (403), and other concrete 4xx short-circuit after one attempt. The SDK's internal retry is disabled (`maxRetries: 0`) so attempts don't multiply.
- **Abort** interrupts the backoff sleep immediately; **exhaustion** keeps the last real failure on the `cause` chain so `ProviderError` classification still keys on the true HTTP status; **billing headers** re-resolve through the `ResolveBilling` seam per attempt.
- **Streaming** retries establishment only, covering both SDK failure-surfacing paths: a stream that errors before any delta rejects the first `textStream` pull, while a `doStream` promise rejection (the shape a real connection failure takes) is deferred by `streamText` past a clean `{done: true}` first pull — so a `done` first pull awaits the deferred result promises *inside* the envelope, and a genuine text-less turn still completes without a spurious retry. Once a delta reaches the consumer, errors propagate unchanged and text is never re-emitted.

The `harness-providers` spec gains the bounded-retry requirement with 5 scenarios; the OpenSpec change (`add-provider-retry`) is synced and archived in this PR. 16 new tests cover the delay function, chat retry/short-circuit/exhaustion, billing freshness, abort-during-backoff, and the streaming paths — fast retryable-exhaustion tests drive delay through a real `retry-after-ms: "0"` header so the header-honoring path is exercised for real.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`).*

\* `tsc -p tsconfig.json` clean; full `bun test` green (1392 pass, 1 pre-existing key-gated skip, 0 fail). Two lint issues exist at `main` and are untouched by this branch: `eslint .` crashes on `scripts/smoke.mjs` (typed rule applied to an `.mjs` without parser type info), and `src/providers/ai-sdk.test.ts` carries one `neverthrow/must-use-result` error (line 346 at `main`). The files changed here introduce no new findings (`eslint` on both changed source files verified).


vivere-dally:
@claude review

radutanasa (dismissed):

radutanasa (approved):

Issue #197, opened 2026-07-22
Closing the loop: bringing harness report generation up to the artifact-authoring loop


vivere-dally:
# Closing the loop: bringing harness report generation up to the artifact-authoring loop

**Status:** proposal — nothing in this document has been implemented.
**Scope:** `harness/src/tools/report/`, `harness/src/execution/report-{runner,render}.ts`, `harness/src/tools/iterate-report.ts`, `harness/src/prompts/report-builder.ts`, `templates/report-html/`, plus two CLI seams.
**Verified against:** working tree at `69e76b2`. Harness report tests 22 pass / 0 fail; `tsc -p tsconfig.json` clean. Every claim below was read out of the source rather than inferred; see the verification log at the end.

---

## 1. Why this document exists

While building a teaching artifact about this subsystem, I ended up running an unintentional control experiment: the same task — turn an analysis into a themed, self-describing HTML document — executed twice, by two different loops.

- **The harness's loop.** A brief crosses into a report-builder agent that cannot see the analysis, cannot search the filesystem, and (in the OSS build) cannot see the page it just wrote. It finishes by calling a gate that checks three things.
- **My loop.** Same task, but I rendered the page, drove headless Chrome against it, screenshotted it, and ran programmatic assertions over the result.

My loop found four real defects after I had already published once and believed the page was finished:

| Defect | Severity | Would the harness gate catch it? |
|---|---|---|
| Dark-mode active chip at 2.3:1 contrast (white on light pink) | Unreadable control | No |
| `--ink-3` text tier at ~4.0:1 across both themes | Below WCAG AA | No |
| Mobile grid track resolving to 477px inside a 325px box | Content clipped | No |
| Hero overflowing its container by 11px | Layout break | No |

All four pass `submit_report` trivially. The file existed and was non-empty, there were no unrendered Jinja markers (there was no Jinja at all), and every local reference resolved — my page had zero external references.

The mobile defect is the instructive one. Because the container had `overflow: hidden`, the blowout was *clipped* rather than pushed onto the page, so even a document-level overflow check would have reported clean. I only found it by measuring the computed grid track after a screenshot looked wrong.

**This document asks: what would it take for the report-builder's loop to catch what my loop caught?**

---

## 2. What my loop had that the builder's loop does not

| Capability | My loop | Report-builder | Gap addressed in |
|---|---|---|---|
| Render the document | Yes | Yes (`build_report`) | — |
[...]

PR #199, opened 2026-07-22
fix(harness): make the report tool surface match its implementation

vivere-dally:
## What & why

The report path had accumulated a set of claims the code never honoured. Prompts and tool descriptions are the one layer no typechecker validates, so each stayed green until an agent acted on it — and then failed silently, produced mislabelled output, or burned turns.

A user hit the worst case. Their conversation agent composed a valid report brief, was told the report "must be an object, not a JSON string", retried four times reasoning correctly that this was a tooling problem, then gave up and advised filing a bug against `anthropics/claude-code`. The brief was recoverable on every attempt.

**Root cause.** `submit_report` types its `report` field as `z.unknown()` to keep the ~12k brief schema off every conversation turn. The loop's automatic repair for double-encoded arguments (`repairToolInput`) is keyed on the tool's input schema *failing* — and a JSON string satisfies `z.unknown()`. So the schema succeeded, the loop returned early at `run-agent.ts:268`, and the repair at `:274` was never reached. The real schema rejected it later at a site that only emits a hint. The hint was accurate and unactionable: the model was told to pass an object directly, believed it already had, and resubmitted verbatim.

This is worth noting for the rework in #197, because the asymmetry is invisible until you look for it. `z.unknown()` is otherwise used only by the dry-run *validators*; every *committing* tool puts its real schema on the wire:

| Tool | Wire schema | Loop repair fires? |
|---|---|---|
| `submit_plan` | `PlannerPlanSchema` | yes |
| `submit_synthesis` | `RunSynthesisSchema` | yes |
| `submit_report` | `z.unknown()` | **no** |

Triaging that turned up more of the same class, which this PR closes as a pre-ship hardening pass.

Related to #197 — this does **not** close it. The larger rework tracked there (brief-aware gate, design-system conformance, Chrome sidecar, preview audit) is deliberately deferred and recorded as such in the archived change.

## What changed

- **A double-encoded brief is repaired instead of trapping the agent.** `acceptBrief` runs the same issue-guided repair the loop performs, before re-validating. When a brief decodes but is then genuinely invalid it reports the *decoded* value's issues, so the model isn't sent chasing a problem already solved.
- **`format` no longer accepts `"pdf"`.** It was accepted, threaded into the builder's instruction as `Build a new PDF report.`, never read by the build path, then persisted and emitted so the UI labelled HTML as a PDF. Persisted `"html" | "pdf"` unions stay wide so previews written earlier still parse.
- **Iterating an unknown `previewId` is refused** before minting access, staging assets, or starting the builder. It previously produced a fresh `v1` while telling the builder the prior template was "already in your working directory" — with no brief either, leaving it a change request and nothing to change.
- **The version filesystem rejects absolute paths** rather than stripping the leading slash. Stripping made the containment check unreachable: a deep absolute path landed a real file below the version dir and returned `ok`, after which `build_report` reported the template missing and the builder — which has no directory-listing tool — could not find it.
- **An unavailable preview seam is logged and its message is well-formed.** Every report in the OSS build emitted `preview-access mint failed: status=undefined …` with nothing logged, so nobody could tell the build's only visual verification step never ran. The composition is now shared, so `mint_preview_url` and `preview_snapshot` cannot describe the same failure differently.
- **Four stated capabilities that do not exist are removed**: a `grep` tool on the builder roster, oversized-source "skipping" that is actually a hard failure, a "15-min TTL" that exists nowhere on the path, and a `theme.css` the builder has no tool capable of reading.
- **The copy contract is now pinned by a test.** Review alone is what missed all four — a prior sweep for this exact class (`fea90fb`) did not touch this file. The test scans every `defineTool` id out of the package at runtime and asserts the builder prompt names exactly the roster `createVersionFsTools` returns, so a tool added anywhere immediately becomes a name the prompt may not use unless the builder holds it.

## Notes for review

**Two changes are breaking for an agent mid-conversation** (not for any consumer API): `format: "pdf"` is now rejected at the tool boundary, and an absolute path that previously "worked" (`/index.html`) now returns `out_of_scope`. Both failures are immediate and name the correction, and the builder prompt already mandates relative paths. Tolerating the benign absolute path is what made the harmful one undetectable — both arrive through the same normalization.

**Known spec drift, left in place.** `openspec/specs/iterative-report/spec.md`'s Purpose section and several requirement headers still name `iterate_report`, which was split into `plan_report` + `submit_report` without a spec update. The delta here is written against the shipped names so its own requirements are coherent; reconciling the rest is separate work, recorded on #197. It is not cosmetic — that split is what introduced the `z.unknown()` field behind the bug above.

**`bun run lint` is broken on `main`, independently of this PR.** It exits 2 on `harness/scripts/smoke.mjs` (a type-aware rule running on a file outside the tsconfig project). `eslint src/` runs, and reports 2 pre-existing errors in `src/providers/ai-sdk.test.ts` and `src/tools/sandbox/list-available-refs.ts`. Neither file is touched here, and every file this PR changes is lint-clean.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally — `bun run typecheck` clean; `bun test` 1418 pass / 1 skip / 0 fail (needs `TESTCONTAINERS_RYUK_DISABLED=true`; the plain run has a known Ryuk testcontainer failure unrelated to this change). See the lint note above for the pre-existing `bun run lint` breakage.
- [x] Tests added or updated for the change — 3 new test files, 13 new tests. Each was mutation-checked: reverting the fix makes them fail.
- [x] Documentation updated if user-facing behavior changed — the `iterative-report` spec gains 3 requirements and 2 are rewritten; the change is synced and archived.
- [x] This PR is focused on a single logical change — one hardening pass over the report tool surface, sequenced as fix → follow-ups → sync/archive.


vivere-dally:
@claude review

vivere-dally:
Thanks — both findings were real and both are fixed in d4f702e, along with a third that the review pass missed. Detail below, then a re-review request.

## Your finding 1 — hint computed against the pre-repair value

Correct on mechanism. I measured the impact before fixing, and it is narrower than "worth a look if double-encoded-then-invalid briefs turn out to be common":

```
issue path=["audience"]  expected=string   hint w/ raw: undefined   hint w/ decoded: undefined
issue path=["sections"]  expected=array    hint w/ raw: undefined   hint w/ decoded: undefined
```

For the ordinary decoded-then-invalid brief both arms yield nothing — `repairableExpectation` only accepts `object`/`array`, and a *missing* field resolves to `undefined` rather than a string. The two differ in exactly one shape: an outer double-encoded brief whose nested array/object field is itself a JSON string.

```
issue path=["sections"] expected=array
  hint w/ raw string : undefined
  hint w/ repaired   : "This argument arrived as a JSON-encoded string. Pass the array directly…"
```

Fixed anyway — `acceptBrief` now returns `{ parsed, validated }` and `hintForZodIssue` gets the value the issues are actually addressed to.

## Your finding 2 — auto-generated id in the refusal

Confirmed, and it pointed at something structural. Verified the schema accepts iteration with no id at all:

```
submitReportInputSchema.safeParse({ modifications: "..." })  ->  success: true
```

So the real defect wasn't the message — it was that `modifications` never required a `previewId`. The envelope enforced `report` XOR `modifications` and stopped there. Fixed at the input boundary with a third refinement, which removes the confusing message as a side effect rather than papering over it.

## What the review missed — `baseVersion` bypasses the guard entirely

The guard asks only whether the preview holds *any* version. But `baseVersion` selects a *specific* one, and `report-runner.ts:141-155` resolves it independently and tolerates a missing base template in silence. `baseVersion` is `z.number().int().min(1)` — no upper bound, no cross-check.

A preview legitimately holding only `v1`, iterated with `baseVersion: 7`:

```
guard passed, builder started : true
resulting version             : 8
error                         : (none)
v8 template                   : <<NO TEMPLATE COPIED>>
```

The builder then receives the modification prompt asserting *"The previous version's template is already in your working directory… Do NOT rewrite from scratch"* — with no template and no brief. That is verbatim the defect this PR exists to close, reached through a different door, and it is reachable whenever an agent guesses a version number — which `baseVersion`'s own description invites (*"use when the user prefers an earlier version"*).

Same probe after the fix:

```
builder started : false
version         : 0
error           : Preview 'prv-real' has no version 7 to branch from — its latest version is 1.
                  Retry with a `baseVersion` that exists, or omit `baseVersion` to branch from the latest version.
```

Worth naming why this one slipped: it isn't visible by reading either side alone. It needs the guard's condition and the runner's independent base resolution held together, and the confirming evidence was an executable probe. Your note that `bun install` had no network access, so `bun test` / `tsc` couldn't run, is probably the whole story — a read-only pass is structurally better at finding stated-vs-implemented drift than at finding two correct-looking pieces that don't compose.

## Also changed since your pass

- The prompt copy guard banned *any* duration figure. The requirement is that a **stated access lifetime** match the code, not that the prompt avoid durations — a future legitimate line like "waits up to 8 seconds for theme-ready" would have failed spuriously, and the likely response would have been deleting the assertion rather than scoping it. Now narrowed to lifetime vocabulary near a figure, verified both directions: a reintroduced "valid for 15 minutes" fails, "waits up to 8 seconds" passes.
- The spec drift I'd disclosed as known debt is closed. All seven `iterate_report` references are gone, including the requirement header whose body already said `submit_report`. The Purpose section now also records that the builder's visual self-check exists only where the preview seam can mint a URL.
- The `Iteration against an unknown preview` requirement gained three scenarios covering the `baseVersion` arms and the `previewId` requirement.

## State

1422 pass / 1 skip / 0 fail, `bun run typecheck` clean, eslint clean on changed files. Every new test was mutation-checked — reverting each fix fails exactly its own test and nothing else; the `baseVersion` mutation reproduced the bypass precisely (*expected version 0, received 8*).

Standing caveat, unchanged from the PR body: `bun run lint` is broken on `main` independently of this branch (exits 2 on `harness/scripts/smoke.mjs`, a type-aware rule on a file outside the tsconfig project), and `eslint src/` reports 2 pre-existing errors in files this PR does not touch.

@claude please re-review the PR at d4f702e. Two things I'd especially like a second pair of eyes on:

1. **Does the guard now cover every path to an empty-handed builder?** It refuses when the preview holds no versions and when `baseVersion` names one that doesn't exist. I scoped it to iteration mode only, reasoning that creation always hands the builder a complete brief so a missing base template there is recoverable — please check that reasoning, and whether a version directory that exists but contains no `report.html.j2` is a fourth path worth refusing.
2. **Is the narrowed copy guard in `report-builder.test.ts` still strong enough?** It now requires lifetime vocabulary near a duration figure. I'd like a check on whether a plausible reintroduced TTL claim could slip past that proximity window.

radutanasa (approved):

PR #201, opened 2026-07-22
feat(cli): rework reference selection, progress, concurrency, and sizing in setup

vivere-dally:
## What & why

The reference-data step of `inflexa setup` opened with 32 of the catalog's 56 datasets already ticked, then transferred them in silence. Four changes, all in `cli/src/modules/refs/`.

### Selection is one picker with bulk keys

The grouped per-dataset listing is now the only selection surface. It opens with **nothing** ticked, and three keys each replace the selection outright:

```
◇  32 datasets are already installed and intact — not listed below.
◆  Reference datasets to download                     ╭─ No rush ──────────────────────────────╮
│  ◻ gene-identifiers                                 │  Reference data is only needed once    │
│  │ ◻ NCBI Entrez-to-Ensembl gene mapping (1 file)   │  an analysis asks for it, so taking    │
│  └ ◻ NCBI Entrez-to-RefSeq accession mapping (1 f…) │  none now is a real choice — not a     │
│  ◻ pathways                                         │  skipped step.                         │
│  │ ◻ Reactome identifier mappings (large) (3 files) │                                        │
│  ...                                                │    inflexa refs download <id>          │
│  ↑/↓ move · space toggle · a all · r recommended (32 already installed) · n none · enter confirm
```

Reaching a large set costs one keystroke and so does abandoning one, which is what lets the listing be the only surface without bringing back the wall of pre-ticked boxes. `r` is **inert rather than clearing** when nothing offered is recommended — a key named "recommended" that wipes a selection is a trap, and `n` already means none.

An earlier revision of this PR led with a four-option preset `select` instead. It was removed: a preset names a set without showing it, and — because presets resolve against the datasets setup is *offering*, which excludes anything already installed — the Recommended entry **disappeared** for exactly the users who had installed it. That is the reported bug this now fixes at the root. The legend keeps the key visible and names the cause: `r recommended (32 already installed)`, counting recommended installs rather than every install, since those diverge the moment an optional dataset is installed and only the recommended ones explain that key.

`@clack/prompts`' `groupMultiselect` returns the prompt promise and exposes no instance to bind keys to, and `updateSettings({aliases})` only remaps onto the existing action set — so this constructs `GroupMultiSelectPrompt` from **`@clack/core`** directly with a local `render`. Only the composition is local: glyphs, sliding window, and wrapping stay clack's.

### Transfers report one combined readout

```
Downloading references  ▪▪▪▪▪▪▪▪░░░░░░░  12/38 files · 1.4 GB/8.2 GB · 8.2 MB/s · 4 in flight
```

The installer gained an optional progress reporter on `ReferenceInstallDeps` and a pass-through counting `Transform` inside the existing `pipeline`, so `store.ts` stays free of presentation and backpressure stays with `pipeline`. A throwing renderer can never fail a download. Without a TTY it degrades to one plain line per completed file, so captured logs and the agent's `run_inflexa` subprocess get readable text.

### Datasets install four at a time

A `p-queue` capped at 4 replaces the serial loop. Concurrency spans *datasets* because 38 of 55 carry a single artifact. Two things made it safe: staging is now isolated per dataset (the attempt root appends a dataset discriminator unconditionally — a fixed test attempt id previously gave every dataset one shared root that each would `rm -rf`), and results are written by plan index rather than pushed, so the installed list stays in plan order and a failure reports the lowest-ordered dataset rather than whichever lost the race. Work stops after a failure via a flag checked as each job starts, not `queue.clear()`, whose cleared tasks never settle.

### The plan is sized against the publishers before consent

The catalog pins no sizes, so consent used to quote a file count and the readout printed a byte counter that only climbed — indistinguishable from a total that keeps growing. A `HEAD` sweep now asks the publishers first:

```
◇  Sized 103 of 122 files
Reference download plan: 122 files, at least 21.9 GB to fetch from the upstream publishers. 19 files did not report a size, so the total is a lower bound.
```

Measured against the live catalog: all 122 artifacts answer `200`, 103 declare a trustworthy size (21.9 GB), and the sweep takes **1.6 s**. Probes run through a `p-queue` capped at 20 rather than all at once — unbounded is ~0.5 s faster here and worse from there, since the catalog clusters per host (12 artifacts on zenodo, 11 on `data.broadinstitute.org`) and would arrive as a burst inviting rate-limits from the hosts the download then depends on. Every probe shares one `AbortSignal` started for the sweep, so the step is capped by a budget rather than by artifact count.

`measureReferenceDownload` returns a plain value, **not** a `Result`: a transport error, timeout, refusal, redirect off https, and an untrustworthy header all mean "this artifact's size is unknown", a state the readout already handled. Surfacing any as an error would let a metadata probe fail a download that would otherwise succeed. The sweep is skipped where it informs nothing — a scripted run lacking consent returns first, making no request and keeping the count-only wording scripts match on.

**This also fixed a defect the sweep exposed.** `data.broadinstitute.org` answers `Content-Length: 20551` *together with* `Content-Encoding: gzip`, while `fetch` inflates the body before it reaches the counted stream, so 48690 bytes land — 2.4× the declared figure. `declaredContentLength` accepted that header, and its own comment asserted the case could not arise. It can, on five catalog artifacts, so the in-flight readout could show more received than declared. Any coding other than `identity` is now no declaration at all; the decoded size is not derivable from the compressed one, so rejecting is the only honest option. Verified that Bun preserves `content-encoding` on the inflated `GET`, so the guard fires on the download path and not only on `HEAD`.

### Byte formatting

`Number.prototype.formatBytes()` replaces the refs-local `formatReferenceBytes` and the hand-rolled `written / 1024 / 1024` in the embedding setup step, on a 1024 base with `B`/`KB`/`MB`/`GB` — the vocabulary the rest of the wizard prints. That is the only change here touching a non-refs source file.

**New dependency**: `@clack/core`, pinned to `1.4.3` — the exact version `@clack/prompts@1.7.0` already resolves, so `bun install` reports no change to the tree. Approved for this change. (`p-queue` was already present.)

## Verification

- `bun run lint`, `bun run typecheck`, `bun test` in `cli/` — 1614 pass, 0 fail.
- Declared sizes checked against real downloads across **8 distinct hosts including both that redirect — 8/8 exact**. A live 4-file install predicted 5 369 525 bytes and landed 5 369 525.
- The picker driven under a real pty against the live catalog with `XDG_DATA_HOME` pointed at a throwaway store: `enter` → nothing, `a` → 80 files, `a` then `r` → still 80 (inert, not destructive), `esc` → declined. Piped to a file the output carried zero escape sequences; under a pty it painted the animated bar.
- A store seeded with all 32 recommended datasets reproduced the reported state end to end and showed the installed count plus the annotated key.
- `AbortSignal.timeout` confirmed not to hold the process open (exit in 481 ms, not 15 s).

## Deliberately not in scope

Sizes in the catalog (a pinned size is a second thing to keep true about a mutable upstream; the publisher answers authoritatively at plan time), an ETA, a byte-denominated progress bar (the file count is the one denominator always fully known), resumable downloads (the `If-Range` `TODO(robustness)` stays open), and concurrency *within* a single dataset.

Specs `reference-data-provisioning` and `unit-test-coverage` are updated; four OpenSpec changes are archived (`improve-refs-setup-selection-progress`, `parallel-reference-downloads`, `refs-picker-bulk-keys`, `refs-download-size-sweep`).

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Documentation updated if user-facing behavior changed. `cli/README.md` documents the command surface, which is unchanged — no new commands or flags — so the behavior change is captured in the capability specs instead.
- [ ] This PR is focused on a single logical change. **It is not: four OpenSpec changes landed here — the picker, the progress readout, four-at-a-time installs, and pre-consent sizing. They share one surface and each reshaped the readout the previous one built, which is why they grew together, but they are separable and I'm happy to split the later two into their own PR if you'd rather review them apart.**


vivere-dally:
@claude review

vivere-dally:
@claude review

radutanasa (approved):

PR #202, opened 2026-07-22
feat: turn interrupt affordance + retract-and-edit of the just-sent message

### LINKED ISSUE #198: Interrupt the running turn + retract-and-edit the just-sent message before first output
## Goal

Two chat affordances, one mechanism:

1. **Interrupt** — a first-class, discoverable way to stop the in-flight turn. The cancellation plumbing already exists end-to-end (`conversation.abort()` at `cli/src/tui/hooks/conversation.ts:1045`, per-turn `AbortController` → `runChatTurn` signal → provider `streamText` abort); today it is reachable only as the middle case of the three-way ctrl+c (`cli/src/tui/app.tsx:242`) and via `/quit`-while-busy. This issue gives it its own key, a visible hint, and a defined surface state — it does not invent cancellation.
2. **Retract-and-edit** — while a turn is in flight **and the assistant has produced nothing yet**, let the user take the just-sent message back: interrupt the run, remove the user turn as if it were never sent, and seed the composer with its text for editing. Gated on "nothing produced" precisely because that is the window in which the turn can be made to have never happened — no partial output to keep or explain.

Modeled on how others do it (prior art at the end); ours differs mainly in that our durable history is append-only, which is the one real design question below.

## What exists today

- Abort path: module-private per-turn `AbortController` (`conversation.ts:877,975`), signal threaded through `runChatTurn` (`cli/src/modules/harness/turn.ts:152`) into `runAgent` and the AI-SDK `streamText` call; abort classifies the outcome as `{kind:"aborted"}` (`turn.ts:194-197`) and `finishTurn` flushes what streamed and returns to idle with no error banner.
- Esc is taken while the composer is focused: it focuses the scroll pane / NORMAL mode (`app.tsx:413`); the dialog host owns the structural Esc when a modal is open.
- An aborted turn still persists `[userMessage]` unconditionally to the pg thread (`turn.ts:197,204`), and `ThreadHistory` is append-only — `appendTurn`/`loadRecent`/`loadPage` only (`harness/src/memory/thread-history.ts:68-90`). No retract capability exists anywhere.
- "Nothing produced yet" is derivable, not stored: `chatStatus() === "busy"` ∧ `streamText() === ""` ∧ the in-flight assistant message still holds only its pre-minted empty text part (`conversation.ts:505-516`) ∧ no open tools.
- No composer history/recall of any kind exists (`text_area.tsx` binds only submit/newline).

## CLI — interrupt

- New remappable id in `KEYBIND_DEFAULTS` (`cli/src/tui/keymap.ts:234`), e.g. `app.interrupt`, default **Esc, double-press**: Esc while busy keeps its current meaning (focus the scroll pane — you want to scroll while it streams) *and* arms the interrupt; a second Esc within a short window (others use 5 s) fires `conversation.abort()`. A status-line hint flips `esc interrupt` → `esc again to interrupt` while armed — the hint is what makes the feature discoverable at all.
- Busy-gated layer: when idle, Esc behaves exactly as today. The ctrl+c three-way stays as-is.
- Surface state on interrupt: whatever already streamed stays on screen (current `finishTurn` behavior), and the turn ends quietly — a muted "interrupted" marker on the assistant message rather than an error banner or toast. Interruption is a user action, not a failure.

## CLI — retract-and-edit

- Trigger: while the no-output window holds (definition above), an edit binding — recommend up-arrow in an empty composer, which later degrades naturally into general history recall — runs: `conversation.abort()` → splice the last user `UIMessage` and the pre-minted assistant message from the store (the `resetHotState` / `dropEmptyAssistant` machinery, `conversation.ts:884,569`) → `textareaRef.setText(originalText)` with the cursor at the end.
- Once the first delta or tool part lands, the binding does nothing — the message is answered history. Interrupt still works; retraction doesn't.
- Plain interrupt does **not** retract: an interrupted-but-kept user message stays in history as context for the next turn. Retract is its own deliberate action.

## Harness — durable retract

The UI part is trivial; the thread is not. By the time the user retracts, `appendTurn(threadId, [userMessage])` has already run (`turn.ts:204` executes at abort time), so the pg thread holds an orphan user turn that the edited resend would duplicate. Options:

- **(a) `ThreadHistory` retract capability (recommended):** a minimal harness-first method — remove the last turn / truncate from a message id — invoked by the CLI's retract action. Harness-first per the boundary rule, and it is the primitive any future full undo/redo needs anyway. One constraint to respect: `loadRecent` keeps a byte-stable prefix for prompt caching by advancing in whole eviction blocks (`thread-history.ts:75-80`); deleting the *tail* turn is compatible with that, deleting mid-history would not be.
- (b) Defer/skip the persist: teach `runChatTurn` not to append when the abort was a retract. Rejected as primary — it threads UI intent through the turn engine and leaves no primitive for a future revert.

## Prior art (how others behave)

- Interrupt is Esc, double-press within a window, gated on the session being busy, with a status-bar hint that flips to "again to interrupt" after the first press. Single-press variants exist elsewhere; the double-press exists exactly so Esc can keep a local meaning.
- Abort cancels the run at the request layer (abort signal into the streaming LLM call; running tools get the same signal), keeps and persists whatever partial output streamed, marks the assistant message aborted, and suppresses error toasts.
- The replay rule that motivates our gate: an aborted assistant message that produced **no** content is dropped from the model conversation on the next request — the turn behaves as if it never happened. One that produced real content is kept.
- "Undo last message" implicitly aborts first when busy, places the retracted text back into the composer for editing, and marks a soft revert point whose destructive cleanup happens lazily on the next send. (Full undo there also restores file snapshots — out of scope here; our file mutations live behind sandboxed workflows, a different story.)

## Done when

- Esc-Esc while streaming stops the turn; streamed text stays; no error surface; the status hint shows both stages.
- Send, then edit before any output: the composer holds the original text, the transcript shows nothing from the aborted attempt, and the pg thread contains no orphan user turn (resend produces exactly one user message).
- Esc while idle, Esc in dialogs, and ctrl+c behave exactly as today.

## Out of scope (follow-ups)

- Durably persisting partial streamed text of an interrupted turn (today it survives on screen but not reload).
- Editing/undoing an *answered* message (needs the revert-marker + snapshot model).
- Composer history recall (up-arrow through past prompts) — the retract binding is chosen not to conflict with adding it later.

vivere-dally:
## What & why

Interrupting a running turn existed only as the middle case of the three-way ctrl+c, and taking back a just-sent message did not exist at all — the turn engine persisted `[userMessage]` even on an aborted run, so a mistyped prompt was answered or orphaned, never edited, and an interrupted reply survived only on the live screen. This PR ships interrupt, retract-and-edit, and durable interruption end-to-end:

- **Interrupt as a first-class affordance (cli).** A remappable `app.interrupt` binding — double-press Esc while a turn is busy, counted only when the chat is the main focus in NORMAL mode: presses owned by a dialog, an active text selection, or the composer's INSERT→NORMAL switch never count toward it. The gesture teaches itself from the ChatBar footer, beside the mode word it depends on: `esc interrupt` → `esc again to interrupt` (5-second arm window, warn-colored when armed), and while busy in INSERT the same slot advertises the one-press ctrl+c that already interrupts mid-typing. The status bar carries no turn affordances. An interrupted turn ends quietly — no error banner, no toast — and the ctrl+c three-way chord keeps its dismiss → abort → quit ordering.
- **Interruption is durable (harness + cli).** The streaming provider wrapper resolves a client abort as an `"aborted"` finish carrying the partial already streamed — including a transport failure that races the user's interrupt — instead of re-throwing; the loop returns the partial (the iteration-cap wrap-up call included), stamps a marker that survives storage in the harness `providerOptions` namespace, and the engine persists `[userMessage, …partial]`. A restart therefore renders exactly what the live view showed — partial text with a muted `· interrupted` marker — and the model keeps the partial in context, so "continue" continues. A no-output abort still persists the user message alone and drops the empty assistant shell. Workflow loops run on the non-streaming provider and keep cancellation-by-throw untouched. Display-side, coalescing is now assistant-only (adjacent user rows — the shape an aborted turn leaves — render as the two bubbles they are) and the loop's truncation nudge no longer renders as a user bubble the user never typed.
- **Retract-and-edit before first output (cli).** While the turn has produced nothing, up-arrow — from the pane or the empty composer — aborts the run, awaits settlement, removes the just-persisted orphan user turn from the pg thread via `ThreadHistory.retractLastTurn` (tail-turn-only, transactional, same per-thread lock as `appendTurn`), and seeds the composer with the original text, focused with the cursor at end. Output racing the keypress downgrades to a plain interrupt with a notice; a durable-retract fault self-heals on the next send after re-checking the tail is still a lone orphan.
- **Focus follows the turn (cli).** An accepted submit lands in NORMAL — the resting state of a running turn, where esc-esc, up-arrow, and the scroll keys live — so interrupting is two presses from everywhere; refused submits keep focus and text. A completed retract is the one automatic hand-back to INSERT. The ask-queue drain refocuses busy-aware (pane mid-turn, composer when the turn ended), and turn completion never steals focus.

Specced as four OpenSpec changes (`thread-history-retract` + `abort-preserves-partial-turn` in harness, `chat-interrupt-retract-edit` + `chat-interrupt-persist-flow` in cli) — implemented, verified against their artifacts, deltas synced into nine main specs across the two packages, and archived, all within this PR.

Known gap, recorded in the archived tasks: no TUI e2e exists because the repo has no PTY/tmux e2e harness — key-level dispatch is covered by render tests that register the actual exported layer factories (double-esc arm→fire, pane retract vs scroll precedence, selection exclusion, idle inertness), plus hook-level coverage of every retract path; three App-mount focus slices (accepted-submit → pane, refused-submit keeps text, completion moves nothing) are structural + typecheck only.

Closes #198

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — harness: 1432 pass / 0 fail against schema-scoped Postgres plus a package-boundary consumer check exercising the retract outcomes through the built `@inflexa-ai/harness` barrel; cli: 1341 pass / 0 fail.
- [x] Tests added or updated for the change (harness: `retractLastTurn` scenarios incl. the append/retract race, the abort-resolution matrix incl. the raced-failure and wrap-up cases, marker round-trip + window snap; cli: hook-path unit tests, key-dispatch render tests through the real layer factories, span-color render tests for the footer hint and marker).
- [x] Documentation updated if user-facing behavior changed (the OpenSpec trees are the docs: nine main specs updated across `harness-thread-history`, `harness-providers`, `harness-agent-loop`, `ai-sdk-message-storage`, `tui-harness-chat`, `chat-command`, `key-bindings`, `tui-layout`, `tui-ask-approval`; design-gallery exhibits for the new surfaces).


vivere-dally:
@claude review

vivere-dally:
@claude review

radutanasa (approved):

Issue #204, opened 2026-07-22
Closing the TUI blocks on the in-flight workflow instead of interrupting it — exit should leave the run recoverable and reopening should resume


vivere-dally:
## Summary

Quitting the TUI while a run is in flight does not interrupt the run — it **blocks the process until
the workflow finishes on its own**, with the terminal already restored and no output. There is no
deadline and no escape short of `kill -9`.

The right behaviour is the one DBOS already makes cheap: **stop waiting, exit, leave the workflow
`PENDING`, and let the next launch recover it.** The recovery half already works
(stable `executorID`, see below). Only the shutdown half is missing.

## Observed

2026-07-22, ~2h session, run `7e8cd842-364d-4b09-9641-e08758bbcfe3`.
`~/.local/share/inflexa/logs/inflexa-2026-07-22.log`, pid 78370:

```
15:34:43.126  [shutdown] requested    signal=cli-shutdown        ← ctrl+c
15:34:57.100  prov.run_completed      status=completed
15:34:57.127  [dbos] shutdown         durationMs=13999
```

14.0s, entirely inside `DBOS.shutdown()`. It was 14s only by luck: the parent was in
`synthesize-findings`, which started at 12:18:53Z and finished at 12:34:57Z. **Ctrl+C landed 14
seconds before it ended.** At 15:20 the same keystroke would have blocked for ~15 minutes; a long
sandbox step would block for as long as that step runs.

## Mechanism

The wait is unconditional and lives in the SDK:

1. `cli/src/tui/app.tsx:85-88` — `quit()`:
   ```ts
   async function quit(): Promise<void> {
       renderer.destroy();   // alt-screen exited, terminal restored
       await shutdown(0);    // ← the entire wait happens here, with no output
   }
   ```
2. `cli/src/lib/shutdown.ts:22-30` — `Promise.allSettled` over the registered hooks.
3. `cli/src/modules/harness/runtime.ts:1062-1073` — the hook calls `booted.shutdown("cli-shutdown")`.
4. `harness/src/runtime/shutdown.ts:44-57` — `runShutdownSequence`:
   `markDraining → closeHttpServer → shutdownDbos → closePool → otel → logger → exit`.
5. `harness/src/runtime/dbos.ts:129-146` — `shutdownDbos` → `DBOS.shutdown()`.
6. `@dbos-inc/dbos-sdk@4.23.6`:
[...]

Issue #203, opened 2026-07-22
Run synthesis is invisible in the TUI — the progress embed reads 5/5 done while `synthesize-findings` runs for 16 minutes


vivere-dally:
## Summary

`executeAnalysis` runs a final `synthesize-findings` step **after** the last DAG step completes. It can
run for many minutes, it emits 246 run-stream parts while it works — and the TUI shows **none of it**.
During that window the sidebar progress embed reads `5/5`, every step glyph is `done`, and nothing on
screen says the run is still working.

Observed live on 2026-07-22: the user believed the run had finished, chatted with the agent for ~16
minutes, then quit — and the CLI appeared to hang on exit because the parent workflow was still in
`synthesize-findings` (see the companion shutdown issue).

## Ground truth

Run `7e8cd842-364d-4b09-9641-e08758bbcfe3`, analysis `019f8995-dd17-7000-b51d-e4ac374a5f79`
(local prod stack, `dbos` + `public` schemas). All times UTC.

**The step ledger is complete at 12:18:53 — 16 minutes before the run ends:**

```
 step_id |  status   |        started_at        |       completed_at
---------+-----------+--------------------------+--------------------------
 T1S1    | completed | 2026-07-22T11:38:35.195Z | 2026-07-22T11:43:15.926Z
 T2S1    | completed | 2026-07-22T11:43:16.453Z | 2026-07-22T11:51:07.343Z
 T3S1    | completed | 2026-07-22T11:51:07.799Z | 2026-07-22T12:05:15.143Z
 T4S1    | completed | 2026-07-22T12:05:15.870Z | 2026-07-22T12:11:29.716Z
 T5S1    | completed | 2026-07-22T12:11:30.533Z | 2026-07-22T12:18:53.275Z
(5 rows)
```

**The run row stays `running` for another 16 minutes:**

```
 run_id                               |  status   | completed_at             | synthesis_status
 7e8cd842-364d-4b09-9641-e08758bbcfe3 | completed | 2026-07-22T12:34:57.102Z | produced
```

**The gap is `synthesize-findings`.** From `dbos.operation_outputs` for that workflow, the step is
bracketed by two `DBOS.now` steps:

| function_id | step | epoch ms | UTC |
|-|-|-|-|
| 49 | `DBOS.now` (T5S1 settled) | 1784722733717 | 12:18:53.717 |
| **52** | **`synthesize-findings`** | — | **~16m 03s** |
[...]

PR #205, opened 2026-07-22
feat(cli): narrate a slow exit instead of leaving a dead prompt

vivere-dally:
Quitting the TUI after a long session appeared to hang: the shell took ~14s to come back with nothing on screen explaining why. `App.quit` calls `renderer.destroy()` before `shutdown(0)`, so the terminal is already restored to normal scrollback while the drain runs — every second of it silent.

The drain itself is honest work, not a stall. The harness shutdown hook reaches `DBOS.shutdown()`, which awaits `awaitRunningWorkflows()` — it waits for in-flight workflows to finish and never cancels or times out. In the observed session the parent was 14s from the end of a 16-minute `synthesize-findings` step; the same keystroke a few minutes earlier would have blocked for a quarter of an hour. That unbounded wait is a separate defect (#204) and is not addressed here — this change only stops the wait from being invisible.

`shutdown()` is the right home rather than the TUI quit paths: it runs on every exit, including SIGTERM/SIGHUP and `beforeExit`, and all three of them share the symptom. Three gates keep it from corrupting anything it now sits in front of:

- stderr, and only when it is a TTY. `shutdown()` also ends piped text commands whose stdout is a machine-readable payload, so the indicator must have no route into that stream.
- A 500ms grace period before the first frame. An idle runtime drains in single-digit milliseconds, so the ordinary exit prints nothing at all, and the window also covers opentui's asynchronous terminal restore.
- The cursor is never hidden. A second SIGTERM force-exits past the stop function (`src/index.ts`), and a hidden cursor would outlive the process into the user's shell.

Past five seconds the copy stops spinning and explains, because by then the reader's question is no longer "is it working" but "is this expected or should I kill it".

Verified end to end through the real `shutdown()` under a pty with a slow hook: prior command output intact, frames animating in place on one row, the row erased on exit. Full suite 1559 pass / 0 fail; typecheck and lint clean.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/cli/src/lib/shutdown.test.ts
+++ b/cli/src/lib/shutdown.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, test } from "bun:test";
 
-import { startExitIndicator } from "./shutdown.ts";
+import { exitIndicatorFrame, startExitIndicator } from "./shutdown.ts";
 
 // The indicator's grace period is a module constant (500ms). These tests exercise the real timers
 // rather than injecting a clock: the whole contract is "stay silent unless the drain is genuinely
@@ -8,6 +8,7 @@ import { startExitIndicator } from "./shutdown.ts";
 const PAST_GRACE_MS = 640;
 
 const realIsTTY = process.stderr.isTTY;
+const realColumns = process.stderr.columns;
 const realWrite = process.stderr.write.bind(process.stderr);
 
 /** Capture stderr for the duration of one test, returning the frames written. */
@@ -29,6 +30,7 @@ function sleep(ms: number): Promise<void> {
 
 afterEach(() => {
     process.stderr.isTTY = realIsTTY;
+    process.stderr.columns = realColumns;
     process.stderr.write = realWrite;
 });
 
@@ -56,7 +58,7 @@ describe("startExitIndicator", () => {
         const drawn = frames.join("");
         expect(drawn).toContain("Exiting");
         expect(drawn).toMatch(/[⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏]/);
-        // Elapsed seconds are what tell the reader the process is alive rather than wedged.
+        // The age is what tells the reader the process is alive rather than wedged.
         expect(drawn).toMatch(/\(\d+s\)/);
     });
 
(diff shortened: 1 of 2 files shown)


vivere-dally:
@claude review

vivere-dally:
Both findings addressed in 3a92956, plus a third that fell out of verifying them.

### 1. Hand-rolled duration formatting — fixed

Correct call, and the convention citation is the right one. Swapped to `Date.relativeAge(startedAt)`.
The readability point lands exactly where it matters: the 16-minute drain that motivated this PR
printed `(960s)` and now prints `(16m00s)`.

### 2. Line wrap — fixed, and it was not low severity

Measured rather than estimated. The explain frame with a two-unit age is **80 columns exactly**:

```
⠋ Exiting — waiting for in-flight work to finish, this can take a while (16m00s)   ← 80
```

So it wrapped on the *conventional default width*, not on a narrow pane — the common case. And the
consequence is worse than stale characters: `\r\x1b[2K` erases only the row the cursor is on, so a
wrapped frame strands its first row and the next tick redraws below it. The indicator smears down the
screen for the whole drain.

Frames are now composed by taking the widest variant that fits `columns - 1`:

| columns | frame |
|-|-|
| 120 | `⠋ Exiting — waiting for in-flight work to finish, this can take a while (16m00s)` |
| 80 | `⠋ Exiting — waiting for in-flight work to finish (16m00s)` |
| 24 | `⠋ Exiting… (16m00s)` |
| 6 | `⠋` |

Two deliberate choices there:

- **A shortened explanation rung was added.** With a single message plus a fit check, an 80-column
  terminal falls straight through to the bare `Exiting…` — losing the explanation during precisely the
  long waits it exists for.
- **Degrading through variants, not truncating the composed line.** The age sits at the end, so a plain
  truncation would cut off the one part that distinguishes a live drain from a wedged one. The last
  rung is the bare glyph, which still animates.

### 3. Uncaught throw on the exit path — found while verifying, fixed

Driving the indicator through a real pty (rather than the unit tests, which run under a preload that
installs the runtime extensions) produced:

```
TypeError: Date.relativeAge is not a function
```

An artifact of the harness — the real entry point installs extensions first — but it exposed a genuine
hazard in the shape of the code. The throw happened **inside the `setInterval` callback**, where it is
uncaught: it kills the process before `shutdown()` reaches `process.exit`, replacing the real exit code
with a stack trace. `process.stderr.write` throwing `EPIPE` once the terminal goes away mid-drain — the
SIGHUP case `src/index.ts` routes into this very path — does the same thing.

Writes now go through a guard that swallows failure, mirroring the existing rule in `lib/otel.ts` that
telemetry must never break logging. Decoration on the last code path the process runs must not be able
to turn a clean exit into a crash; losing the animation is the correct degradation.

### Verification

- 13 tests on this file, each **mutation-checked**: removing the width ladder fails 4, reverting to the
  hand-rolled formatter fails 2, removing the write guard fails 1.
- Live pty at 120 / 80 / 40 columns — widest frame 76 cols, never over budget, prior command output
  intact, indicator row erased, exit 0.
- Full suite 1568 pass / 0 fail; typecheck and lint clean.

### Not addressed here

The unbounded `DBOS.shutdown()` wait this PR narrates is tracked separately in #204 — this change only
stops the wait from being invisible. #203 covers the related finding that `synthesize-findings` (the
16-minute step behind the original report) is invisible in the TUI while it runs.

radutanasa (approved):

PR #212, opened 2026-07-23
fix: Fixes 3 critical-to-intermediate severity bugs across the harness, plus 2 additional instances of the same dead-signal pattern.

extremecoder-rgb:
Summary

Bug #1: Silent teardown failure in Docker client
harness/src/sandbox/docker-client.ts
removeContainerIgnoreMissing used .catch(() => {}) on both container.stop() and container.remove(), which swallowed all errors — not just the intended 404 (container already gone). A Docker daemon disconnection, permission error, or engine failure during teardown was invisible, causing trySandbox to always report success.
Fix: Replace blanket .catch(() => {}) with targeted try/catch that only suppresses 404 responses and lets all other errors propagate for proper classification.

Bug #2: Unreachable AbortController in ephemeral runner
harness/src/execution/ephemeral-runner.ts
The ephemeral runner created new AbortController().signal but never called abort(), leaving the signal permanently green. Provider-level cooperative cancellation could not work — AI SDK retries and backoff sleeps were never interrupted on workflow cancellation.
Fix:
- Hoist AbortController to outer scope (accessible in finally)
- Pass signal to runAgent for AI SDK cooperative cancellation
- Add AbortError to isFatalLoopError so the loop stops cleanly
- Abort controller in finally block

Bug #3: Dead AbortController signals in synthesis/describer/summary runners
Files: harness/src/execution/run-synthesis.ts, artifact-metadata.ts, step-summary.ts
Same dead-signal pattern affected three more execution paths: input.signal ?? new AbortController().signal created a signal that was never aborted, disabling provider-level cooperative cancellation.
Fix: Same pattern for all three — create AbortController when no external signal provided, add AbortError to isFatalLoopError, abort in finally.**

### DIFF
--- a/harness/src/execution/artifact-metadata.ts
+++ b/harness/src/execution/artifact-metadata.ts
@@ -240,7 +240,8 @@ export async function generateFileMetadata(opts: GenerateFileMetadataOptions): P
         maxIterations: opts.maxIterations ?? DEFAULT_MAX_ITERATIONS,
     };
 
-    const signal = opts.signal ?? new AbortController().signal;
+    const abortController = opts.signal ? undefined : new AbortController();
+    const signal = abortController ? abortController.signal : opts.signal;
 
     const transcript = opts.messages ? sanitizeTranscript(opts.messages) : [];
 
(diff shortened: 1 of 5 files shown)


vivere-dally:
Hi. Thank you for your contribution!
The items 2 and 3 are intended, and applying your change would introduce regressions.
Item 1 is somewhat valid, but it's not needed for now, and there are other status codes that we should handle.
We will keep these in mind for the future.
Thanks!

PR #214, opened 2026-07-23
feat: make run synthesis visible in the TUI via a step-ledger row

### LINKED ISSUE #203: Run synthesis is invisible in the TUI — the progress embed reads 5/5 done while `synthesize-findings` runs for 16 minutes

vivere-dally:
## What & why

Closes #203.

Run-level synthesis (`synthesize-findings`) is frequently the longest phase of a run — 16m03s observed against 4–14 min per DAG step in the reported incident — but it had no row in `cortex_step_executions`. Every ledger reader therefore went blind while it worked: the TUI sidebar embed showed `5/5` with every glyph a checkmark, `inflexa run` narrated "5/5 step(s) complete" (with a stale child-step label), and `inspect_run` had no row to report. A user watching any of them concluded the run had finished while the workflow kept going, and a synthesis *failure* flipped a full bar straight to a dead run with no phase named.

This implements #203's recommended **ledger-honest** direction: give synthesis a real step-ledger row. Because the step ledger is the shared source all three surfaces already poll, one harness change lights up all of them with no new transport, read path, or UI code. A small CLI change then makes any long-running step legible.

### Harness — `synthesis-step-ledger-row`

- `executeAnalysis` seeds a reserved `synthesis` row (`agent_id = "run-synthesizer"`, `wave` after every DAG level) alongside the existing DAG seed, gated on `synthesisEnabled`, so `done/total` is honest from the first frame.
- It marks the row `running` before `synthesize-findings` and stamps its terminal status from the classified outcome, strictly **before** the run row's terminal write — so no reader ever sees a terminal run beside a `running` synthesis row.
- Outcome maps onto the existing status vocabulary: `produced → completed`, `skipped_no_summaries → skipped`, `skipped_blocker → blocked` (+ `blocked_reason`), thrown → `failed` (+ `error`). Ledger writes are log-don't-fail; a `DBOSWorkflowCancelledError` re-propagates before classification, so cancellation is never recorded as a synthesis outcome.
- `synthesis` joins the reserved step-id set, so a plan step can never collide with the row's primary key or with `runs/{runId}/synthesis.json`.
- `inspect_run` includes the row but omits a per-step `summaryPath` for it (its product is the run-level `synthesis.json`, already surfaced as `synthesisPath`) — avoiding a path to a file that never exists.
- No schema migration and no new read path: existing consumers pick the row up through `queryStepsByRun` and their existing status mappings. **Run step denominators become N+1 by design** (a 5-step plan reports a 6-row run).

### CLI — `run-step-elapsed-age`

- `RunStepView` carries the running step's `started_at`; `RunBlock` renders a muted `Date.relativeAge` beside running rows so a long step reads as *live*, not wedged — in the sidebar embed, the run-detail dialog, and the design gallery. Zero new reads: `started_at` already rides every ledger row.

### Not in scope

- The unbounded-shutdown wait this run also exhibited is **#204** (separate change; PR #205 already narrates the slow exit).
- Consuming the run-event stream for live activity labels (#203's "full-fidelity" direction) needs the OSS-side stream read helper, tracked separately.

## Type of change

- [x] Bug fix
- [x] New feature / capability (lands as a new harness capability: a reserved run-phase ledger row)
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — harness 1477 pass / 0 fail, cli 1686 pass / 0 fail.
- [x] Documentation updated if user-facing behavior changed (OpenSpec specs synced + archived; harness 62/62, cli 77/77 validate).
- [x] This PR is focused on a single logical change (make synthesis visible — the harness row makes it exist, the CLI age makes it legible).


cli/src/tui/components/run_block.tsx
```diff
@@ -258,10 +264,22 @@ export function RunBlock(props: RunBlockProps) {
...
+                        // in the row→view mapping, so both mapping sites stay dumb projections and every
+                        // rule about WHEN an age shows is in one place: running rows only, and only when
+                        // the timestamp parses. Muted TEXT tier (not the fgSubtle decoration tier) — it is
+                        // information and must clear the 4.5:1 floor. `Date.relativeAge` is the shared
+                        // elapsed-indicator vocabulary (never a hand-rolled formatter — see cli CLAUDE.md).
+                        const age = (): string | null => {
+                            if (step.state !== "running" || !step.startedAt) return null;
+                            const t = Date.parse(step.startedAt);
+                            return Number.isNaN(t) ? null : Date.relativeAge(t);
+                        };
```
vivere-dally: I think this could be an extension method on String.
We have `relativeAge` which I believe is an extension too.

But we should have date string extensions that we can use in contexts where we know that a string is a serialized date

cli/src/tui/layout/design_gallery_fixtures.ts
```diff
@@ -301,7 +303,7 @@ export const mockRun: Run = {
     total: 20,
     steps: [
         { id: "mock-step-12", label: "rank consensus", state: "done" },
-        { id: "mock-step-13", label: "build report", state: "running" },
+        { id: "mock-step-13", label: "build report", state: "running", startedAt: ago(4 * 60_000) },
```
vivere-dally: `ago` should also be an extension method on Date

harness/src/workflows/execute-analysis.ts
```diff
@@ -637,10 +708,40 @@ export async function runExecuteAnalysisBody(input: ExecuteAnalysisInput, deps:
...
             synthesisError = err;
             synthesisOutcome = { status: "failed", reason: err instanceof Error ? err.message : String(err) };
             logger.error("synthesizeFindings failed — failing the run", { runId, ...logger.errorFields(err) });
         }
+        // Terminal stamp for the `synthesis` row — deliberately BEFORE
+        // `collectAndComplete` writes the run row's terminal status, so no reader
+        // can observe a terminal run beside a still-`running` synthesis row.
+        // Log-don't-fail, like the sibling finalisation writes; the null guard is
+        // for the type only (both paths above set an outcome or threw).
+        if (synthesisOutcome !== null) {
```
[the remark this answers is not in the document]
vivere-dally: Removed. Confirmed redundant at the type level too — `tsc` passes without it (TS narrows `synthesisOutcome` to non-null after the try/catch), so the "for the type only" comment was wrong. The `const rowOutcome` capture stays to keep it non-null inside the runStep closure.

cli/src/tui/components/run_block.tsx
vivere-dally: Done — added `String.prototype.relativeAge()` (string counterpart to `Date.relativeAge`). RunBlock and the sidebar `relAge` now share it. Bonus: it lets this `components/` file reuse the logic without importing the hooks-layer `relAge`. Kept glyph-free (returns null; caller owns the em-dash fallback).

cli/src/tui/layout/design_gallery_fixtures.ts
vivere-dally: Done — added `Date.ago(ms)` beside `Date.relativeAge`/`Date.formatDuration`; fixtures use it at all call sites, local helper removed.

radutanasa (approved):

Issue #217, opened 2026-07-24
TUI: composer-carried reject feedback for docked asks (n <text>)


vivere-dally:
## Context

The `ask-answer-interactions` change (cli/openspec/changes/ask-answer-interactions/) made the chat composer a second answer path for a docked approval prompt: submitting exactly `y` / `a` / `n` (trimmed, case-insensitive) answers the head ask. Bare `n` rejects **with no feedback** — the prompt's feedback mode remains the only way to attach reject feedback.

## Ask

Let the composer carry reject feedback in one submit, e.g. `n rerun it with more threads` → reject with feedback "rerun it with more threads".

## Why it was deliberately excluded

From the change's design Non-Goals: a mistyped real message would silently become a rejection-with-feedback — the failure mode is destructive (it denies the suspended tool call and ends the turn) and invisible at submit time. Any design here needs a guard against that (confirmation, a distinct prefix, or an echo affordance) before it ships.

## Pointers

- Token parsing: `parseAskAnswer` in `cli/src/tui/app.tsx` (pure helper; the natural extension point)
- Gateway funnel: `answerAsk` in `cli/src/tui/app.tsx`
- Spec: `cli/openspec/specs/tui-ask-approval/spec.md` — the composer answer path requirement explicitly states the composer carries no reject feedback; that requirement must be amended by whatever change implements this.

PR #218, opened 2026-07-24
feat(tui): answer docked asks by clicking options or typing y/a/n in the composer

vivere-dally:
## What & why

Answering a docked approval prompt (`ctx.ask`) required focusing the prompt card first and then pressing a bare `y`/`a`/`n` key — the focus gate that makes bare printables legal also made the two most natural interactions dead ends: clicking an option label did nothing, and typing `y` into the chat composer was silently swallowed by the busy-turn submit gate. Both read as the TUI ignoring the user at the exact moment it is blocked waiting on them.

This PR adds two answer paths to the docked prompt, both funneling through the existing `answerAsk` gateway:

- **Clickable options** — each choice option (`y approve` / `a always` / `n reject`) is an individual mouse target routed through the *same handler as its key* (a click on `n` enters feedback mode), so click and key cannot drift. Activation is mouse-up only, and a release carrying a live text selection never activates — a selection drag ending on an option cannot approve a command. Options are `selectable={false}` (buttons, not prose) and design-gallery exhibits pass `inert`, which also no-ops clicks.
- **Composer answers** — while an ask is docked, submitting exactly `y`, `a`, or `n` (trimmed, case-insensitive) answers the head ask; the buffer clears when the answer settles, so a failed gateway write keeps the typed token for a retry. The existing settle/drain choreography owns every focus/mode transition (mid-turn drain lands in NORMAL on the stream pane). Bare `n` rejects without feedback. Any other text keeps the draft-preserving refusal but surfaces a transient notice naming the `y`/`a`/`n` answer path instead of failing silently (`/quit` aliases keep their precedence). An inert gallery exhibit is also non-focusable, so exhibit clicks can't yank focus off the gallery pane.

The decision logic is pure and unit-tested without booting the App (`parseAskAnswer` + `askSubmitAction`, following the `interrupt_hint.test.ts` altitude). The `tui-ask-approval` spec is synced (two added requirements, one reworded) and the OpenSpec change is archived as `2026-07-24-ask-answer-interactions`.

Composer-carried reject feedback (`n <text>`) is deliberately excluded — a mistyped message must not silently become a rejection — and tracked in #217.

**Verification:** TUI suite 456/456 green (5 new render tests — 4 mouse-interaction plus a structural `selectable`/focusability pin — and 8 new pure-helper tests); lint clean; typecheck clean for every touched file. Pre-existing, unrelated failures (missing `p-queue`, harness-contract drift in `modules/harness/turn.ts`, `modules/refs/store.ts`, `hooks/conversation.ts`) predate this branch and are untouched by it.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — clean for all touched files; the only failures are the pre-existing `p-queue`/harness-drift ones noted above, unrelated to this PR.
- [x] Documentation updated if user-facing behavior changed (OpenSpec `tui-ask-approval` spec synced + change archived).

### DIFF
--- a/cli/openspec/changes/archive/2026-07-24-ask-answer-interactions/.openspec.yaml
+++ b/cli/openspec/changes/archive/2026-07-24-ask-answer-interactions/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-24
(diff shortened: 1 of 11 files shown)


vivere-dally:
@claude review

vivere-dally:
@claude review again, latest commit

radutanasa (approved):

PR #219, opened 2026-07-24
fix(harness): guard read_file against binary content

vivere-dally:
read_file UTF-8-decoded whatever bytes it read, with no binary detection — unlike its sibling grep, which skips files with NULs. Reading a gzipped/binary file (e.g. via headLines) returned lossy garbage to the model AND injected NUL bytes into the transcript, which then failed the JSONB append in appendTurn (Postgres 22P05) and silently dropped the whole turn from thread history.

Detect a NUL in the returned window and return a `binary` data variant instead of decoding, keeping binary content out of the conversation at the source.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/harness/src/tools/workspace/read-file.test.ts
+++ b/harness/src/tools/workspace/read-file.test.ts
@@ -34,6 +34,26 @@ describe("createReadFileTool", () => {
         if (r.status === "ok") expect(r.content).toBe("hello");
     });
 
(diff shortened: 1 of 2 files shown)


vivere-dally:
@claude review

radutanasa (approved):

PR #220, opened 2026-07-24
Download a GEO Series into the analysis folder (`inflexa geo download`)

### LINKED ISSUE #216: Download GEO datasets into an analysis — host-side fetch through a staging seam, offline parse in the sandbox
## Summary

Add a first-class way to **download a GEO Series into an analysis as input data**, so an agent (or user) can say "pull GSE12345 and analyze it" and have the processed dataset land in the analysis workspace. Today `search_geo_datasets` can *find and cite* accessions but explicitly cannot fetch them — its own description carries a hard caveat (`harness/src/tools/bio/search-geo-datasets.ts:50-53`):

> "HARD CAVEAT: sandbox containers cannot download GEO data. This tool identifies and cites accessions only — never plan an analysis step that fetches a GEO dataset."

That caveat exists because the sandbox has **zero egress** — the entrypoint installs `iptables -P OUTPUT DROP` (+ ip6tables) before dropping to the workload uid (`images/sandbox-base/sandbox-entrypoint.sh:16-36`); there is no proxy and no allowlist. So no in-sandbox package (GEOparse, `Bio.Entrez`, ftplib, wget) can ever reach NCBI. The download has to happen **host-side**, and the bytes reach the sandbox through a read-only mount — the same shape every bio tool and the reference-data installer already use.

> **Design is captured as an OpenSpec change:** `harness/openspec/changes/download-geo-datasets/` (proposal, design, specs for the two new capabilities `external-input-staging` + `geo-dataset-download`, and tasks). This issue is the tracking summary; that change is the source of truth for requirements and scenarios.

## Design (decided)

- **Scope (v1):** processed + supplementary — the SOFT family file, the series matrix, and author-deposited supplementary files (processed count/expression matrices, phenotype tables). **No raw SRA reads** (that's a separate, much larger capability — see Out of scope).
- **Landing / role:** the files become **analysis input data** under `data/inputs/…` (Storage Layout in `harness/CLAUDE.md`), so the data profiler and `generate_plan` pick them up automatically. Per-analysis; matches the "download a project to analyze it" mental model.
- **Mechanism:** the host fetches the deterministic NCBI URLs (no host Python/R). The **parsing** — the part the well-established packages are actually good at — happens **inside the sandbox in offline local-file mode**: `GEOparse.get_GEO(filepath=…)` / `GEOquery::getGEO(filename=…)`, the one retrieval path in those packages that works with no egress.
- **Layer (harness-first):** the harness **declares the GEO tool and a new "materialize external dataset as input" seam**; an embedder realizes the egress + storage. The OSS/CLI realization reuses the existing host-side downloader (`cli/src/modules/refs/store.ts`) and input staging (`cli/src/modules/staging/staging.ts`); a managed/cloud host can realize the same seam against its own fetch service + object storage. This keeps a host-agnostic product capability in the harness, per the root `CLAUDE.md` design rule ("Design new capabilities harness-first, then wire them from the embedder").

### Flow

```
conversation agent
  └─ download_geo_dataset(accession="GSE12345")           [harness tool, host process]
       1. validate accession → resolve NCBI URL templates (soft / matrix / suppl)
       2. size estimate (HEAD probe)  ── egress ──┐
       3. ctx.ask approval (show estimate)        │  behind the staging seam
       4. ExternalInputStager.stage(...)  ── egress + write ──┘   → data/inputs/geo/GSE12345/…
       5. register staged files as analysis inputs (+ optionally refresh data profile)
       6. return { accession, files:[{path, bytes}], receipt }
                                   │
   sandbox step (no egress) ───────┘  reads /{resourceId}/data/inputs/geo/GSE12345/… read-only
       GEOparse.get_GEO(filepath="…/GSE12345_family.soft.gz")   # offline parse
```

## The staging seam (new harness capability)

A new capability seam declared by the harness — the **egress + storage boundary** for bringing a described external dataset in as analysis input. Illustrative shape (final names/granularity TBD in implementation):

```ts
interface ExternalInputStager {
  stage(req: StageExternalInputRequest, session: AgentSession):
    ResultAsync<StagedInput[], StagingError>;
}
interface StageExternalInputRequest {
  resourceId: string;                 // the analysis
  datasetLabel: string;               // e.g. "GSE12345"
  artifacts: ExternalArtifact[];      // { sourceUrl (https), destRelPath, format, contents }
  maxTotalBytes?: number;             // size cap
}
interface StagedInput { relativePath: string; bytes: number; sha256: string; }
```

- **OSS/local realization** streams each HTTPS URL into the workspace `data/inputs/…` tree under the embedder-resolved root (`ResolveWorkspaceRoot`, `harness/src/workspace/paths.ts`), with the guardrails the ref downloader already implements in `cli/src/modules/refs/store.ts:downloadArtifact` — HTTPS-only re-checked on the post-redirect `response.url`, sha256, `.part`→atomic activate, progress events. The reusable transfer primitive should be **factored so the ref installer and this seam share it** rather than duplicating it.
- **CLI wiring** reuses `store.ts` (transfer) + `staging.ts` (`StagedInput`, `inputs/{mountName}/{key}` layout).
- Everything that touches the network lives behind this seam — including **supplementary-file enumeration** (a `suppl/` directory listing is itself an egress call), so the seam owns discovery *and* transfer of the concrete file set.

## The GEO tool (harness, conversation agent)

A `defineTool` on the conversation agent (`harness/src/agents/conversation-agent.ts:203-290`), constructed via a factory that captures the seam dep. It owns the **GEO domain knowledge** only:

- **Input:** `accession` (`GSE…`), optional `include: ("metadata"|"matrix"|"supplementary")[]` (default all three), optional `maxSizeMb`.
- **Resolve:** accession → deterministic URL bucket (`GSE12345` → `GSE12nnn/GSE12345/`) and per-class templates:
  - `…/soft/GSE12345_family.soft.gz`
  - `…/matrix/GSE12345_series_matrix.txt.gz` (per-platform `GSE12345-GPLxxx_series_matrix.txt.gz` when multi-platform)
  - `…/suppl/` — enumerate and fetch author-deposited supplementary files (or the bundled `?acc=GSE12345&format=file` RAW.tar)
- **Guardrails:** HEAD-probe size estimate → `ctx.ask` approval showing the estimate (downloads write disk + hit the network + can be large; mirror the ref installer's `measureReferenceDownload` and the approval-gating precedent of `run_inflexa`, `cli/src/modules/harness/inflexa_tool.ts:376-393` / `cli/src/cli/index.ts:527-539`) → enforce the size cap → progress via `ctx.emit`.
- **Provenance:** write a receipt (accession, files, bytes, sha256, source URLs, fetchedAt) beside the staged data, mirroring `harness/src/reference-data/receipt.ts`.
- **After staging:** register the files as analysis inputs so `inspect_data_profile` / `generate_plan` see them.

## Offline parse in the sandbox

The "well-established packages" run **in the sandbox, on the already-staged local files, offline** — no network. Both support it:
- `GEOparse.get_GEO(filepath="…/GSE12345_family.soft.gz")`
- `GEOquery::getGEO(filename="…/GSE12345_series_matrix.txt.gz")`

These must be present in the sandbox library store (`/mnt/libs`) — confirm/stage as part of this work.

## Follow-on edits

- **Revise the search tool caveat** — `search_geo_datasets`'s "never plan an analysis step that fetches a GEO dataset" (`harness/src/tools/bio/search-geo-datasets.ts:50-53`) becomes "to bring a dataset in, use the GEO download tool," so search → download is a coherent path.
- **A thin skill note (declares *what*, never *how*).** Per the root `CLAUDE.md` rule, a skill says "the GEO Series arrives pre-staged as input data; parse the local SOFT/matrix with the offline reader" — it must **not** name the tool, a path, or a format. The externally-authored `geo-database` skill assumes full egress on every path and is unusable in-sandbox as written; this is the inflexa-native replacement note.

## Open questions (settle during implementation)

- **Seam granularity:** does the seam receive a concrete artifact list, or the accession + artifact-class selection (owning GEO-native resolution)? Keeping it GEO-agnostic argues for the harness resolving templates and the seam handling enumeration + transfer of a described set — but `suppl/` enumeration needs egress, which pins it behind the seam.
- **Where the shared transfer primitive lands** once factored out of `cli/src/modules/refs/store.ts` (a harness lib the local seam realization uses, vs. staying CLI-side and injected).
- **Multi-platform series matrices** (`GSE…-GPLxxx_…`) — fetch all, or let `include`/the agent choose.
- **Supplementary strategy:** enumerate `suppl/` and fetch selectively, vs. the bundled `RAW.tar` endpoint (simpler, but coarser and can include unwanted raw payloads).
- **Auto-profile:** trigger `runDataProfile` on completion, or just register inputs and let the normal flow pick them up.
- **Size-cap default** and behavior when exceeded (hard fail vs. ask-to-proceed).

## Out of scope (v1)

- **Raw SRA sequencing reads** (FASTQ via SRA / `download_sra=True`) — GB–TB scale, needs sra-tools + an nf-core-style pipeline and a separate SRA egress path. Track separately.
- **GSM / GDS / GPL** single-sample / curated-dataset / platform accessions — GSE Series is "the project" here; the others are follow-ups.
- **Cross-analysis, accession-keyed caching / dedup** — GEO accessions are immutable, so a shared cache hardlinked into inputs is a clear later optimization, not v1.

## Anchor files

- `harness/openspec/changes/download-geo-datasets/` — the OpenSpec change (proposal / design / specs / tasks)

- `harness/src/tools/bio/search-geo-datasets.ts` — existing search tool + the no-download caveat to revise
- `images/sandbox-base/sandbox-entrypoint.sh:16-36` — the egress firewall (why host-side)
- `harness/src/sandbox/mount-plan.ts` — read-only `/{resourceId}` + `/mnt/refs` mounts
- `cli/src/modules/refs/store.ts` — the host-side download primitive to reuse/factor (`downloadArtifact`, `measureReferenceDownload`)
- `cli/src/modules/staging/staging.ts` — input staging (`StagedInput`, `inputs/{mountName}/{key}`)
- `harness/src/reference-data/catalog.ts` + `receipt.ts` — provenance/receipt pattern
- `harness/src/tools/sandbox/list-available-refs.ts` — discovery precedent (find-by-what-it-is)
- `harness/src/agents/conversation-agent.ts:203-290` (registration) + `:167-173` (`hostTools` seam)
- `cli/src/modules/harness/inflexa_tool.ts` / `cli/src/cli/index.ts:527-539` — approval-gating precedent
- `harness/src/workspace/paths.ts` — `ResolveWorkspaceRoot`

radutanasa:
Implements the OpenSpec change **`download-geo-datasets`** — `inflexa geo download <GSE>` downloads a GEO Series' processed data host-side into the analysis's folder. Closes #216.

> **This PR's scope changed materially after review.** It previously enrolled the downloaded files as analysis inputs and drove staging/re-profiling through a post-action reconcile hook. That is gone: the command downloads and stops. The reasoning is under [Why enrolment was dropped](#why-enrolment-was-dropped) — it is the most important thing to read here.

## What the command does

> **Renamed since review: `geo add` → `geo download`.** The command no longer adds anything — `inputs add` is what adds — so `add` promised the one thing it does not do. `src/modules/geo/add.ts` moved to `download.ts` and `runGeoAdd` to `runGeoDownload`.

`inflexa geo download <gse>` resolves a Series accession to its NCBI file set — SOFT family file, series matrix (per-platform parts when the Series spans platforms), author-deposited supplementary files — and fetches them over HTTPS into `<anchor>/<accession>/`, the folder the user ran `inflexa` in. Raw SRA reads are out of scope.

It writes files and touches nothing else: no input rows, no provenance, no staging, no profiling, no harness runtime. Registered `approval` per `agent-command-policy`, so the conversation agent reaches it through the existing `run_inflexa` tool behind an approval prompt. No new agent tool, no TUI dialog.

Making the files inputs is a separate, explicit user action through the path that already exists — `manage_inputs` in chat, the palette's `AddInputDialog`, or `inflexa inputs add` — all of which already stage and profile what they enrol.

## Why enrolment was dropped

Auto-enrolment bought nothing a user cannot do in one sentence, and cost a great deal:

- **It broke the analysis lock.** Enrolment emits provenance, and every other mutating surface claims the analysis instance lock first (`inputs add/remove`, `chat`, `profile`, `run`). A subprocess spawned by `run_inflexa` beside a live TUI cannot hold that lock — which is exactly why `inputs add` is classified **`blocked`** for the agent, with the reason *"running this as a subprocess would be refused by the analysis lock the chat holds."* An auto-enrolling download was the same mutation through a door that skipped the guard. Reproduced against real sqlite: after host → child → host, the child's input record was gone from the signed document and `verifyAnalysisIntegrity` returned `tampered`.
- **It needed a reconcile channel back to the host.** The child emits `prov.input_added` on its own bus, which the host never sees, so the host had to be poked out of band to re-stage and re-profile. That hook fired after every successful analysis-scoped action, including read-only `auto` ones.
- **It coupled two independent intents.** "Fetch this dataset" and "make it an input" are separate decisions; a user may want the files without profiling them.

Dropping it deletes all of it: no lock to claim, no provenance to fork, no reconcile channel, no staging trigger. The reconcile wiring (`onReconcile`, the `reconcile` boot parameter, `reconcileOpenAnalysisProfile`/`bindOpenWorkspace`) is reverted across five files.

## The bug that motivated the rewrite

The original implementation **could not complete a single real accession**, while 15 tests passed green.

`parseAutoindex` filtered hrefs by pattern. Every NCBI autoindex page ends with a site-wide footer link:

```html
<a href="https://www.hhs.gov/vulnerability-disclosure-policy/index.html">HHS Vulnerability Disclosure</a>
```

It starts with none of the filtered prefixes and ends with no slash, so it survived as a *file name*. Against the real page the shipped function returned:

```
["GSE185553_series_matrix.txt.gz", "https://www.hhs.gov/vulnerability-disclosure-policy/index.html"]
```

The fabricated URL 404s, and `downloadGeoSeries` aborts the whole set on the first failed transfer. `soft/` is enumerated first, so the run died before ever reaching `suppl/` — which is where the count matrices live for most modern RNA-seq Series. That is the reported "supp files are not downloaded": not a supplementary-specific bug, but nothing after the first directory could download at all.

Two faults compounded it. NCBI answers **403** — not 404 — for a directory with no index, which is the normal state of `suppl/` for a Series with no supplementary files; `listDirectory` treated any non-404 as fatal. And NCBI sheds load with that same 403, observed flapping 200↔403 on identical URLs seconds apart, with no retry anywhere.

**The fix:** resolve each href against the directory URL and keep it only when same-origin and exactly one segment deeper. That single test subsumes every case the negative filter got wrong — sort links carry a query, both parent-link spellings resolve above the directory, subdirectory links leave a trailing segment, `mailto:` has no matching origin, the footer is off-origin — and disposes of path traversal for free, since `a/../../etc/passwd` normalizes during resolution and lands outside the directory. Names are then HTML-unescaped, percent-decoded, and admitted only as single safe segments; the decode order is load-bearing, because `%2e%2e%2f` survives URL parsing intact and becomes `../` only afterward.

## Also in this PR

**Retry and pacing** — 403/429/5xx and transport throws retry with exponential backoff, spaced between requests. A persistent 403 settles as "empty directory" rather than failing the resolve.

**Staged activation** — transfers land beside the destination and move in only on full success, so a failed run leaves nothing in the user's data folder.

**Size bound with an override** — a HEAD sweep estimates the set before transferring; `GEO_SERIES_MAX_BYTES` (32 GiB) refuses runaways and `--max-size 500MB|64gb|<bytes>` raises it. Binary units, so a size copied out of the command's own output round-trips.

**`run_inflexa` is bounded by silence, not duration** (`DEFAULT_IDLE_TIMEOUT_MS`). A flat wall-clock deadline cannot tell a working command from a wedged one, so any value is wrong for someone: 120s killed a legitimate multi-gigabyte download, and a ceiling generous enough to spare it would let a hung command hold the turn just as long. The bound is now the gap between output — 120s, rearmed on every chunk either stream produces — plus a 30-minute absolute backstop. Activity counts *before* the output cap is applied: past the cap the bytes are dropped, but they still prove the child is alive. `geo download` feeds it a progress heartbeat every 5s, which is the readout a long download needed anyway.

**Subcommand flags were silently swallowed** (pre-existing, since `864d25a`). The root program declares `--analysis`/`--project`, and commander lets a program option appear anywhere on the line — so `inflexa profile --analysis x` binds the value to the *root* and hands the subcommand's action an empty options object. Six commands took the flag and ignored it. `registerAction` now hands handlers `optsWithGlobals()`, fixing every command at one choke point. Deliberately **not** `enablePositionalOptions()`: that breaks `inflexa sessions --project x` into a hard "unknown option", was tried and reverted once before, and its regression test is what caught the second attempt — both halves are now pinned side by side.

**`search_geo_datasets`** no longer tells the agent that fetching is impossible. It still forbids planning an in-sandbox download (the sandbox genuinely has no egress) but defers whether the data can be obtained to the agent's actual tool set. It names no host command, since the harness is host-agnostic.

**`--project` is gone from the command** — `resolveContext` answers a project ref with a picker, never a single analysis, so it could only ever have failed.

## Verification

Live, against the accessions that were failing:

```
GSE185553: 5 artifacts, 153.1 MB  → soft, series_matrix, barcodes.tar.gz, RAW.tar, filelist.txt
GSE110004: 5 artifacts, 2.4 GB    → soft, GPL17342 + GPL21656 matrices, RAW.tar, filelist.txt
```

Supplementary files resolve, and `GSE110004` confirms multi-platform enumeration. A full transfer of `GSE185553` landed all 5 files with heartbeats every ~5s through the 152.8 MB artifact and the staging directory cleaned up; an over-cap refusal and a `--max-size`-raised run were both exercised against the live endpoint.

Tests go from 15 to 35 in the GEO module, over **captured real NCBI HTML** — the populated autoindex with its absolute parent link and hhs.gov footer, and the real "Access forbidden" page. The stub renderer emits that same shape, so no fixture is kinder than production; the idealized fixtures are precisely why the broken version shipped green.

`typecheck`, `lint`, `prettier`, `docs:gen`, and `openspec validate --strict` clean.

**On the suite:** 1738 pass / 38 fail. I diffed the failure sets rather than counting — 37 are pre-existing (identical set with these changes stashed; each file passes in isolation), and the 38th is this PR's own idle-rearm test joining that same cluster: six `spawnInflexa` tests that fail only in full-suite runs because spawned children yield no captured output under 124-file parallel load. It fails at exactly the idle bound with empty stdout, the same signature as its neighbours. Tracked as gap 6.2; it predates this branch.

## Targeting the chat's analysis

`run_inflexa` spawns the subprocess **in the session analysis's folder**. That is what makes a bare chat request — "download geo dataset GSE12345" — land in the chat analysis's folder rather than in whatever directory the host process was started in; the two diverge after a resume, an `--analysis` launch, or a mid-session swap. The child then behaves as if the user had `cd`'d there, so the ordinary marker walk-up does the work.

An earlier revision of this PR threaded the analysis **id** instead, as `INFLEXA_ANALYSIS`, with a new ambient precedence tier in `resolveContext` to honour it. Review pushed back on it (thanks @vivere-dally), and the pushback was right: the command only ever wanted a folder — `resolveTargetFolder` never read the analysis, both branches returned `ctx.anchorPath` — so that machinery resolved an id, hit the database, and resolved its anchor row to obtain something it could have been handed.

Setting the working directory is smaller *and* more general: the ambient tier only helped commands that opted into it (one, in practice), while a working directory fixes every command the agent runs. The trust property is unchanged and arguably stronger — the host picks the *working directory* from the session scope, and nothing in the model's argv can move it, so a bare agent-composed command lands on the chat's analysis by construction. To be precise about what that does and does not cover: it is a guarantee about the working directory, not a containment boundary. `geo download` declares `--analysis`, so an argv carrying it retargets the folder just as it would for a human at a shell; the approval prompt, which shows the exact argv, is the boundary there — the same one every other approval-gated command has. Deleted along the way: `INFLEXA_ANALYSIS` and `ambientAnalysisRef()`, `ContextFlags.ambientAnalysis` and the resolver tier, their tests, and the whole `context-resolution` spec delta — that capability is now consumed unchanged.

Accepted trade-off: a relative path in an agent-run command now resolves against the analysis's folder rather than the launch folder. Known limitation: a working directory cannot disambiguate *which* analysis when one folder holds several — only an id could. Nothing needs that today, and this command never will.

## The refs deduplication (task 1.2)

`refs/store.ts` carried a *second* copy of the primitive `lib/download.ts` was extracted from — a byte-identical `declaredContentLength`, and a `downloadArtifact` reimplementing the same stale-part discard, https-on-redirect check, byte counter, `pipeline`, `stat`, and `sha256File`. The extraction had one caller and deduped nothing, which also left `lib/download.ts` in breach of the rule that a new file needs multiple callers.

`downloadArtifact` is now a wrapper keeping only what is genuinely the installer's:

- **`assertOwnedPath`** — the generic utility has no notion of an installer-owned tree, and this is the guard that stops a symlinked ancestor redirecting a write outside the store.
- **Dataset/path attribution** — the generic `DownloadProgress` cannot carry it, so its three events remap onto `artifact_started`/`_bytes`/`_completed`, preserving both invariants the originals documented (started only after the response is accepted; completed only after hashing succeeded).

Two things fell out of doing it properly rather than mechanically:

`lib/download.ts` now reports an unreachable host as `http_failed` rather than `io_failed`. That fixes a misclassification against its own doc (`io_failed` is described as a local fs fault) *and* makes the installer's remap faithful — every wire fault becomes `download_failed`, only a local fs fault stays `io_failed` — so all three existing error-type assertions (a fetch throw, a 404, a non-https redirect) hold unchanged.

The download cache entry is now dropped as soon as it is staged rather than after activation: it is a second full copy of the artifact, nothing resumes from it, and holding every one until the end doubled the peak disk a multi-gigabyte dataset needs. The install test asserted only that no `*.part` remained — which a leaked cache entry passes — so it now asserts the whole cache directory is empty, and I verified it fails without the cleanup.

94 refs tests green; the full-suite failure set is byte-identical to before.

## Deferred (tracked in `tasks.md`)

- **4.3** — an end-to-end integration test (download → add via the existing inputs path → profile → read the series matrix). Blocked on there being no headless way to create an analysis; `inflexa new` is a TUI launcher.

---

### Follow-up: automated review + reviewer thread (commit `1437d92`)

The bot's three findings, plus two defects it missed that an independent pass turned up by re-running the code:

- **`geo download` had no tests.** Now has 13 — nine subprocess cases over every arm of target-folder resolution (empty / anchor / analysis / copy / unmatched ref / desynced marker) plus the two argument gates, and four over the error messages. `describeGeoDownloadError` is exported for those, since the command threads no fetch seam and no offline run can reach them. The `too_large` case pins that the `--max-size` it suggests actually admits the Series — swapping its `Math.ceil` for `Math.floor` fails the test.
- **The p-queue ask.** Answered on [the thread](https://github.com/inflexa-ai/inflexa/pull/220#discussion_r3659926606) and adopted where it belongs: the size sweep now runs behind a queue of 4 under one 15s deadline. Declined on the directory listings and on the transfers, for reasons recorded as D12.
- **Overstated prose.** "The model cannot name a different analysis because no channel exists to name one through" was wrong: the *working directory* can't be retargeted by argv, but `--analysis` still retargets the folder. Corrected here, in the spec delta, in design.md, and in the `inflexa_tool` comment. D10 also described a handler-argument rewrite the `preAction` hook had replaced.

Missed by the bot, both demonstrated by running the code:

- **A shed transfer discarded the whole Series.** The artifact GET was the one GEO request never routed through `fetchGeo` — `downloadToFile` fetched once and treated any non-ok status as terminal — so one transient 403/503 aborted the set and the staging teardown threw away everything already transferred. It also left this change's own requirement ("the command lists **or fetches** … retries with backoff") half-implemented. `downloadToFile` now takes an opt-in retry schedule, defaulted off so refs is behaviourally untouched.
- **The size sweep could get the child killed.** Unbudgeted and silent — one HEAD per artifact, each free to spend ~3.5s retrying a shed 403, no progress event until it finished. Measured at ~3.6s per throttled artifact, ~34 artifacts crossed the 120s idle bound before a byte moved. The 15s sweep deadline makes that structurally impossible; the test times out at 5s without it.

Also fixed along the way: `runCli` captured child output through pipes, which empties out under full-suite load — the child ran and exited correctly, but `stdout`/`stderr` arrived as `""`, so assertions about what a command *printed* failed while its exit-code assertion passed. It now redirects to files. That repairs three pre-existing failures (`project new` ×2, `repair` ×1).

`bun test`: 1771 pass / 35 fail, the failure set a strict subset of the pre-existing one captured with this branch stashed — no new failures, three fixed.

---

### Follow-up: second review pass (commits `1610981`…`3081b91`)

Eight findings from a review of the branch as it stood. Each fix has a test that was confirmed to fail against the unfixed source — the source changes were stashed and the suite re-run, and all nine behavioural tests went red.

Three were defects the code contradicted itself on:

- **The idle deadline left a timer nothing owned.** `spawnInflexa` clears its timers once the child is reaped, but keeps reading the pipes for a further flush grace — so a chunk arriving in that window rearmed the idle timer *past* its own clears. It then held the event loop open for the full 120s and aborted a run that had already returned. Demonstrated by instrumenting `setTimeout`/`clearTimeout` around a spawn whose grandchild inherits the pipes and writes after the child exits: one timer armed, none cleared. The function's own comment already argues at length that a pending timer must never outlive the spawn; a `settled` latch makes that true.
- **A killed download littered the user's data folder.** Staging is a sibling of `destDir` removed in a `finally`, which does not run on a signal — and a signal is this command's likeliest abort, since `run_inflexa` escalates SIGTERM→SIGKILL on its deadline and ctrl-c does the same. Every killed run left a partial `.incoming-…` directory behind, cumulatively, with nothing anywhere to reclaim it. "An aborted run leaves nothing in the user's data folder" held only for a graceful in-process error. The next run now sweeps them.
- **A re-download merged instead of replacing.** Files moved into `destDir` one at a time, so a Series that had since dropped an artifact kept the stale copy locally and still reported complete — against this file's own "either the complete published Series or does not exist at all". The staging directory is now swapped in whole. `rename` refuses a non-empty destination, so an existing copy steps aside to `.replaced-…` and is dropped only once the new set is in place; that leaves one window where `destDir` does not exist and `.replaced-…` is the user's *only* copy, so the sweep restores it rather than discarding it.

Two were silent-omission paths:

- **`parseAutoindex` dropped names it could not reproduce** — a leading dash, an embedded separator, an unusable percent-escape — while the closing line still claimed the Series was downloaded. Reported on the progress channel and warned to stderr now, so the one case where "complete" overstates what landed is the one case the user hears about. Page furniture stays unreported: a sort link was never a file.
- **Discarded responses were never drained.** A response owns its socket until the body is cancelled. `downloadToFile`'s retry path knew this; its two refusal arms (bad status, non-https redirect) did not, and neither did `fetchGeo`, which abandons up to three shed listings per directory.

Three were conventions:

- **`registerAction` had no JSDoc.** Two blocks had ended up stacked back to back, so the exported function documented itself to nothing and lost its hover text — the entire reason `cli/CLAUDE.md` mandates JSDoc over `//`. The ancestor-hydration prose moved to the function it actually describes.
- **Hydration copied every ancestor option, not the ones a command declares.** `optsWithGlobals()` carries the root's whole set, so `geo download` — which declines `--project` deliberately, with a comment saying why — was handed one anyway. Inert (handlers read only what they declare), but the plumbing meant to honour a declared flag was quietly undoing a deliberate omission.
- **`runCli` could leak a descriptor** when the second `openSync` failed after the first succeeded.

**On the new `spawnInflexa` test.** Its first shape used the grandchild mechanism and so joined gap 6.2 — it failed every full-suite run, because under 125-file parallel load a spawned child's pipes yield nothing and all six of its neighbours fail together for that reason. Rather than add a seventh permanent red, the timer assertion is now unconditional (an empty set is the correct answer either way) and the precondition is asserted separately with a comment stating plainly that a green full-suite run is **not** coverage here — the file has to be read in isolation, where its cluster is meaningful, for the guard to be genuinely exercised.

Spec delta gains four scenarios for the new guarantees: replace-not-merge, previous copy preserved on a failed re-download, staging reclaimed after an interrupted run, and an unusable published name reported.

`typecheck`, `lint`, `prettier --check`, `docs:gen`, `openspec validate --strict` all clean.

`bun test`: **1785 pass / 35 fail**, up from 1771/35. `diff` against the failure set captured with these changes stashed is empty — the same 35, no new failures.


cli/src/lib/download.ts
vivere-dally, line 1: We already have some downloading mechanism for `refs`.
Can you re-use that, and if you're downloading many files, maybe use `p-queue` with a limit of 4.

cli/src/modules/analysis/context.ts
```diff
@@ -19,7 +19,17 @@ export type ResolvedContext =
...
+export type ContextFlags = {
+    analysis?: IdOrName;
+    project?: IdOrName;
+    /**
+     * The ambient analysis ref (the `INFLEXA_ANALYSIS` value, read at the CLI boundary via `lib/env`).
+     * Ranked below an explicit `analysis`/`project` — a command may always override it — and above the
+     * `.inflexa` marker walk-up, so an agent-run command targets the chat's analysis regardless of the
+     * subprocess's working directory.
+     */
+    ambientAnalysis?: IdOrName;
```
vivere-dally: What is this supposed to do? Why do we have it

cli/src/modules/harness/runtime.ts
```diff
@@ -435,14 +435,21 @@ export function __resetHarnessRuntimeForTest(): void {
  * {@link bootHarnessRuntimeOnce}.
  */
 export function bootHarnessRuntime(
-    options: { seams?: Partial<BootSeams>; config?: ResolvedHarnessConfig; connection?: ResolvedModelConnection } = {},
+    options: {
+        seams?: Partial<BootSeams>;
+        config?: ResolvedHarnessConfig;
+        connection?: ResolvedModelConnection;
+        /** Post-action reconcile hook wired into `run_inflexa`; the host supplies it (see the TUI's reconcileOpenAnalysisProfile). A live callback, not persisted config. */
+        reconcile?: (analysisId: string) => void;
```
vivere-dally: Seems complicated, why do we have this?

cli/src/tui/hooks/boot.ts
```diff
@@ -62,13 +62,19 @@ export type BootDriver = typeof bootHarnessRuntime;
  * `ready` is a no-op, so the launcher may fire it once post-render (fire-and-forget) without
  * guarding against a double-open. `driver` is injected only by tests.
  */
-export async function startHarnessBoot(config: ResolvedHarnessConfig, driver: BootDriver = bootHarnessRuntime): Promise<void> {
+export async function startHarnessBoot(
+    config: ResolvedHarnessConfig,
+    driver: BootDriver = bootHarnessRuntime,
+    reconcile?: (analysisId: string) => void,
```
vivere-dally: Same, why?

vivere-dally (changes requested):
Please follow the details and implementation of PR #226

cli/src/modules/analysis/context.ts
radutanasa: Removed — you were right to push on it.

The command only ever wanted a **folder**: `resolveTargetFolder` never read the analysis, both of its branches returned `ctx.anchorPath`. So this tier resolved an id, hit the database, and resolved its anchor row to obtain something the caller could have simply been handed.

`run_inflexa` now spawns the subprocess *in the session analysis's folder*, and the ordinary `.inflexa` marker walk-up does the rest. That deletes `INFLEXA_ANALYSIS`, `ambientAnalysisRef()`, `ContextFlags.ambientAnalysis`, this tier, their tests, and the whole `context-resolution` spec delta — that capability is consumed unchanged now. It's also strictly more general: the tier only helped commands that opted into it (one, in practice), whereas a working directory fixes every command the agent runs. Trust property is unchanged and arguably stronger — the host picks the folder from the session scope, so there's no longer any channel through which the model could name a different analysis.

Done in 32b9617.

One thing worth flagging that wasn't visible from this hunk: the *"a command may always override it"* claim in this very comment was **false** when you read it. The root program declares `--analysis`, and commander binds the value there rather than on the subcommand — so the flag reached no handler at all, on six commands, since 864d25a. Fixed separately in 50e3aa8 (`registerAction` now hands handlers `optsWithGlobals()`), with both halves of that trade-off pinned in `cli.test.ts`.

cli/src/cli/agent_policy.ts
```diff
@@ -59,12 +59,30 @@ export function getAgentPolicy(command: Command): AgentPolicy | undefined {
...
     setAgentPolicy(command, policy);
-    return command.action(handler);
+    // Commander invokes an action as (...declaredArguments, options, command), always passing both
+    // trailing values — so the options slot is at `length - 2` whatever the command's arity.
+    return command.action((...args: unknown[]) => {
+        const self = args[args.length - 1] as Command;
+        const merged = [...args];
+        merged[args.length - 2] = self.optsWithGlobals();
+        return handler(...(merged as unknown as Args));
+    });
```
vivere-dally: Explain this

radutanasa: **The problem it solves.** The root program declares `--analysis`/`--project` for the bare-`inflexa` flow. Commander binds a program option to the command that *declared* it, wherever it appears on the line — so for `inflexa profile --analysis x` the value lands on the **root**, the subcommand's identically-named option never receives it, and its handler is called with `{}`. The flag reaches no code and nothing reports that it was ignored. Six commands had been taking `--analysis` and silently dropping it since 864d25a.

Proof, with the root declaration toggled and everything else held fixed:

```
rootOpts=true   ->  subcommand action options = {}
rootOpts=false  ->  subcommand action options = {"analysis":"foo"}
```

and where the value actually goes:

```
subcmd action options : {}
cli.opts()  <-- HERE  : {"analysis":"foo"}
```

`registerAction` is the only sanctioned way to attach an action, so fixing it there fixes every command at one choke point.

**You were right that the implementation was opaque**, and I've replaced it in ba008c9. It used to rewrite the handler's arguments positionally:

```ts
const self = args[args.length - 1] as Command;
merged[args.length - 2] = self.optsWithGlobals();
return handler(...(merged as unknown as Args));
```

That asks a reader to take three things on trust at once: that commander always passes `(...declaredArguments, options, command)`, that the options slot is therefore at `length - 2` whatever the arity, and that the double cast back to `Args` is sound.

None of it was necessary. Commander builds those arguments from `command.opts()` at *invocation* time, which is after `preAction` hooks run — so seeding the values in a hook means the handler receives its ordinary, correctly-typed options object and nothing downstream knows this happened:

```ts
command.hook("preAction", (_parent, actionCommand) => hydrateFromAncestors(actionCommand));
return command.action(handler);
```

`hydrateFromAncestors` fills only keys the command has no value for, so an explicit flag on the subcommand always wins, and it's idempotent — which matters because the root registers the hook too and ancestor hooks fire for a subcommand's action as well.

Verified against the real registry shape (root, nested subcommand, variadic argument, and a subcommand declaring no options of its own): the flag reaches every handler, `inflexa sessions --project x` still parses, and bare `inflexa` still gets the root's own options.

**Why not `enablePositionalOptions()`** — the obvious fix — is worth recording, because it was tried and reverted once before this PR: it turns a root-style flag placed *after* a subcommand into a hard "unknown option" error, breaking `inflexa sessions --project x`. The regression test guarding that shape is what caught my second attempt at it. Both halves are now pinned side by side in `cli.test.ts`.

vivere-dally:
@claude review

cli/src/lib/download.ts
radutanasa: Both halves, separately.

**Re-use: done.** `downloadToFile` + `declaredContentLength` are now `src/lib/download.ts`, and `refs/store.ts`'s `downloadArtifact` is repointed at them (design D11) — one primitive, two production callers. refs keeps only what is genuinely its own: `assertOwnedPath` (the generic utility has no notion of an installer-owned tree), its dataset-attributed progress events, and dropping the cache entry at staging rather than at activation. The error unions reconcile by classifying an unreachable host as `http_failed`, so `io_failed` stays what it is documented to be — a local fs fault.

**p-queue: yes on one of the three loops, no on the other two.** They aren't one decision — the difference is what a shed request costs, and NCBI answers 403 both for load-shedding and for a directory with no index (D7 records flapping 200↔403 on identical URLs seconds apart).

*Adopted on the size sweep* (`measureGeoArtifacts`). You were right, and it was worse than "a bit slow": that sweep was unbudgeted and completely silent — `downloadGeoSeries` emitted `resolved` only after it drained — and it sits in front of a subprocess whose bound IS silence (`DEFAULT_IDLE_TIMEOUT_MS = 120_000`). GSE167118's `suppl/` lists 128 files, a HEAD to ftp.ncbi.nlm.nih.gov measures ~0.35s, and every probe goes through `fetchGeo`, which retries a 403 four times over ~3.5s. So a throttled sweep could burn minutes and get the child killed, for a step whose worst honest outcome is "size unknown". `measureReferenceDownload` already solves exactly this against the same class of hosts: PQueue plus one `SIZE_PROBE_BUDGET_MS` `AbortSignal.timeout` shared by every probe. The budget matters more than the concurrency — it's what makes a metadata step structurally incapable of causing that kill. Mirrored at concurrency 4 with a 15s sweep deadline, and `fetchGeo` now stops retrying once the caller's signal has aborted, so the deadline bounds the drain instead of each queued probe spending its backoff on a dead signal. The test fails (5s timeout) without the budget.

*Declined on the directory listings* (`resolveGeoArtifacts`). Three requests; concurrency saves at most 300ms of spacing, and the downside is the worst in the file. `listDirectory` maps a settled 403 to `ok([])` because a 403 is also how NCBI says "no index here". Firing all three at once raises the odds of provoking a shed; if it lands on `suppl/`, the command downloads soft+matrix, exits 0, and tells the user it has the Series. A silently incomplete Series is not worth 300ms — and that is the bug this PR exists to fix.

*Declined on the transfers* (`downloadGeoSeries`) — for a specific reason, not an aesthetic one. That path has the least room to absorb a shed response: one failure aborts the set and the staging teardown discards every byte already transferred, so parallelism would multiply exposure to precisely the status GEO sheds with, on the path that can least afford it. Three more: p-queue can't cancel in flight — `store.ts` documents that cleared tasks never settle, which is why it uses a skip-on-entry flag — so a decided failure would leave up to three more transfers, potentially multi-GB, running while staging is torn down, where the serial loop stops at once; the readout is a single-threaded transcript by construction (`file_progress` prints bare indented bytes with no filename), and that transcript is what rearms the idle bound; and the dominant expensive shape is one large `_RAW.tar`, where concurrency buys nothing.

`DEFAULT_INSTALL_CONCURRENCY = 4` isn't the precedent it looks like — its own comment says it "keeps a handful of independent publishers busy", and `store.ts` states the concurrency is across datasets rather than within one: four sockets spread over NCBI/UniProt/Reactome/WikiPathways, not four into one GEO directory on one host.

**Your comment also surfaced a real bug, which is the more valuable half.** Chasing "why is the transfer loop different from the other two" turned up that the transfer is the one GEO request never routed through `fetchGeo`: `downloadToFile` fetched exactly once and treated any non-ok status as terminal. So a single shed 403/503 mid-Series aborted the set and threw away every artifact already staged — and it left this change's own spec delta ("the command lists **or fetches** … retries with backoff") half-implemented, with only the *listing* half true. `downloadToFile` now takes an opt-in retry schedule, defaulted off so refs is behaviourally untouched (which statuses are worth retrying is a property of the upstream, and only the caller knows which one it's talking to). Pinned by a test that fails on the previous commit.

Both declines are recorded in `geo.ts` and as D12 in design.md, so the next reader doesn't re-litigate them.

radutanasa:
@claude review

radutanasa:
@claude review

vivere-dally (dismissed):

vivere-dally (approved):

PR #226, opened 2026-07-24
feat(cli): chat-driven, provenance-safe analysis input management

vivere-dally:
## What & why

An analysis's inputs are DB-registered references that the harness profiles; today the only production way to register one is the TUI file picker. So when a user launches `inflexa` in a folder and tells the agent "the files are in this folder," the agent can neither **see** those files (the workspace file tools are scoped to the analysis tree, not the launch folder) nor **add** them — and it falls back to useless "copy the files into a folder" advice. This PR lets the agent see and manage inputs mid-chat, without forking the analysis's signed provenance chain.

### CLI — OpenSpec change `chat-input-staging`

- **`list_launch_dir`** — read-only host tool listing candidate files in the analysis's anchor/launch folder (reusing the staging noise-dir walk), marking each as already-registered or not.
- **`manage_inputs`** — in-process host tool to add / remove / list inputs during a live chat. It runs **in-process, not via a `run_inflexa` subprocess**, because the CLI event bus is in-process only and provenance is a per-process signed hash chain guarded by the per-analysis instance lock — a subprocess writer would be refused by the lock or fork the chain. In-process it mutates under the lock the chat already holds, and its `prov.input_added`/`prov.input_removed` stays on the bus the recorder and profile-parity watcher observe. Approval-gated, and it asserts the lock is held (`holdsInstanceLock`) before mutating.
- **`inflexa inputs add|remove|ls`** — the terminal (human) surface over the same shared `addInputs`/`removeInput`. `add`/`remove` acquire the analysis lock and refuse if held elsewhere; they are **agent-blocked** so the agent uses the in-process tool. `add` verifies on-disk existence (rejecting a hallucinated path); `remove` matches the registered set so a moved/deleted file's input stays removable.
- **Register-only** — staging and re-profiling stay with `input-staging` and the parity engine (an added/removed input drifts the profiled set; the existing `watchProfileParity` edges re-profile).
- Refactor: `resolveSingleAnalysis` moved from `harness/profile.ts` to `analysis/context.ts` so every analysis-scoped command reduces context identically (a multi-analysis anchor now gets a proper "pick one with `--analysis`" error).

### Harness — OpenSpec change `conversation-unstaged-data-guidance`

- Conversation prompt now offers to add referenced-but-unstaged files (host-agnostically — it names no specific host tool) and no longer frames data profiling as init-only.

### Provenance safety

An audit of all 10 `prov.*` emit sites confirmed every mutation surface already holds the analysis lock **except** the analysis-creation path, which is safely lock-free (fresh UUIDv7 → no possible second writer). So there is intentionally **no recorder-level assertion** (it would break that legitimate lock-free emit); the single-writer invariant is upheld caller-side, and `manage_inputs` adds a defense-in-depth lock check. The signed hash-chain format is unchanged. Full reasoning is in the design doc's decisions + emit-site audit.

Both OpenSpec changes are included (proposal / design / specs / tasks) and validate `--strict`.

## Type of change

## Checklist

- [x] Documentation updated if user-facing behavior changed (tool descriptions + conversation prompt carry the agent-facing guidance).

- **Methods:** N/A — no analytical method or library version is involved.
- [x] **Sandbox:** unchanged — no sandbox, mount, network, or capability changes.
- [x] **Provenance:** prior analyses remain reproducible; the signed hash-chain format is unchanged. This PR *strengthens* the single-writer invariant by adding a lock guard on the new mutation surface rather than altering the chain — see the emit-site audit in `chat-input-staging/design.md`.


vivere-dally:
@claude review

vivere-dally:
Thanks @claude — went through each observation. Two follow-up commits since your review: `f250816` and `cee35af`.

**1. Partial-failure reporting in the `manage_inputs` remove loop** — fixed in `cee35af`. Both remove surfaces now name what was already removed before a mid-loop fault: `manage_inputs` appends `(already removed: …)` to the `error` message, and the terminal `runInputsRemove` does the same in its `dieOn` message. A partial failure no longer reads as a clean no-op — each prior `removeInput` has already emitted its `prov.input_removed`.

**2. `classifyInputPath:notFound` names the whole batch, not the offender** — acknowledged, left as-is. It's reachable only on a TOCTOU race: the `existsSync` pre-check already names the specific missing path in the common case, so the batch-level message only surfaces if a file vanishes between the pre-check and `addInputs`. Noting rather than fixing this round; happy to sharpen if it's worth the extra plumbing.

**3. `askRequest` grantKey scoped to `action:analysisId`, not paths** — intended and documented (tool description + design risk section). An "always" grant deliberately blesses future add/remove for that analysis; path-scoping it would defeat the point of "always". No change.

One extra fix beyond your list, from a self-review pass: `findAnalysis(...).unwrapOr(null) → no_anchor` in both `manage_inputs` and `list_launch_dir` was collapsing three distinct misses — a DB read fault, a vanished analysis row, and a moved anchor — into one status, so a DB hiccup was reported to the agent as a moved anchor. Split in `cee35af`: DB fault → `error`, gone row → `no_analysis`, and `no_anchor` keeps its true "folder moved" meaning. Regression tests added.

typecheck / lint / tests green across the touched files. Appreciate the independent emit-site re-derivation.

radutanasa (approved):

krajla (approved):

PR #227, opened 2026-07-24
feat(cli): manual model-id entry in the palette model picker

vivere-dally:
## What & why

Free-text model-id entry existed only as a *fallback* — the palette picker offered it when the connection's `/models` listing failed, and cliproxy `setup` skipped its default-model step entirely when the proxy listing was down or listed nothing servable. In both cases an id the connection does not enumerate was unreachable exactly where a model gets chosen. Direct-mode `inflexa setup` has always prompted for the model id as free text; this makes the other two surfaces agree with it.

**1. Palette picker (`Switch chat model` / `Switch sandbox model`)**

- `src/tui/commands.tsx` — a manual-entry row is appended to the listing (`modelPickerItems`, extracted so the row set is assertable as data). Selecting it opens the same free-text `PromptDialog` and commits through the identical accessibility-validate → persist → apply path as a listed pick. The field opens empty (the current id is already listed and marked `current`); esc there returns to the listing rather than closing the picker.
- `src/tui/components/list_core.tsx` — the row is `pinned`: a new `SelectItem` flag that exempts a row from fuzzy ranking. Without it the row disappears the moment the user types the very id they opened it to enter (the filter query is a model id; it shares no subsequence with the row's own label), leaving an empty list and no visible escape hatch.
- `src/tui/components/dialog/prompt_dialog.tsx` — an optional `onBack`: esc on a prompt reached *from* another surface of the same dialog entry returns to that surface instead of closing the entry. Its footer label follows (`esc back`). `dismiss` (click-outside, ctrl+c) still closes.

**2. CLIProxy `setup` default-model step**

- `src/modules/infra/setup.ts` — the empty-list case routes to free-text manual entry with Auto (blank) as the default instead of skipping. Setup first states *which* cause applies — the proxy never answered its listing, or it answered and nothing it lists is servable — because the two call for different judgement. A typed id is accessibility-checked with the same `count_tokens` request the list uses when the proxy demonstrably answers; when the listing itself was unavailable the check is skipped, since it could only spend its full timeout to reach `inconclusive`, which persists anyway.
- `src/lib/cli.ts` — `promptTextOptional`: cancelling resolves to `null` rather than aborting the command. This step is appended to work that has already succeeded, so a reflexive ctrl+c must not exit non-zero and disown it. `promptText` keeps its abort-on-cancel contract for values a command cannot proceed without.

**Specs** — updated in place: `agent-model-selection` (manual row survives the filter; esc backs out to the list), `model-connection` (the two no-list causes, the skipped check, declining keeps Auto), `list-primitives` (`pinned`), `dialog-system` (`onBack`), `tui-components`.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`; `bun run typecheck` → 0 errors; `bun test src/tui src/modules/infra/setup.test.ts src/lib` → 759 pass / 0 fail). The full suite has 36 failures unrelated to this change — identical on `main`'s tree at this commit's base (harness agent-switch, provenance emitters, spawn e2e).
- [x] Tests added or updated for the change — pinned-row filtering (`list_primitives.render.test.tsx`), the picker's row set and an end-to-end keyboard flow through the real dialog host: filter by an unlisted id → the row survives → enter opens the empty free-text field → esc returns to the list (`model_picker.render.test.tsx`), and the setup step's two no-list causes (`setup.test.ts`).
- [x] Documentation updated (specs updated in place; `cli/CLAUDE.md` for `pinned`).
- [x] Focused on a single capability — reaching an unlisted model id — across the two surfaces that could not.


vivere-dally:
@claude review

krajla (approved):

PR #228, opened 2026-07-24
feat(harness,cli): disambiguate "cwd" between the analysis tree and t…

vivere-dally:
…he launch folder

A "what files are in the cwd?" question routed to the workspace file tools (the analysis tree) and never surfaced the launch folder, so the user had to correct the agent before it reached for list_launch_dir. The word names two scopes and nothing biased the first, ambiguous ask.

- cli list_launch_dir: description now owns the "cwd / current directory / this folder / where I ran/launched inflexa" phrasings and states the workspace tools cannot see that folder — the primary routing lever.
- harness list_files: description states its scope is the analysis's own tree only, never a directory outside it (kept context-neutral — no launch-folder/host-tool wording, since sandbox agents share this tool and receive no host tools).
- harness conversation prompt: new "Two filesystem scopes" guidance — treat a "cwd / this folder" question as ambiguous, name which scope was listed and offer the other in the same turn, and reach for the host tool first when the phrasing points squarely outside the analysis tree. Host-agnostic: names no specific host tool.

Captured the prompt behavior in the OpenSpec change conversation-filesystem-scope-guidance (host-conversation-tools delta, validates --strict). No runtime behavior change — prompt and tool descriptions are string constants.

## What & why

Closes #

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist


vivere-dally:
@krajla pls review

harness/src/prompts/conversation.ts
krajla, line 1: Remove tool specific guidance from system prompt. Tools self define.

vivere-dally: Agreed — done in afe81ea. The whole "Two filesystem scopes" block is gone from the prompt; the two tool descriptions carry it now.

Two things I found while removing it that make the point concrete:

**The enumerations were already stale.** The prompt listed `workspace_search, read_file, list_files, grep` — but `file_stat` is attached to the conversation agent too (`agents/conversation-agent.ts:286`) and sees the same tree. The `list_launch_dir` description listed `list_files / read_file / workspace_search`, omitting both `grep` (line 287) and `file_stat`. Two hand-kept lists of the same set, in two independently-versioned packages, neither typechecked, and they didn't agree with each other. Both are gone — `list_launch_dir` now just says "the workspace file tools".

**It also broke the seam's own invariant.** `host-conversation-tools/spec.md` says the harness "SHALL NOT inspect, branch on, or otherwise learn anything domain-specific about a host tool." The block taught the prompt that a launch folder exists, that a host tool may list it, and that unstaged input candidates live there. Not naming the tool id dodged the naming, not the coupling.

Routing now rests on the descriptions alone: `list_launch_dir` claims the "cwd / current directory / this folder / where I ran inflexa" phrasings, and `list_files` disclaims them by naming the host process's cwd as outside its scope — kept context-neutral, since sandbox agents share `list_files` and get no host tools.

The OpenSpec delta moved with it: the change only touches `list_files`, owned by `harness-workspace-tools`, so it no longer sits under `host-conversation-tools` and is renamed `workspace-tool-scope-in-descriptions`. Validates `--strict`.

One trade-off worth calling out: the "name which scope you listed and offer the other in the same turn" behavior is lost — a tool description can say *when to reach for me*, but not *how to phrase the answer*. The agent will now pick a scope rather than surfacing both. I think that's the right price; flagging in case you'd rather have it back somewhere.

Verified under podman (harness `1483 pass / 0 fail`, typecheck + lint clean). Separately: 36 CLI tests fail on this branch, but they fail identically on the parent commit — pre-existing, not from this PR.

radutanasa (approved):

PR #229, opened 2026-07-24
fix(cli): trust the OS certificate store so fetch works behind TLS-intercepting proxies

vivere-dally:
## What & why

Behind a corporate TLS-intercepting proxy (ZScaler, Netskope, …), `inflexa refs download` — and every other CLI `fetch` — fails with `UNABLE_TO_GET_ISSUER_CERT_LOCALLY`, even though the *same machine's* `curl` succeeds after IT installs the proxy's root certificate. The proxy re-signs HTTPS with a root CA that IT installs into the OS keychain/trust store; `curl` (via Secure Transport) trusts the keychain, but **Bun's `fetch` uses a compiled-in Mozilla root set and ignores the OS trust store by default**, so it rejects the re-signed certificate. This is documented Bun behaviour ([oven-sh/bun#23735](https://github.com/oven-sh/bun/issues/23735)); the supported remedy is opting in to the OS store via `NODE_USE_SYSTEM_CA` / `--use-system-ca`.

### The fix

Enable `--use-system-ca` **at launch**, in the two places the CLI is actually launched from:

- **Release binaries**: `scripts/build.ts` embeds `--use-system-ca` into each compiled executable via `compile.execArgv`. This also covers the self-spawned `run_inflexa` child, which is this same binary.
- **Dev runs**: the `dev` script carries a `NODE_USE_SYSTEM_CA=1` launch-env prefix, which child processes inherit.

The merge is additive — Bun's bundled Mozilla roots stay; the OS store's roots are added — so the CLI defers to the machine's own trust policy the way native tools do.

### Why launch time, not an in-process write

An earlier revision of this branch set `process.env.NODE_USE_SYSTEM_CA ??= "1"` at the top of `index.ts`. **That approach does not work**: Bun reads the variable once at process boot, before any user code runs, so an in-process write never reaches the TLS verification contexts `fetch`/`node:https` actually use. The write *looked* verified because `tls.getCACertificates("default")` re-reads the variable lazily at call time (145 → 147 roots) — but that probe reflects a different store than the one doing certificate verification. A real handshake still failed.

### How it was verified

End-to-end against a local HTTPS server whose leaf certificate is signed by an OS-keychain-trusted root that Bun does not bundle (an mkcert root — the same trust shape as a corporate proxy root), on macOS with Bun 1.3.14:

| Scenario | Result |
|---|---|
| `fetch`, defaults (pre-fix) | ❌ `UNABLE_TO_VERIFY_LEAF_SIGNATURE` |
| `fetch` after an in-process `process.env` write, even as the first statement | ❌ same failure |
| `fetch` with `NODE_USE_SYSTEM_CA=1` in the launch environment (dev-script form) | ✅ |
| binary compiled with `compile.execArgv: ["--use-system-ca"]` (release form) | ✅ |
| `curl` on the same machine | ✅ (the bug's defining asymmetry, now gone) |

A real `bun run build` of the darwin-arm64 binary passes its smoke test and embeds the flag.

### Scope and trade-offs

- **Covers in-process fetches**: auth device flow, refs download, direct-mode provider calls, embedding model/runtime downloads, OTel export.
- **Does not cover container-hosted TLS**: in `cliproxy` mode the CLI's own leg is loopback HTTP — the upstream TLS handshake to the provider happens **inside the CLIProxyAPI container**, and sandbox tools fetch inside their Docker containers. Behind an intercepting proxy those need container-level trust configuration; no CLI env var can reach them.
- **The embedded flag cannot be disabled at runtime** (`NODE_USE_SYSTEM_CA=0` does not override it). Accepted: trusting one's own OS store is the policy every native tool already applies, and the merge never drops the bundled roots.
- For a corporate root shipped as a loose `.pem` rather than installed in the OS store, `NODE_EXTRA_CA_CERTS=/path/to/root.pem` remains a fallback Bun honours (single-cert PEM, per [oven-sh/bun#13867](https://github.com/oven-sh/bun/issues/13867)).

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint (`bun run lint`) and typecheck (`bun run typecheck`) pass locally.
- [ ] `bun test`: the full suite shows the **36 pre-existing failures, unchanged by this PR** (1688 pass / 36 fail — the same counts as `main`; the failures are in provenance/artifact-registry unit tests and e2e timing, unrelated to CA trust). The targeted `src/lib/env.test.ts` (36 tests) passes.
- [ ] Tests added or updated — none: the change is build configuration plus a script prefix; there is no runtime code to unit-test, and an honest automated check needs a TLS server signed by an OS-trusted root, which the suite can't assume on a runner. The manual verification protocol is documented above.
- [ ] Documentation updated — no user-facing surface change; behaviour is transparent.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -6,7 +6,7 @@
     "scripts": {
         "build": "bun scripts/build.ts",
         "build:all": "bun scripts/build.ts --all",
-        "dev": "bun run src/index.ts",
+        "dev": "NODE_USE_SYSTEM_CA=1 bun run src/index.ts",
         "dev:install": "bun scripts/dev_install.ts",
         "docs:gen": "bun scripts/gen_docs.ts",
         "format": "prettier --write src/",
--- a/cli/scripts/build.ts
+++ b/cli/scripts/build.ts
@@ -378,6 +378,20 @@ for (const target of targets) {
             target: `bun-${target.os}-${target.arch}`,
             // Bun appends .exe automatically for windows targets.
             outfile: `dist/${name}`,
+            // Trust the OS certificate store in addition to Bun's bundled Mozilla roots, so fetch works
+            // behind corporate TLS-intercepting proxies (ZScaler, Netskope, …) whose root CA sits in the
+            // OS keychain/trust store (oven-sh/bun#23735). It MUST be an embedded runtime flag: Bun reads
+            // NODE_USE_SYSTEM_CA once at process boot, so an in-process `process.env` write is invisible
+            // to the TLS verification contexts `fetch`/`node:https` actually use — only
+            // `tls.getCACertificates()` re-reads the variable lazily, which makes that probe a false
+            // positive for the write-at-startup approach. The merge is additive (bundled roots stay), and
+            // it propagates to self-spawned children (the `run_inflexa` subprocess is this same binary).
+            // Accepted trade-off: an embedded flag cannot be disabled at runtime — NODE_USE_SYSTEM_CA=0
+            // does not override it; trusting one's own OS store is the same policy every native tool
+            // (curl included) already applies. Loose per-cert needs stay on NODE_EXTRA_CA_CERTS. The
+            // source-run counterpart is the `NODE_USE_SYSTEM_CA=1` prefix on package.json's dev script,
+            // where the launch environment satisfies the same boot-time read.
+            execArgv: ["--use-system-ca"],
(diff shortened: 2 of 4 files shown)


vivere-dally:
@claude review

radutanasa (dismissed):

radutanasa (approved):

PR #231, opened 2026-07-24
feat(cli): manual model-id entry in cliproxy setup when the listing is unavailable

vivere-dally:
## What & why

`inflexa setup` (CLIPROXY mode) previously **skipped** the default-model step entirely whenever the proxy `/models` listing was down/empty or listed nothing the account can serve — leaving no way to pin a model id the listing could not enumerate. Direct-mode setup and the palette picker both allow free-text entry in that situation, so cliproxy setup was the odd one out.

This routes the empty-list case to a **free-text manual-entry prompt** with **Auto** (blank) as the default, instead of skipping:

- Blank keeps the adaptive `model: null` default — the step stays optional, so it never becomes a new way for setup to fail.
- A typed id is validated with the **same** `count_tokens` accessibility check the list uses: a definite `not_found_error` is rejected with a warning that keeps Auto; a `served`/inconclusive outcome persists the id to **both** user-facing agents.
- A non-TTY still skips (the `isInteractive` guard returns first).

This is the setup-side companion to #227 (manual entry in the palette picker); together they make manual model-id entry uniformly available wherever the `/models` listing is unavailable.

**Changes**

- `src/modules/infra/setup.ts` — `selectDefaultModel` now falls back to `promptManual` when there is no offerable list (empty candidates, or a sweep that rules out every candidate); wired in `runDefaultModelSetup` via `promptText`. New `promptManual` seam on `DefaultModelDeps`.
- `openspec/specs/model-connection/spec.md` — updated in place: the CLIProxy default-model requirement now specifies the manual-entry fallback, plus a covering scenario.
- `src/modules/infra/setup.test.ts` — four new cases (served id pins both agents; blank keeps Auto; typed `not_found` warns and keeps Auto; an all-inaccessible sweep also routes to manual entry).

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and tests pass locally (`bun run lint`; `bun test src/modules/infra/setup.test.ts` → 110 pass / 0 fail). No new `typecheck` errors in the setup files; the repo's pre-existing `typecheck` failures are unrelated harness-API drift.
- [x] Documentation updated (the `model-connection` spec is updated in place).

### DIFF
--- a/cli/openspec/specs/model-connection/spec.md
+++ b/cli/openspec/specs/model-connection/spec.md
@@ -161,12 +161,17 @@ agents. On a non-TTY the direct path SHALL write no model, and boot's actionable
 failure remains the contract.
 
 After the CLIProxy login, once the provisioned proxy is answering (setup runs no credential
-probe — that is the launch gate's; the step skips gracefully, writing nothing, when the proxy
-or its listing is not available), interactive setup SHALL present a default-model
+probe — that is the launch gate's), interactive setup SHALL present a default-model
 selection: a preselected **Auto** row labeled with the currently elected model
 (`default-model-election`), followed by the connection-family models from the proxy's `/models`
 list, accessibility-checked via bounded-concurrency `count_tokens` requests — only a definite
 `not_found_error` excludes a model from the list; an inconclusive check keeps it listed.
+When the proxy listing is unavailable — a down or not-yet-answering proxy — or nothing it lists is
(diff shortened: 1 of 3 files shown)


vivere-dally:
Folded into #227 (the setup change is cherry-picked there as 3c64724). Closing this in favor of the single PR.

PR #237, opened 2026-07-27
feat(cli): non-interactive setup — `setup --yes` batch mode + `--config` answers file

vivere-dally:
Makes `inflexa setup` fully scriptable for fleet provisioning: `--yes` turns setup into a batch run that never prompts, and `--config file.yml` supplies answers from a file — so N client machines with the same configuration can be provisioned unattended, with misconfigurations failing at provision time instead of on the client's first chat.

## The answers mechanism

Every interactive decision in the setup wizard is now a named **answer** with one resolution chain:

```
flag  >  --config file value  >  interactive prompt (TTY, no --yes)  >  default-or-error (--yes / non-TTY)
```

- A supplied answer skips its prompt **even interactively** — ship a file answering 90% of the questions and a human answers the rest.
- Under `--yes` the **whole answer set is validated before anything mutates**: unknown file keys, missing required answers, reserved Postgres ports, and mode-mismatched answers (e.g. `--base-url` under cliproxy) all fail upfront, each error naming both spellings (`--base-url` / `connection.baseURL`). No answer is ever silently ignored.
- The file carries *configuration answers only*; invocation behavior (`--no-start`, `--force`, `--no-validate`, …) stays flag-only, so a fleet file can't bake per-run behavior.
- Validation probes default **on** and fail hard on anything short of a pass (the 1-token model ping, the credential-source probe ladder, the embedding endpoint probe). `--no-validate` skips the network probes for air-gapped staging; offline GGUF verification always runs.
- Secrets never ride flags or the file: the model key stays `INFLEXA_MODEL_API_KEY`/provider-conventional env, and api-key embeddings read `INFLEXA_EMBEDDING_API_KEY`.

## Usage

```bash
# Pre-stage a cliproxy client (everything except the browser sign-in; first launch offers it) — exits 0
inflexa setup --yes --refs recommended --sandbox python

# Fully unattended direct-connection client
INFLEXA_MODEL_API_KEY=sk-… inflexa setup --yes \
  --connection direct --base-url https://api.anthropic.com/v1 \
  --provider anthropic --model claude-sonnet-5

# Fleet: one answers file, flags override per-machine
inflexa setup --yes --config fleet.yml --postgres-password "$PG_PASS"
```

```yaml
# fleet.yml
connection:
  mode: direct
  provider: anthropic
  baseURL: https://api.anthropic.com/v1
  model: claude-sonnet-5
postgres: { user: inflexa, port: 5555 }
resources: { sharePct: 50 }
embedding: { mode: local }
refs: recommended
sandbox: python
runtime: docker
```

All 20 batch options are grouped under `Batch mode:` in `inflexa setup --help`.

## BREAKING changes

- **`setup --yes` never prompts, even on a TTY** (previously it only pre-consented reference downloads).
- **Headless `--yes` no longer auto-downloads the recommended reference set** — the value is the consent; pass `--refs recommended` to restore the old outcome.
- **Non-TTY direct setup no longer adopts `ANTHROPIC_*`/`OPENAI_*` from the environment** — batch requires explicit `--base-url`/`--provider` (adoption remains an interactive affordance).

Minor: an answered `--embeddings off` now persists `embedding.mode: "off"` (an answer declares state); the interactive picker's "Off / skip" still leaves config unchanged.

## Notes for reviewers

- Batch cliproxy is deliberately **pre-staging**: provider OAuth needs a human in a browser and a refresh-token credential cannot be shared across a fleet, so a missing credential is a notice + exit 0, and the first launch offers the sign-in.
- The cliproxy `count_tokens` accessibility check keeps its inconclusive-proceeds behavior (a pre-staged proxy has no credential loaded); strict pass-or-fail applies to the probes that validate user-supplied endpoint facts.
- An answered runtime (`--runtime`) is a hard gate — probed alone, never silently switched — so a fleet can't end up heterogeneous because Docker happened to be down on one client.

## Verification

- 450+ scoped tests green (`infra`, `embedding`, `refs`, `cli`), including batch integration tests: bare-`--yes` defaults, fail-before-mutate, ambiguous-probe fail-hard, idempotent second run (byte-identical config), runtime hard gate.
- Live smokes: typo'd/missing/unparseable `--config`, missing direct answers, and mode-mismatch all exit 1 before any mutation with both-spellings errors.
- `docs:gen` green (all options documented); agent-policy snapshot unchanged (`setup` stays `blocked`).
- Full design/spec trail in `cli/openspec/changes/non-interactive-setup/` (proposal, design D1–D12, delta specs, 30/30 tasks).


krajla:
@claude can this be invoked also with a longer inflexa setup with a bunch or args instead of a yaml file?

krajla (approved):

radutanasa (approved):

PR #238, opened 2026-07-27
fix(tui): recover the chat after a terminal-side screen wipe (Terminal.app ⌘K)

vivere-dally:
## What & why

Pressing ⌘K in macOS Terminal.app while the chat is open blanks the entire UI, and it stays blank. The only recovery is scrolling the terminal back up to find the old frame sitting in scrollback, which reads as the app having navigated somewhere. ⌘K is an easy mis-keypress precisely because the command palette is `ctrl+k`.

Two independent facts meet to cause it.

**Terminal.app's ⌘K is "Clear to Start"**, not "Clear Scrollback". I read this from the app's own menu definition rather than inferring it: `Clear to Start` carries the `k` key equivalent and invokes `clearAll:`, while `Clear Scrollback` (`clearScrollback:`) has no key equivalent and `Clear to Previous Mark` takes `l`. "Clear to Start" scrolls the visible screen into scrollback and leaves a blank viewport — and it does this to the alternate screen too, which is why the old frame is still scrollable. Being a menu action, it sends the app no bytes, raises no `SIGWINCH`, and changes no focus state, so **there is no signal to react to**.

**OpenTUI renders diffs.** Each frame compares against a shadow buffer of what it believes is on screen and writes only changed cells. Measured in a pty on the pinned 0.4.2: a re-render with unchanged content emits **zero bytes**.

So after the wipe every surviving cell reads as "already correct" and is never rewritten. The chat renders correctly the whole time, into a screen the terminal emptied. Only cells that happen to change afterwards appear — which is why a lone scrollbar pip shows on an otherwise blank screen. This is structural to diff rendering, not a Terminal.app quirk: anything that changes the surface without going through the renderer leaves the shadow buffer lying, and no number of later frames repairs it.

Since the damage can't be detected, recovery has to be user-triggered. `forceFullRepaint` invalidates the shadow buffer so the next frame rewrites every cell, reached two ways:

- **`ctrl+l`** (`app.redraw`, remappable) — the universal terminal redraw convention, on a **mode-less** layer so it survives an open modal. A redraw key that modal capture suspends is unavailable exactly when a modal is the thing that got wiped. Also `<leader>l`, which lists it in the which-key panel.
- **Automatic heal** — one forced repaint on the first keystroke after 500 ms of keyboard idle. This is the part that matters for a non-expert: they press anything and the UI returns, with nothing to learn. The idle gap exploits the shape of the failure (a wipe is necessarily followed by a pause, because the user has to notice it) so a typing burst never crosses the threshold and normal use pays nothing.

Measurements taken in a real pty at 200x50, which drove the design:

| Action | Bytes written |
|-|-|
| Re-render, nothing changed | **0** — the mechanism behind the bug |
| Forced full repaint | 13,048 — all rows, wrapped in synchronized output, stays in the alternate screen |
| `suspend()` + `resume()` | 6,761 — but emits `CSI ?1049l` then `CSI ?1049h` |
| `resize()` jiggle | 6,553 — stays in the alternate screen |

### The one thing to review carefully

`forceFullRepaintRequested` is **private** in `@opentui/core`, and there is no public equivalent — I checked the pinned 0.4.2 *and* the current 0.4.5; neither exposes any repaint or redraw method. Both public alternatives were measured and rejected: `suspend()`+`resume()` leaves and re-enters the alternate screen, which on Terminal.app dumps the frame to scrollback and so **re-causes the symptom**; and jiggling `resize()` relayouts the tree at a wrong size for a pass and emits spurious `resized` events that every `useTerminalDimensions` consumer observes.

So the fix writes a private field, and it's guarded **at the wire** rather than in memory. Asserting that the renderer accepted and cleared the flag isn't enough — a version that kept the field but stopped honoring it would pass every in-memory check and still ship a chat that never recovers. Over a capturing stdout, the test asserts a plain re-render emits zero bytes while a forced one re-emits the frame. That also pins the *premise* of the bug, so if opentui ever stops diffing, the suite tells us this machinery can be reconsidered. I verified the guard bites by removing the field write: 5 of the 7 tests fail, including the wire test.

## Type of change

- [x] Bug fix
- [x] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). — Lint and typecheck clean; 463 TUI tests pass. To be precise about the full suite: it reports 36 failures, all pre-existing and unrelated (harness-provenance, agent-switch, e2e spawn). I confirmed that against a stashed baseline — identical 36 with and without this change, and this branch adds 7 passing tests and no new failures.
- [x] Documentation updated if user-facing behavior changed. — New `screen-repaint` spec (synced + archived via the OpenSpec flow), a postmortem in `HORRIBLE_BUG_FIXES.md`, and the key self-documents in-app through the which-key panel via its `desc`. There is no user-facing keybinding reference doc in the repo to update.

## Trade-offs and known gaps

Stated here rather than left for a reviewer to find, and written into the spec as a limitation:

- **The heal only observes keystrokes, so mouse-first users get nothing.** Someone facing a blank screen may click before typing, and mouse events don't reach the keypress observer. This is the gap I'd close first; it's recorded as a follow-up and is small enough to fold into this PR if you'd prefer.
- **Mid-stream damage isn't self-healing.** If the wipe lands while a turn is streaming, changed cells still paint, so you get text on a stale background until the keyboard is touched — arguably more confusing than a cleanly blank screen. Closing it would need a polling repaint, which would defeat on-demand rendering (an idle chat currently costs zero frames), or a damage notification no terminal sends.
- **It can mask genuine rendering bugs in our own TUI.** If we ever fail to invalidate a cell correctly, the heal quietly repairs it after any pause, making real diff bugs harder to notice and reproduce. Given the `<For>`/`insertBefore` postmortem already in `HORRIBLE_BUG_FIXES.md`, silent repair has a track record here. Noted in the spec's design so a future investigator knows to disable the heal while reproducing.
- **Remote sessions pay for a local terminal's bug.** A forced repaint scales with terminal area (~30 KB at 300x80), so the first keystroke after a pause could be a perceptible hitch over a slow link. Narrowing the heal by terminal identity was considered and rejected as fragile, since multiplexers and other terminals damage screens too.
- **`ctrl+l` is now globally reserved** at default priority on a mode-less layer. Equal-priority layers tiebreak on registration order and this one registers during `App` setup, so a future dialog binding `ctrl+l` would silently lose rather than error.

The durable resolution for the private-field dependency is upstream: a public repaint method on `CliRenderer`. Recorded as a follow-up.


vivere-dally:
TBD

MacOS's default terminal is not a suitable home for inflexa tui currently, the colors don't render right, and this patch may introduce regressions that we are not currently aware of.
If it becomes a pressing issue such that we really need to make support for this terminal as seamless as the other ones, then we can resume work on this branch.

vivere-dally:
Not needed anymore, will reopen if we do

PR #240, opened 2026-07-27
Pin sandbox image versions and clean superseded images

krajla:
## Summary

- publish immutable sandbox version and revision metadata on built images
- use `latest` only for discovery, then retag and persist the resolved versioned image reference
- verify pulled image identity before committing configuration and roll back safely on failures
- remove superseded versions only within the managed sandbox repository, without global prune or forced deletion
- restore the exact configured image after preflight checks and expose the pinned version in status output
- sync the provisioning specification and archive the completed OpenSpec change

## Why

Persisting `latest` obscured when an installation moved between sandbox releases and made lifecycle cleanup ambiguous. Recording the immutable version makes upgrades observable and provides a safe ownership boundary for removing old images and unused layers.

## Impact

Existing installs continue to pull through `latest`, but successful pulls now update configuration to the resolved version tag. Superseded managed sandbox versions are cleaned up after the new version is committed, while unrelated images and images still referenced by containers are preserved.

## Validation

- `bun run typecheck`
- `bun run lint`
- shell syntax validation for `scripts/lib-store-label-packages.sh`
- 183 CLI regression tests passed
- focused image and preflight tests passed
- `openspec validate --all --strict` — 77 specs passed
- `git diff --check`

### DIFF
--- a/.github/workflows/lib-store.yml
+++ b/.github/workflows/lib-store.yml
@@ -156,7 +156,7 @@ jobs:
           # Stamp the inventory on BEFORE the push: a host that never bind-mounts the
           # store reads the label, so an unlabelled published image is one whose agents
           # cannot tell what is installed.
-          bash scripts/lib-store-label-packages.sh "${IMAGE_PYTHON}:${VERSION}-${ARCH}"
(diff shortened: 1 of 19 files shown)


vivere-dally:
@krajla is this relevant anymore?

krajla:
No @vivere-dally

PR #249, opened 2026-07-28
feat(cli): recall previously sent prompts with up/down in the composer

### LINKED ISSUE #242: Being able to go back in history of commands quickly.
### Pre-flight

- [x] I searched existing [issues](../../issues) and [Discussions](../../discussions) and didn't find a duplicate.

### Problem / motivation

Can you have a feature that allows users to look back at command history quickly and either use the up arrows to go back to previous command history. This is a convenient feature. Not eesential. 

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Area

CLI / TUI experience

radutanasa:
Closes #242.

## Why

The composer had no way to get a sent prompt back — re-running a near-identical request, or recovering the wording after a failed turn, meant retyping it.

The `up` chord was reserved for exactly this when the retract binding landed. `key-bindings` pinned a scenario saying so:

> **Scenario: Idle up-arrow does nothing** — nothing happens (reserved for future history recall)

This cashes that reservation in.

## What it does

`up` in the composer walks back through the prompts already sent in this session, `down` walks forward, and `down` past the newest restores the empty buffer. Composer-only — the stream pane's `up`/`down` stay scroll keys.

## Design notes worth a reviewer's attention

**Entries come from the conversation store, not a new record of what was typed.** That makes the exclusions structural rather than a filter list someone has to maintain: `pushUserMessage` is called only by `send`, so docked-ask answers and the `/quit` aliases can never appear, and a retract splices its turn out of the store. It also means recall survives a restart, and needs no table, query, or migration.

**The recall position is a stored index, never a search of the entry list.** Adjacent duplicates collapse (a prompt re-sent after a failed turn is one step), but non-adjacent ones stay distinct — so `A, B, A` keeps two `A`s. An `indexOf`-style lookup would resolve both to the newest and silently skip `B` on the way back. There is a test for exactly this; it is the kind of bug a fixture without a repeated non-adjacent prompt never catches.

**Recall is gated on `!canRetract()` — the exact negation of the retract's own gate.** `up` has two claimants on one focus target, and negating makes them mutually exclusive *by construction*, whatever order the layers were registered in. Priority would only decide who wins when both are live. Recall takes the chord over the moment the retract window closes, including mid-turn once output has arrived, so the next message can be prepared while the agent works.

**Staying in recall across presses is derived, not tracked.** The seed makes the buffer non-empty, so an emptiness-only gate would strand the user one entry deep. The layer stays live while the buffer still *equals* the entry it seeded, and goes inert as soon as it differs. No edit subscription, and no flag needing to be cleared on submit, clear-input, session switch, or a retract seeding the composer — being a function of the buffer, it is right in all of them for free. The same reasoning kills the reset wiring: entry from an empty buffer always resumes at the newest entry, so an abandoned recall leaves nothing stale behind.

**The bindings array is rebuilt per keystroke**, so a chord is bound only when it will act — `down` only during a recall, `up` only when its target index has an entry. The engine `preventDefault`s whatever it matches, so a binding that runs a no-op steals the key from the editor underneath. (`preventDefault: false` does not help: a matched binding returns `handled` and stops dispatch regardless.)

## Known limit, spec'd rather than discovered

History reaches only as far as the mounted window (`MESSAGE_CAP = 200`). Older turns live in the thread but are not mounted, so recall stops there. Written into the spec as behavior; paging is a follow-up if anyone hits it.

## Out of scope

Fuzzy history search over the same entries (a `ctrl+r`-style picker), cross-session/cross-analysis recall, and recall from the stream pane. Each is additive on top of this.

## Spec impact

One modified capability, `key-bindings`: a new requirement for recall, plus an amendment to the retract requirement whose prose reserved the chord and whose "Idle up-arrow does nothing" scenario this replaces.

The OpenSpec change ships in this PR but is **not archived** — archiving rewrites the main `key-bindings` spec, which seemed worth a separate beat after review.

## Verification

```
bun run typecheck   →  clean
bun run lint        →  clean
bun test src/tui/   →  484 pass, 0 fail (48 files)
openspec validate prompt-history-recall --strict  →  valid
```

19 new tests: 7 unit for `promptHistory()`, 12 rendered dispatch tests driving the real keyboard bus through the exported factory (not a hand-copied replica), plus one added to the existing retract suite proving a retracted prompt leaves no entry.

One gap stated plainly: the ask-answer and `/quit` exclusions are not separately testable here — reaching them needs `App.handleSubmit`, which the render harnesses deliberately do not mount. They hold because those paths never call `send`, and their precedence is already pinned by `ask_answer.test.ts`.


radutanasa:
All four review findings addressed in `0e615c6`.

## 1. Multi-line navigability (HIGH) — fixed with the readline rule

A chord now steps history only from the buffer **edge it moves away from**: `up` recalls from the first row, `down` from the last, and on every row in between neither is bound, so both fall through to the textarea's own caret movement. A single-line entry is the first and last row at once, so the common case still recalls in one press per direction and pays nothing for the rule.

Rows are **logical**, read from `editBuffer.getCursorPosition().row` against `getLineCount()` — not wrapped display lines. A soft-wrapped long prompt is one row however narrow the terminal; a chord that recalled at 120 columns and moved the caret at 60 would be indefensible.

One consequence worth naming rather than burying: stepping back through several multi-line entries now costs one keystroke per row of each entry traversed, because each seed lands the caret at the end. That is exactly how bash and zsh behave, and the alternative — seeding the caret at the top when travelling backwards — just inverts the problem onto `down`.

Covered by three new tests, including the review's exact reproduction:

```
entries = ["line one\nline two\nline three", "older prompt"]
up  →  seeded, caret row 2
up  →  caret row 1, buffer unchanged
up  →  caret row 0, buffer unchanged
up  →  "older prompt"
```

## 2. `promptHistory()` on every keystroke — fixed

Guarded to the states that can use it. Entries are needed only when the buffer is empty (to enter recall) or a position is set (to test the equality gate), so the ordinary typing path — non-empty buffer, no recall in progress — now reads nothing. Confirms your reading of `activeLayers`: it maps every thunk before filtering, so the unguarded version really was walking the transcript on keys typed inside dialogs where the layer can never fire.

## 3. Comma operator — fixed

Statement body, and the redundant `index`/`target` checks now carry a comment saying they exist for tsc narrowing that `inRecall` already implies.

## 4. Discoverability — placeholder affordance

Used the empty-buffer placeholder rather than the footer. The footer already carries the mode word, the interrupt affordance, and the newline hint; a fourth permanent item would crowd the one row whose value is being scannable mid-turn.

The placeholder's visibility conditions already match the feature's exactly: it renders only when the buffer is empty — which is precisely when a recall can be entered — and vanishes the instant the user types, so it cannot become furniture. `ChatBar` takes a host-derived `canRecall` (keeping its no-domain-imports rule) and appends the affordance, labelled from a new `RECALL_LABEL` so the advertised key derives from the bound chord instead of being hand-written.

Two honesty gates mirroring the interrupt hint's: absent while the retract owns the chord, and outranked entirely by a boot gate — that placeholder explains why typing goes nowhere, and a recall hint on top would advertise a chord whose result could not be sent.

## Spec impact

`key-bindings` gains the caret rule and three scenarios. A new `tui-layout` delta owns the placeholder affordance, since that spec already owns `ChatBar`'s composition and placeholder.

## Verification

```
bun run typecheck   →  clean
bun run lint        →  clean
bun test src/tui/   →  490 pass, 0 fail  (was 484; +3 multi-line recall, +3 placeholder)
openspec validate prompt-history-recall --strict  →  valid
```

radutanasa:
@claude review

radutanasa:
Confirmed and fixed in `5911bf5`. Good catch — the guard was expressed in terms of the wrong variable, and the failure mode is worse than "sometimes": it's permanent for the session after a single recall-then-edit.

## Why it happened

The two decisions were individually sound and jointly wrong. Design decision 3 makes a position deliberately **outlive** its recall — nothing clears it on an edit, submit, clear-input, or session swap, because it's simply overwritten at the next entry, and that's precisely what buys the layer its freedom from lifecycle wiring. Guarding on `index !== null` then asks a question that goes false exactly never. First abandoned recall, walk re-armed, rest of session.

## The fix

Rather than guard the read more carefully, I removed the need for it — which is your suggestion, taken one step further so `entries()` leaves the config path entirely rather than merely being reached less often.

A position now carries the text it seeded:

```ts
export type RecallPosition = { readonly index: number; readonly text: string };
```

so liveness is `buffer === position.text` — O(1), no entry list involved. `promptHistory()` is now called **only** inside a binding's `run`, on a press that actually steps history.

That left one config-time question: *is there anything at all to recall?* — needed so an empty history leaves `up` unbound and falling through to the editor (a pinned spec scenario). Answered by a new `hasPromptHistory()` that returns at the first qualifying turn rather than building the list, so it's a step or two from the tail in the common case.

Index and text travel in one value rather than two signals because they must never disagree: updated separately they could address one entry while holding another's text — the exact confusion the stored-index rule exists to prevent. The index is still what *steps* (entry texts are non-unique, so only a position walks the list correctly); the text is only what *gates*.

## Test

You were right that nothing covered it, and a behavioral test can't — the symptom is invisible in output. So it's a call-count assertion:

```
up                      →  entryReads = 1   (the press that stepped history)
type "!"                    ← abandons the recall, leaving a stale position
type "more text here"
up, down
                        →  entryReads still 1
```

It fails against the previous implementation, which is what makes it a regression test rather than a description.

## Verification

```
bun run typecheck   →  clean
bun run lint        →  clean
bun test src/tui/   →  495 pass, 0 fail  (was 490; +1 call-count, +4 hasPromptHistory)
openspec validate prompt-history-recall --strict  →  valid
```

Design decision 11 is rewritten to record the interaction rather than just the new shape — the failed guard is more instructive than the fix, since anything else reading state per keystroke in this layer will hit the same trap.

radutanasa:
@claude review

radutanasa:
Verified and fixed in `b7cb548`.

## Reproduced first

Wrote the test before touching the code, against a multi-line single-entry history with the caret walked to row 0:

```
FAIL  up at the oldest MULTI-LINE entry is a true no-op, caret included
      expect(caret.row).toBe(0)
      Expected: 0
      Received: 2
```

Exactly as described: the text is byte-identical, and the caret is teleported from row 0 to the last row.

## Fix

`step` now short-circuits when the resolved index **and** text both match the live position — a step that resolves to where it already is touches nothing:

```ts
if (inRecall && position !== null && to === position.index && text === position.text) return;
```

One thing worth flagging about the shape of that guard, since it's the part that could go wrong later: it has to be gated on `inRecall`. A position deliberately outlives its recall (decision 3, the same property behind the last finding), so an ungated equality check would let a *stale* position suppress a legitimate fresh entry that happened to address the same index and text — recall from an empty buffer would silently do nothing. The two prior findings and this one are all the same underlying hazard: a position that is allowed to be stale must never be read without first confirming the recall is live.

Also confirmed the symmetric case doesn't exist on `down`: it either leaves recall at index 0 or steps to a strictly different index, so it has no clamped-hold branch.

## Why existing coverage missed it

Your diagnosis is right. `"up at the oldest entry holds rather than falling off the end"` uses single-line entries where row 0 *is* the last row, so the caret jump has nowhere to show; and the multi-line walk test always has an older entry to reach, so it never exercises the clamp. The new test is multi-line specifically to close that intersection, and the design doc now records that the case is invisible on single-line fixtures — that's the part a future reader needs, more than the fix itself.

The spec's `Up at the oldest entry stays put` scenario now pins the caret alongside the buffer.

## Verification

```
bun run typecheck   →  clean
bun run lint        →  clean
bun test src/tui/   →  496 pass, 0 fail  (was 495)
openspec validate prompt-history-recall --strict  →  valid
```

cli/src/tui/app.tsx
```diff
@@ -174,6 +198,179 @@ export function paneRetractLayer(deps: PaneRetractLayerDeps): LayerConfig {
...
+    }
+
+    return {
+        mode: MODE_BASE,
+        target,
+        enabled: !deps.conversation.canRetract() && (buffer === "" || inRecall),
+        bindings,
+    };
+}
+
```
vivere-dally: Extract this into a hook file, or something else.
`app.tsx` should be a composer of hooks and it should only define the layout. Not own a business layer vertical

vivere-dally (changes requested):

cli/src/tui/app.tsx
radutanasa: Done in `d629541` — and moved as a *hook*, not just a relocated factory, since the point was app.tsx's role rather than the file it happened to sit in.

`src/tui/hooks/prompt_recall.ts` now owns the whole vertical:

```ts
export function usePromptRecall(composer: () => TextareaRenderable | null): PromptRecall
```

It holds the recall position, registers the layer, and returns only the `canRecall` state the ChatBar renders. app.tsx's side is one line and no state:

```ts
const recall = usePromptRecall(() => textareaRef);
...
<ChatBar canRecall={recall.canRecall()} />
```

which matches how `watchSidebarData` / `watchProfileParity` are already consumed. The composer is a getter because the textarea ref is assigned after the hook runs and the layer re-reads it every keystroke.

Two follow-on moves:

- `historyRecallLayer` stays exported from the hook module, so the dispatch test keeps driving the same config the hook installs over injected seams rather than a replica. The test moved to `hooks/prompt_recall.render.test.tsx`, beside it.
- The set-text + caret-to-end pair shared by the retract seed and recall became `seedTextareaText` in `components/text_area.tsx`. With callers in two modules it's genuinely multi-caller, so a shared home is justified where duplicating three lines wouldn't be.

app.tsx: **1215 → 1003 lines**. No behavior change — full TUI suite green at 509.

One thing worth settling for next time, since I got it wrong in the other direction: `cli/CLAUDE.md` says *"Don't extract single-caller helpers or sub-components into separate files… A new file is justified only when multiple callers exist"*, and I cited it in the change's task list as the reason to keep this in app.tsx. Your rule is a different axis — about what app.tsx is *for*, not about premature abstraction — and I think it should win where the two meet. If you agree, it may be worth a line in that file saying so, because the current wording reads as a blanket prohibition and `retractLayer` / `paneRetractLayer` / `interruptLayer` / `selectionClearLayer` are all still verticals living in app.tsx under it.

I deliberately left those four alone: they predate this PR, and refactoring them here would bury the feature diff. Happy to do them in a follow-up if you want the file consistent.

vivere-dally (dismissed):

vivere-dally (approved):

PR #261, opened 2026-07-29
fix(cli,harness): a failed data profile no longer withholds newly registered inputs

### LINKED ISSUE #258: Reliability Data profiler race condition on newly-registered inputs
### Pre-flight

- [x] I searched existing [issues](../../issues) and [Discussions](../../discussions) and didn't find a duplicate.
- [x] This is not a security vulnerability (those go through SECURITY.md).

### What happened?

 When new input files are registered mid-session, the data profiler appears to trigger immediately — before the underlying files are staged into the analysis workspace filesystem. This produces an outright failure rather than a wait/retry.

Impact: Medium. Confusing failure mode — a "Bad Request" with no earlier-profile fallback gives no indication that the real cause is a timing race, not a data problem. Wasted a full round-trip before the actual files were even accessible for inspection.

### Steps to reproduce

1. Register new input files mid-session (via the input-management flow)
2. Immediately call inspect_data_profile
3. Observed result: {"error":"Bad Request","state":"failed","message":"Data profiling failed and no earlier profile exists: Bad Request."}
4. At the same moment, list_files on the expected staging directory shows the new files are not yet present
5. Several turns later, once files are confirmed staged (list_files shows them), profiling succeeds normally

### Expected behavior

Have profiling wait for input-staging confirmation before running, or retry once on Bad Request if the input set changed within the last N seconds.

### Inflexa version

0.8.0

### Operating system

macOS (Apple Silicon)

### CPU architecture

n/a

### Docker version

n/a

### How did you install Inflexa?

Other (note in description)

### Relevant logs or output

```shell

```

vivere-dally:
## What & why

`inspect_data_profile` returned `Data profiling failed and no earlier profile exists: Bad Request.` and the reporter concluded — reasonably — that the newly-added files were the problem. Two independent defects produced that experience.

**The files never landed.** Materializing an analysis's inputs is currently a side effect of deciding to profile them: `ensureProfileAtParity` returns `skipped_failed` on a `failed` ledger row *before* it reaches `stageAndSeed`, the only call on that path that writes files. So after any profile failure, every input registered afterwards is recorded in SQLite and never reaches the workspace tree — `list_files` shows nothing, the sandbox mounts nothing, and the skip is silent by design. Recovery required a deliberate re-profile that nothing prompted.

Two specs disagreed and the implementation followed the wrong one. `analysis-input-management` promises that drift "SHALL detect the drift and re-profile", with materialization delegated to `input-staging`; `tui-harness-chat` said a `failed` profile gets "no auto-retry". On a `failed` row the second won, the first was silently false, and — because `input-staging` owns the mechanism while the parity ladder owns the only trigger — materialization ended up with no caller at all.

**The error was unreadable.** The `provider` arm of `toProviderError` is the only one of four that forwards the SDK message undecorated. When the AI SDK cannot parse a 4xx body against the provider's error schema it falls back to `response.statusText`, so the bare HTTP reason phrase reached the ledger. The status, workload, endpoint, and the response body the SDK *did* capture were all still on `.cause` and were discarded there.

This is not a race, which is how the issue reads it. `inspect_data_profile` only reads the ledger row and can trigger nothing — it was a terminal state with no recovery edge and no feedback.

Closes #258

## What changed

**cli — `stage-inputs-independent-of-profile`**

- Materialization moves above the profile-status gates. No ledger state except a live `running` run can keep a registered input off disk.
- The `failed` special case is *removed* from the drift rule rather than extended: a set that drifted since the failure claims `failed → running` and runs (the same recovery `forceReprofile` already performs), while a set unchanged since the failure stays deliberate-retry-only. A retry is reachable only through a fresh user edit, so it cannot loop unattended. **This is a behaviour change** relative to the current `tui-harness-chat` requirement that a `failed` profile gets no auto-retry at all.
- "Already materialized" is answered from the staged tree itself — no new persisted file, because the tree *is* the record. That needed one correctness fix: `copyFileSync` does not preserve the source's mtime, so the cross-filesystem fallback stamps it. The stamp is **confined to the copy branch** — a hardlinked staged file shares the source inode, so stamping there would rewrite the user's own input file and make `enumerateInputSignatures` report drift forever. Both behaviours were measured, not assumed.
- Parity edge 3 fires on `running → failed` as well as `running → completed`, so inputs registered during a run that then dies are not skipped twice and lost.

**harness — `data-profile-failure-diagnostics`**

- The `provider` arm composes the workload, the HTTP status, and a 120-character response-body excerpt instead of forwarding the SDK message verbatim. The cap is derived from the tightest downstream consumer, which truncates the whole line at 200, so the excerpt cannot evict the lead.
- `inspect_data_profile`'s `failed` variant carries `failedAt` and states that it records an earlier attempt, rather than implying a verdict on the inputs the analysis holds now. It deliberately reports **no** staleness verdict: the tool reads one ledger row and the current input set is embedder knowledge, so that comparison is not derivable here.
- `isBudgetExceeded` keys on the classified value when one is present. Carrying a response body into the message means a non-402 whose body merely mentions a budget would otherwise trip the text heuristic and self-cancel the step; the patterns remain the fallback for unclassified throwables, which is what they were written for.

## Verification

Reproduced #258 against the real stack — real SQLite, real filesystem, real Postgres ledger, with only the DBOS dispatch stubbed (it needs a booted runtime). After forcing a `failed` row, a newly registered input lands under `data/inputs/local/` and the drift claims `failed → running`. Mutation-checked: reverting the ladder to the pre-fix early return makes the repro fail, and restoring it makes it pass, so the repro genuinely detects the bug.

The harness chain was driven end to end (real AI SDK anthropic provider → `toProviderError` → `ResultError` → `profileFailureReason` → real Postgres → `inspect_data_profile`). The ledger line is 156 characters, single line, inside the 200-character budget:

```
Provider call failed for analysis:… (HTTP 400): Bad Request — response body: {"error":"model 'claude-opus-5' is not enabled for this credential"}
```

The running cliproxy was probed directly and does answer a non-conforming `{"error":"…"}` body rather than the Anthropic `{type:"error",…}` envelope — which is exactly what triggers the SDK's statusText fallback, so the diagnosis is grounded rather than inferred.

**Known gaps, stated rather than hidden:**

- An interactive opentui session was not driven, so the agent-visible `list_files` step is inferred from the file being present on disk, which is what that tool reads.
- A `failed` row whose input is *truncate-edited in place* gets no auto-retry: under hardlinking the staged path is the source inode, so that edit is invisible to the materialization predicate. Drift against a *completed* profile is unaffected. Documented in the `input-staging` delta.
- `already_profiled` reports its set materialized by *inference*, not by check. A hand-deleted staged tree under a completed-at-parity analysis yields no files. Verified pre-existing (the old ladder also returned `already_profiled` above the only tree-writing call); the deliberate re-profile remains the repair path.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). — harness: `tsc` clean, `bun test` 1503 pass / 0 fail. cli: lint + typecheck clean, touched suites 111 pass / 0 fail. The cli full suite has 34 failures that are **pre-existing and unrelated**: verified by running the full suite in a pristine `git worktree` at `main` and diffing the failing-suite set, which is byte-identical.
- [x] Documentation updated if user-facing behavior changed. — the two OpenSpec deltas, committed separately ahead of the code.
- [ ] This PR is focused on a single logical change. — see the note below.

**Note on scope.** This is two subsystem-scoped OpenSpec changes in one PR, because each subsystem owns its own spec tree and the fix genuinely spans both. They are independent — neither blocks the other — and together they cover the reported experience: the cli change makes the files land, the harness change makes the failure readable. Happy to split them into two PRs if you would rather review them separately.


vivere-dally:
@claude review

vivere-dally:
@krajla take a look please

cli/src/modules/harness/profile_trigger.ts
```diff
@@ -36,20 +41,36 @@ import type { HarnessRuntime } from "./runtime.ts";
...
-    | { kind: "skipped_failed" }
-    | { kind: "no_inputs" }
-    | { kind: "failed"; reason: string };
+    | { kind: "triggered"; restarted: boolean; staged: boolean }
+    | { kind: "already_profiled"; staged: boolean }
+    | { kind: "already_running"; staged: boolean }
+    | { kind: "cleared"; staged: boolean }
+    | { kind: "skipped_failed"; staged: boolean }
+    | { kind: "no_inputs"; staged: boolean }
+    | { kind: "failed"; reason: string; staged: boolean };
```
krajla: Can all of these states exist in non staged or staged?

vivere-dally: Good question — and no, they can't. I went through every return site: `staged` is fixed for five of the seven kinds; only `already_running` and `failed` genuinely occur with both values.

First, what `staged` actually means: it's the **materialization state the check finished in** — "the analysis's current input set is on disk in the workspace tree (`data/inputs/local/`), readable by the workspace tools and mountable by the sandbox." It is *not* "this drive did the staging." That distinction is the whole reason `already_profiled` reports `staged: true` despite writing nothing — a completed profile at parity implies its files are already there.

Given that, per kind:

- `triggered` — always `true` (a trigger path always materializes first)
- `already_profiled` — always `true` (parity implies on-disk)
- `skipped_failed` — always `true` (the predicate found it materialized; that's precisely why it's skipped)
- `cleared` / `no_inputs` — always `false` (empty input set, nothing to stage)
- `already_running` — **both**: `false` when the live-run gate suppresses staging up front (the sandbox is reading the tree staging would reconcile-delete), `true` when this drive staged and then lost the trigger CAS to a concurrent run
- `failed` — **both**: `false` for a fault before/at materialization (enumerate, ledger read, dataDir resolve, the staging write itself), `true` for a fault after the files landed (seed, trigger, retry)

So `staged: boolean` on all seven was wider than the code and hid that. Fixed by pinning the five fixed kinds to literal types (`staged: true` / `staged: false`) and leaving `boolean` only on `already_running` and `failed` — the union now states which states are bivalent instead of a doc comment claiming it, and a consumer that switches on `kind` gets `staged` narrowed to a literal for those five.

vivere-dally: Follow-up: renamed the field `staged` → `materialized` (d22affe). Your question landed on the deeper issue — `staged` is the *act* verb, but the field is a *state* predicate ("is the current input set on disk when the check returns?"), which is exactly why it kept needing the "NOT whether this drive did the writing" disclaimer.

The file already keeps those two vocabularies apart: `stage`/`staging` for the act (the `stage` seam, `stageInputs`, the "staged tree") and `materialize`/`materialization` for the state (`isInputSetMaterialized`, `materializeInputs`). `materialized` drops the field into the state vocabulary and lines up with the seam that computes it. Everything above still holds — only the name changed (and the five/two split is now in the type).

radutanasa (approved):

Issue #248, opened 2026-07-28
Close the run → agent loop: the conversation agent is never told a run finished


vivere-dally:
## Context

Follows from #241 and the #244 exploration. In #241's transcript the agent launches a run, says *"I'll check results when it completes"*, and the conversation stops. The user then types **"Are you done?"** — manually supplying the event loop the system does not have.

## Why it happens

Run results are pull-only by design. `harness/src/tools/research/inspect-run.ts:1-15`:

> Results are pull-only via this tool — the workflow does NOT write to the conversation thread on completion. The retired `run-result-writer` push payload (run status, step summaries, run synthesis, artifact paths) is surfaced here on demand.

So the agent *can* find out, but only when a human prompts it to look. The promise it makes at launch is one it structurally cannot keep.

### F-soft — the agent sees run state on its next turn

`prepareChatTurn` (`harness/src/app/chat-turn.ts:47`) takes `{analysisId, threadId, userInput}` and assembles context from analysis status, thread history, and working memory. There is **no run state in the assembly and no host-supplied context seam**, so this is a harness change to `prepareChatTurn` / `assembleMessages` — not CLI wiring.

Effect: "Are you done?" is answerable without a tool round-trip, and the agent can stop promising to check back.

This half carries no product decision and stands on its own.

### F-hard — the agent reacts when a run lands

The signal is already in the host process. `prov.run_completed` is emitted onto the CLI bus by `createRunProvenanceEmitter` (`cli/src/modules/harness/prov_bridge.ts:316-357`), wired into the run-engine deps at `cli/src/modules/harness/run_deps.ts:236`, and already consumed by the model-swap work gauge (`cli/src/modules/harness/agent_switch.ts:258-259`). It carries `{runId, status: completed|partial|failed|canceled, completedAtMs, durationMs?}`.

Note this does **not** require reviving the retired harness-side writer: the host owns the thread and is already its single writer, which is a different arrangement from a durable workflow writing into a thread a live TUI holds.

What is missing is a turn path. `send()` (`cli/src/tui/hooks/conversation.ts:1109`) takes `{sessionId, analysisId, userText}` and appends a user message unconditionally. Waking the agent therefore means either fabricating a user message — which pollutes both the transcript and the signed provenance record — or introducing a genuine system-note turn kind.

Two constraints any bus-driven design has to respect:

- **The CLI bus is in-process only.** This covers runs the agent launches during a chat (which is #241's scenario) but not a run started from another terminal.
[...]

PR #272, opened 2026-07-29
Add conversation run awareness and bounded run inspection

krajla:
## Summary

- inject a fresh analysis-wide `[Run Activity]` tail into conversation turns, with separate running and suspended sections
- make parameterless `inspect_run` active-first and paginated, with explicit targeted readiness states
- add abort-aware bounded waiting, cutoff metadata, and self-wait protection
- return `status: "in_progress"` from `execute_analysis` and strengthen anti-polling prompt guidance
- sync the new capability specs and archive the completed OpenSpec change

## Why

Conversation agents could lose track of long-running workflows because their context did not carry current run activity and parameterless inspection emphasized terminal history. A running run could consequently trigger repeated `inspect_run` calls based on the mistaken assumption that the run had completed or disappeared.

## Impact

Agents now see current analysis-wide run activity on every chat turn without persisting operational state into conversation history. They can perform one explicitly requested bounded wait, receive an unambiguous in-progress response at cutoff, and avoid advertising output paths before results are ready. Workflow completion remains pull-only and does not wake the conversation agent.

## Validation

- `openspec validate --all --strict` — 67 passed
- `bun run typecheck`
- `bun run build`
- `bun run test:full` — 1,516 passed, 1 skipped, 0 failed


krajla:
#248 This is somewhat related, the agent know will have knowledge of ongoing runs during conversations.

radutanasa (approved):

PR #275, opened 2026-07-30
feat(harness,cli)!: thread lifecycle, analysis purge, and the CLI flows that reach them

### LINKED ISSUE #234: Thread lifecycle: archive (soft delete), hard delete, and analysis purge
**Parent:** #224

## Goal

Give threads a real lifecycle — **archive** (recoverable, soft) vs **delete** (hard, disk-reclaiming) — and give the harness the **analysis purge** capability that every delete flow currently lacks.

## Today (verified)

- `cortex_analysis_threads.deleted_at` + `ThreadStore.deleteThread` already implement a soft delete — with **zero production callers**, and it never touches the thread's `messages` rows.
- The harness has **no analysis-delete path at all**. Analysis deletion is embedder-only (CLI SQLite), so both `analysis.delete` in the TUI and `inflexa prune` (the dead-anchor sweep in `modules/anchor/backstop.ts`) orphan the analysis's entire PG footprint forever: threads + messages, `cortex_analysis_state`, `cortex_artifacts`, runs/step executions, plans, working memory, asks, and the per-analysis `search_{slug}` vector table.

## Scope

- **`ThreadStore`, respecced**: `archiveThread` (soft — today's tombstone semantics under the honest name; hidden from default listing, recoverable) and `deleteThread` (hard — the row **and** its `messages`). Spec delta to `harness-thread-store`.
- **`purgeAnalysis(analysisId)`** — a harness-owned hard delete across every analysis-keyed table, including threads/messages and the dynamic vector index.
- **CLI wiring**: session archive + delete commands (palette); `analysis.delete` calls `purgeAnalysis`; `prune` calls `purgeAnalysis` per dead anchor via a lightweight Postgres-ready gate + pool (no full harness boot — prune is headless).
- **Semantics**: archive hides from default listing and is reversible; delete is final. Define how each interacts with in-flight state (an active chat on the thread being archived/deleted).

## Open questions

- `prune` when PG is unreachable: warn-and-proceed SQLite-only (leaves orphans until a later purge) or refuse until the stack is up.
- Whether unarchive ships now or later (the store method is trivial; the UX may not be needed yet).

## Out of scope

- Workspace file disposal (already handled by the analysis-delete dialog).
- Parent/child cascade rules — defined in the data-model sibling issue against these verbs.

vivere-dally:
## What & why

Gives threads a real lifecycle, gives the harness the analysis-purge capability every delete flow lacked, and wires both into the CLI so a user can actually reach them.

**Thread verbs.** `ThreadStore` had one removal verb, `deleteThread`, which wrote a `deleted_at` tombstone and never touched the thread's messages — correct archive behaviour under a name that promised reclamation. It becomes four verbs whose names match their guarantees:

| verb | guarantee |
|-|-|
| `archiveThread` | soft, recoverable; row and messages stay, thread leaves `getThread`/`listThreads` |
| `unarchiveThread` | clears the tombstone; the thread reads as it did |
| `purgeThread` | hard; removes the row **and** its `messages`, in one transaction |
| `listThreads({includeArchived})` | the only way to obtain an archived thread's id |

That last one exists because without it `unarchiveThread` could not be called: archiving removes a thread from every listing, so the recovery verb took an argument nothing would hand back. The archive was a one-way door wearing a recoverable name.

**`purgeAnalysis`.** Analysis deletion was embedder-only, so every delete flow orphaned the analysis's entire Postgres footprint. Measured on a live database, the proportions decide the design: analysis runs and their child step workflows hold **16 MB across 7098 step-output rows**, against roughly **1.8 MB** for every analysis-keyed `cortex_*` table plus the vector index combined. A purge scoped to the `cortex_*` tables would reclaim about a tenth of an analysis while calling itself final — and would destroy the id mapping needed to ever reclaim the rest, since `cortex_runs.run_id` *is* the parent workflow id.

So the purge covers the durability-engine ledger too, reached through a new **`WorkflowPurger` seam** that keeps the engine out of the state layer (the reclaim direction of the quarantine `RunLauncher` already holds for launching). Its realization is built on `DBOSClient` rather than the static facade, because the facade requires a launched engine and a headless embedder — the one that needs a purge most — has none.

Stage order is chosen for what a failure leaves behind: validate the analysis id → capture workflow ids → cancel → delete ledger rows → delete `cortex_*` rows in one transaction → drop the vector table. Ids are captured before the rows recording them are deleted; ledger rows go before the ledger that names them, so a failure leaves the mapping retryable. Across stages the unit of recovery is idempotent retry, not atomicity.

What the purge **cannot** reach is stated rather than left to look covered: scheduled operational workflows (unreachable from any analysis, and a retention concern of their own), target assessments, `messages` whose thread row is already gone, the shared regulatory corpus, and workspace files on disk.

**The CLI half — where the capability becomes real.** Two paths were creating orphans, and both are closed:

- **`analysis.delete`** now runs: flush and export provenance → dispose the workspace → purge Postgres → delete the SQLite row. The row goes **last** because it carries the only copy of the analysis id and the purge needs it; deleting it first strands the whole footprint beyond any retry *while reporting success*. A purge failure aborts with the row intact, so the deletion stays retryable rather than becoming a leak.
- **`inflexa prune`** did the same thing in bulk for every dead anchor's analyses. It now purges each one before any SQLite delete. Prune boots no runtime, so it takes Postgres through the provisioning gate and **starts the stack when it is down** — a stopped stack is a common reason to run a maintenance command, not a reason to block it.

The purge runs on **both** file-disposal modes. The "keep the files?" question has only ever governed the workspace tree — both modes already deleted the SQLite row and the signed provenance chain with it — and keeping the files is the default, so purging only on permanent deletion would have left the ordinary path orphaning exactly as before. The dialog's copy no longer claims otherwise.

On the keep-the-files branch, provenance is flushed and exported into the tree the user is keeping. It runs **before** the disposal and is skipped when no workspace exists: the export writes into the analysis's output directory, so afterwards it would recreate the very directory the disposal exists to clear, and a deletion must not conjure the tree it was asked to retire. A failed export is reported but does not abort — the user asked for a deletion, not an export.

Session commands now match the verbs: **Remove** archives, **Restore** brings one back, **Delete** erases. Only Delete spends the type-the-name danger ritual, because it is the first thread action that cannot be undone.

Closes #234

## ⚠️ Breaking changes

**1. `ThreadStore.deleteThread` is removed.** The hard delete ships as `purgeThread`; the archive behaviour ships as `archiveThread`.

The name is retired rather than repointed on purpose. Repointing would have left the behavioural inversion invisible to the compiler — identical signature, so a consumer upgrades into silent transcript loss on a call site nobody edited, guarded only by someone reading a changelog. Removing it means a stale call site fails to build:

```
error TS2339: Property 'deleteThread' does not exist on type 'ThreadStore'.
```

**2. `Thread` gains a required `deletedAt: Date | null`.** A row is always live or archived, so there is no third state for `undefined` to mean, and an optional field would let a host list archived threads and render them as live — the one mistake the widened listing exists to prevent. Consumers that only read a `Thread` are unaffected; constructors (in practice, test fixtures) must add the field.

**3. Deleting an analysis now removes its conversations and run history**, on both file-disposal modes, and `inflexa prune` does the same for every analysis it prunes. Previously that data survived indefinitely, unreachable.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally — harness `tsc` clean and its full suite green (1690 pass / 0 fail); cli typecheck and lint clean, and every touched test file passes individually. Note: the cli full-suite run is deliberately left to CI — locally it peaks near 9.5 GB, tracked in #282.
- [x] Documentation updated if user-facing behavior changed — three OpenSpec changes carry the proposals, designs, and spec deltas.

## Review notes

An adversarial spec audit and two independent code reviews ran against this branch; their findings are fixed here rather than deferred. The ones worth knowing about:

- The identifier guard validates the **analysis id**, not just its derived table name, and the requirement states both jobs it serves. The id shape is what keeps the `dataprofile:{analysisId}:` namespace lookup unambiguous — that lookup is a prefix match, so an id carrying the delimiter would let a purge of `a` sweep `a:x`'s workflows.
- A purge is **not** serialized against work still starting. A run inserted after the id capture is unreachable afterwards, so the contract states the precondition — quiesce the analysis first — and the CLI enforces it with the existing busy gate.
- An absent engine schema (nothing to purge) is distinguished from a missing workflow table beneath an existing one (a failure), so a half-migrated ledger can never be reported as a purge that reclaimed nothing.
- `cortex_plans` is removed by its own statement rather than by the cascade from `cortex_analysis_state`: `CREATE TABLE IF NOT EXISTS` adds no constraint to a table that already exists, so a database whose `cortex_plans` predates that key never acquires one.
- Exporting provenance before the disposal originally made a delete **create** a directory for a never-opened analysis, flipping the disposal from `absent` to `archived`. The export is now skipped when no workspace exists, guarded at the call site — the palette's own export command should still create the directory on demand.

**Verified against a real database.** `purgeAnalysis` was run over an analysis the CLI had already deleted, whose Postgres footprint survived: 1 run, 3 step executions, 14 artifacts, 1 thread, 13 messages, 4 workflows, 1216 `operation_outputs` (3432 kB) and a 208 kB vector index. After the purge every counter reads 0 and the index is absent; two live analyses in the same database are byte-for-byte unchanged, and a second purge of the same id succeeds reporting zeroes.

Four properties remain believed rather than tested and are recorded in `design.md` rather than left to look covered — chief among them that "no rows reappear after a cancel" needs a live executor to prove.


vivere-dally:
## ⚠️ Breaking change in this PR: `ThreadStore.deleteThread` now destroys data

@krajla flagging this for you directly — it is the silent kind of break, and the migration lands in the CLI rather than here.

### What changed

`deleteThread` kept its name and changed what it does. Before (`main`):

```sql
UPDATE cortex_analysis_threads
   SET deleted_at = NOW()
 WHERE thread_id = $1 AND deleted_at IS NULL
```

After (9cb45da1) — two statements in one transaction:

```sql
DELETE FROM messages WHERE thread_id = $1
DELETE FROM cortex_analysis_threads WHERE thread_id = $1
```

The old tombstone behaviour is preserved verbatim under the new name **`archiveThread`**, alongside a new `unarchiveThread`.

### Why it needs a callout rather than a changelog line

**The signature is unchanged** — `deleteThread(threadId: string): ResultAsync<void, DbError>` before and after. So there is no compile error, no failing type test, and the call site reads exactly as it did. What changed is that the call now irreversibly erases a user's conversation transcript where it previously kept every byte. An upgrade that doesn't read the commit body gets silent data loss on a path nobody touched.

That's also why it breaks in place instead of arriving as a new differently-named method: a new name would leave a soft, misnamed `deleteThread` in the public surface indefinitely, and the misnaming *is* the defect being fixed. Every existing caller has a faithful one-word migration.

### Who is affected — exactly one consumer

`cli/src/tui/commands.tsx:159`:

```ts
deleteThread: (pool, threadId) => createThreadStore(pool).deleteThread(threadId),
```

That feeds the palette's session-remove flow, whose confirmation dialog promises the user (`commands.tsx:1205`):

> It stops appearing in this analysis. The transcript is kept — nothing is erased.

and whose success notice says (`commands.tsx:1228`):

> Session removed — it no longer appears in this analysis.

**Both statements become false** the moment that call resolves to the new implementation.

**Migration:** point line 159 at `archiveThread`. That restores exactly the behaviour the copy describes. Three seam JSDoc lines and two test comments describe it as a tombstone and would then document the opposite of the code — `commands.tsx:142`, `commands.tsx:1165`, `session_remove_dialog.render.test.tsx:16`.

### Nothing is broken right now

`cli/package.json:25` pins `"@inflexa-ai/harness": "0.13.0"` — a registry version, not `file:../harness` — so the CLI resolves the published harness and is unaffected by this branch. The break becomes real only at the commit that bumps that pin, which is where the `archiveThread` swap has to land in the same change. That's why the CLI wiring is a paired follow-on rather than part of this PR.

### One caveat for whoever wires the CLI's hard-delete verb

The hard delete is **not serialized against a concurrent `appendTurn`**. `messages` carries no foreign key to `cortex_analysis_threads`, and the append deliberately tolerates a missing metadata row (a specified scenario in `harness-thread-history`, which this PR does not change). So a turn committing after the delete leaves messages under a `thread_id` that resolves to nothing — unattributable to any analysis and unreachable by any later purge.

The store guarantees only that it creates no orphan *of its own*; stopping writes to a thread first is the host's obligation, and it's stated in the delta spec. The CLI already does the right thing on the soft path — it unbinds the thread to `null` before re-landing (`commands.tsx:1239`) — so the discipline exists and just has to be deliberate for the new verb.

vivere-dally:
@claude review

vivere-dally:
Both findings confirmed against the code, and both will be fixed on this branch. Verification of the mechanism, since the first one is a data-loss path and worth pinning precisely:

**1. `purgeSessionFlow` quiescence gate — confirmed, fixing.**

The palette chord is declared at `MODE_BASE` with no chat-status gate (`app.tsx:741-744`); only an open dialog suspends that layer, so ctrl+k is live during a streaming turn. `session.purge`'s `enabled` predicate (`commands.tsx:2063`) checks analysis, session and boot phase — nothing about in-flight work.

The orphan mechanism is exactly as described: `appendTurn` inserts into `messages` unconditionally (`thread-history.ts:295-302`), and the `cortex_analysis_threads` touch that follows is best-effort behind a savepoint (`:340-349`) — deliberately, so a missing or tombstoned thread row never fails the turn. There is no FK and no existence check, so a turn committing after `purgeThread` writes rows under a `thread_id` that resolves to no analysis.

One correction to the suggested fix: `workspaceBusyReason` is the wrong gate to reuse. It also refuses on an in-flight data profile and on unterminated runs, neither of which has any bearing on a conversation's messages — a user should not be blocked from deleting a chat because a profile is running. `chatStatus() === "busy"` is already thread-scoped by construction here, because `purgeSessionFlow` only ever purges `ctx.sessionId`, the open thread. So the gate is that single check, refusing with a notice rather than hiding the command (leader chords dispatch by id and bypass `enabled`, which is why `openRestoreSession` speaks rather than no-ops).

`deleteSessionFlow` (Remove) is deliberately **not** getting the same gate. Archiving mid-stream leaves the turn's messages on a tombstoned thread, and the touch's `WHERE deleted_at IS NULL` simply matches nothing — Restore brings the thread back with those messages intact. It is recoverable, so it does not warrant blocking the action.

**2. `design.md` Non-Goals stale — confirmed, fixing.**

Correct, and there is a second stale line right below the one you flagged. `design.md:22` claims "Serialising the purge against work still starting … the existing quiescence gate already enforces it here" — true of `analysis.delete`, false of `session.purge`, which is precisely finding 1. Both lines get reconciled.

**Not addressing, and why:**

- **`countDoomed`'s duplicated descendant walk.** The recursive CTE walks `parent_workflow_id` and the engine's `deleteWorkflows` walks it again. That is the deliberate price of reporting a count the engine does not return, and it is documented as such at `dbos-workflow-purger.ts:137-143`.
- **`schemaConfirmed` memoization.** Both CLI call sites build a fresh purger per operation, so the positive memo never survives a call. It costs nothing — the probe only runs on the failure path — and the one-directional caching is correct as written.

Separately, three findings from our own review are being fixed in the same pass: the restore picker can render "No archived conversations" while archived threads exist (the widened listing takes the store's default 100-row page ordered by `updated_at DESC`, and archiving leaves `updated_at` untouched, so archived rows sort behind live ones and fall off page 0); a purge failure after a successful archive reports "nothing was lost" without naming where the files went; and `PruneError`'s JSDoc claims every failure leaves all rows in SQLite, which the final stage's eagerly-evaluated `Result.combine` does not guarantee.

krajla:
This is ok, managed service doesn't expose deleting threads. And archiving was already connected.

krajla:
No silent breaking changes 😄

krajla (approved):

PR #277, opened 2026-07-30
Fix planner terminal submission flow

krajla:
## What changed

- Remove the separate validate_plan tool and make submit_plan return structured schema and semantic issues for correction.
- Pin the internal planner to the conversation provider/model pair.
- Require tool-driven planner completion and stop immediately after a terminal outcome.
- Avoid unnecessary salvage turns after a successful terminal submission.

## Why

Planner runs could spend turns validating separately, continue after recording an outcome, or hit timeout/iteration limits while recovering from provider/tool behavior. The new flow submits, corrects from returned issues, and terminates as soon as the outcome is recorded.

## Validation

- bun run build (harness)
- 42 focused harness tests passing (run-agent, run-to-terminal, conversation-agent)
- git diff --check

The database-backed planner tests require a container runtime, which is unavailable in this environment.

### DIFF
--- a/harness/src/agents/conversation-agent.ts
+++ b/harness/src/agents/conversation-agent.ts
@@ -237,7 +237,13 @@ export function createConversationAgent(deps: ConversationAgentDeps): AgentDefin
         createInspectRunTool(pool),
         // The dataset's own record. No file backs it — the DB row is the only copy.
         createInspectDataProfileTool(pool),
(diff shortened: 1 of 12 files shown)


vivere-dally (dismissed):
The tests are failing, check pls

vivere-dally (approved):

Issue #281, opened 2026-07-30
Tool outcomes render `✓ ok` when the tool did nothing — surface the ok-channel result variants


vivere-dally:
## Problem

A tool chip's status answers one question: *did the loop wrap this call as an error `tool_result`?* It does not answer *did the tool do what the agent asked?*

The tool error contract (`harness/src/tools/define-tool.ts:12-16`) sends every **expected** outcome down the ok channel as a data variant. About 40 sites do this. A sample:

| Tool | Outcome that renders `✓ ok` | Site |
|-|-|-|
| `grep` | `no_matches`, `not_found`, `out_of_scope` | `harness/src/tools/workspace/grep.ts:83-158` |
| `edit_file` | `file_not_found`, `not_found`, `not_unique` | `harness/src/tools/workspace/edit-file.ts:114-120` |
| `read_file` | `invalid_input` | `harness/src/tools/workspace/read-file.ts:100` |
| `list_files` / `file_stat` | `not_found`, `out_of_scope` | `list-files.ts:44`, `file-stat.ts:40` |
| `write_file` (report) | `out_of_scope`, `too_large` | `harness/src/tools/report/version-fs.ts:75-86` |
| `run_inflexa` | `invalid`, `blocked`, `timed_out`, `cancelled` | `cli/src/modules/harness/inflexa_tool.ts:137-141` |
| bio tools | `found: false` | `harness/src/tools/bio/comptox.ts:157`, `research/context7-docs.ts:50` |

Every one of these renders `✓ ok`. A `grep` that matched nothing and a `grep` that read 400 files look identical in the transcript. The `✗` state is almost never seen, so the chip's status carries far less information than it appears to.

## Why the tool must classify its own result

`isErrorOutput` (`harness/src/loop/run-agent.ts:470-472`) is a transport predicate — it reads `output.type` on the `ToolResultPart`. It is correct for what it does and cannot be extended to cover the table above, because those outcomes are indistinguishable from success at the transport layer by design.

Only the tool knows whether its own `ok` was a success. That is the same argument #174 makes for the input side.

## Approach idea

`describeResult(output) => …`, the result-side twin of the `describeCall` hook from #174 — colocated with the tool, typechecked against its exact output type, computed best-effort at dispatch, stamped onto `tool-finished`.

#174 defers this deliberately, and that deferral was right for the reported pain there (four indistinguishable `update_memory` chips). This issue is the case the deferral did not cover: the outcome, not the call.

[...]

PR #273, opened 2026-07-29
Persist versioned conversation display history

krajla:
## Summary

- Persist complete, versioned AI SDK UI-message projections alongside exact model-message history.
- Add a harness-owned recorder for provider deltas, tool lifecycle, durable data parts, and approvals.
- Backfill legacy turns at startup and deprecate tool-name-based card reconstruction.
- Publish the additive harness API as `@inflexa-ai/harness@0.14.0`.

## Why

Historical conversation rendering was reconstructed from tool names, tool inputs, and mutable workspace/database state. Tool renames or resource changes could therefore alter or lose old display content. The stored display envelope makes what was shown the durable historical fact while keeping model history and DBOS streams independent.

## Scope

This PR contains the harness capability only. The CLI consumer change is intentionally separate so it can pin the published harness version.

## Validation

- CI lint/typecheck: CLI and harness pass
- CI full tests: CLI and harness pass
- Harness targeted storage tests: 7 passed
- Harness build and package smoke test pass
- Strict harness OpenSpec validation: 66/66

All commits carry DCO sign-offs.

PR #287, opened 2026-07-31
feat(cli,harness): let tools self-describe a one-line call detail

### LINKED ISSUE #174: Tool calls render as bare chips — let tools self-describe with a one-line call detail
## Problem

The TUI renders live tool calls as name + status + duration only. When the conversation agent updates three memory fields and removes one, the transcript shows four indistinguishable `update_memory` chips — no way to tell what happened. Same for `run_inflexa` (which arguments?), `write_file` (which file?), and every other tool.

## Why not "just show the input"

Forwarding the raw tool input and formatting it per-tool in the UI puts the interpretation across the wire from the schema it interprets:

- The input schema is typechecked in the harness; a UI formatter reading it is string-keyed guesswork that silently breaks when the schema evolves.
- Every embedder (CLI today, a managed deployment tomorrow) would duplicate the same per-tool formatters — against the harness-first boundary rule.
- Inputs can be huge (`write_file` content); shipping and persisting payloads to display one line is waste.

## Approach idea

Let the tool self-describe **at call time**, mirroring how its `description` already self-describes at attach time:

- `defineTool` gains an optional hook, e.g. `describeCall(input) => string`, colocated with the zod input schema — typechecked against the exact input type, written by the tool author who knows which field matters. `update_memory` → `hypotheses +2 −1`; `run_inflexa` → the argv string.
- The loop computes it at dispatch (best-effort: a throwing hook never breaks the call) and stamps it onto `tool-started`/`tool-finished` and the wire `ToolCallPart` as `detail?: string`. The emit site — not each tool — enforces single-line + a length cap.
- Hosts render name + detail generically with zero tool-specific knowledge. (The CLI's `ToolCallPart` already carries a fixture-only `target` field its tool block renders — live details map straight onto it.)
- **Reload parity**: reloaded threads rebuild from the stored AI SDK transcript, which holds the raw input but no detail. Reuse the `resolveCard` reconstruction pattern — one registry-backed `toolName + input → detail` resolver shared by the live loop and the converter — so storage stays the pure model transcript and details survive reload.
- Tools without the hook render exactly as today (sub-agent tools, future tools degrade gracefully). Coverage grows tool-by-tool; first candidates: `update_memory`, `run_inflexa`, `write_file`, `workspace_search`.

This slots into the existing display tiers — generic chip → chip + detail → promoted card (`resolveCard`). When one line stops being enough for a tool, the answer is promotion to a dedicated card, not a fatter detail string.

## Out of scope (natural follow-ups)

- Result-side summaries (`describeResult(output)` → "3 matches", "exit 0") — input-side covers the reported pain, and `isError` already rides the finished event.
- Structured detail (`{verb, target}`) — start with one string; the contract is harness-owned and can widen when a renderer actually needs structure.

vivere-dally:
## What & why

The TUI renders a live tool call as a name, a status, and a duration. When the conversation agent writes three working-memory entries and retires one, the transcript shows four identical `update_working_memory` chips. Same for `run_inflexa` (which argv?) and `workspace_search` (which query?).

Forwarding the raw input and formatting it host-side was rejected: it puts the interpretation across a boundary from the schema it reads, so it is string-keyed guesswork the compiler cannot check, and every embedder ends up writing the same per-tool formatters.

**`defineTool` gains an optional `describeCall(input) => string`**, colocated with the Zod `inputSchema` and typechecked against the exact input type. The loop computes it at dispatch, best-effort — a hook that throws never breaks the call — and stamps it onto `tool-started`/`tool-finished`. The **emit site**, not each tool, enforces single-line, control-character stripping, secret redaction, and a 120-character cap. Tools without the hook render exactly as they do today.

A registry-backed resolver over a supplied `readonly Tool[]` rebuilds the detail on reload, so storage stays the pure model transcript. The list is supplied rather than held inside the harness because `run_inflexa` and friends are contributed by the embedder through the `hostTools` seam and a harness-internal map could never see them.

The hook also **subsumes `activityForTool`**, which carried the same defect in a worse form: a hand-maintained name→label table reading input fields through a `Record<string, unknown>`, one file away from the schemas it interpreted.

### Two adjacent fixes, in the same code

- **A reloaded failed call rendered `✓ ok`.** The failure was always persisted as an `error-text` output on the `tool-result` block; the converter dropped that block as unrenderable, discarding the only record of it. It now pairs each stored call with its result by id and recovers the outcome. No new storage, no migration.
- **A refused approval rendered `✗ error`.** `tool-finished` now reports `outcome: "ok" | "error" | "denied"` instead of a boolean. The loop already treats a denial as distinct from a recoverable error in its control flow; folding both into one flag lost that at the observation boundary, so a user saw their own decision reported as a fault.

### Rendering

The detail sits on the name line when it fits the measured width and reflows to one indented row when it does not. It is never truncated — a workspace path's tail names the file, and the detail is one opaque harness string the block may not parse, so it cannot elide the middle intelligently.

```
▸ grep src/**/*.ts  ▸ running
▸ read_file src/db/types.ts :55-105  ✓ ok · 1.2s
▸ write_file out/report.html  ✗ error · 320ms
▸ execute_analysis plan_01J8F2QK  ⚠ denied · 4ms
▸ read_file  ✓ ok · 14ms
  runs/2026-07-30T09-14-22Z/step-4-pathway-enrichment-and-gsea/output/hallmark_gsea_ranked_by_normalized_enrichment.csv
▸ search_semantic_scholar  ✓ ok · 890ms
```

On the reflow the status stays with the name, which puts it in a near-constant column across split blocks. `denied` uses `GLYPHS.warning` in the `warning` role — a soft state, not a fault; its contrast is measured on a light and a dark theme, since the existing block sweep covers only `ok` and `running` and a wrong role still scores well on dark.

### Also fixed along the way

`paddingLeft` on an opentui `<text>` is a no-op. The sub-agent activity row had been rendering flush against the marker gutter since it shipped, and no assertion caught it because the characters are identical either way. Both subordinate rows now carry their indent on a wrapping `<box>`, with a test that pins it.

### Breaking

`tool-finished` no longer carries `isError`. Consumers read `outcome`. Both in-repo consumers are migrated in this PR.

### Not in scope

Result-side summaries (`describeResult`) and a `warning` status tier for ok-channel outcomes such as `no_matches` and `blocked` — tracked in #281, which depends on this.

Closes #174

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally. `harness`: `tsc -p tsconfig.json` clean, `bun run test:full` → 1717 pass / 1 skip / 0 fail. `cli`: `bun run typecheck` and `bun run lint` clean, `bun run test` → 2376 pass / 1 skip / 0 fail. (`cli` needs `bun run test`, not bare `bun test` — the `--isolate` requirement in `cli/CLAUDE.md`.)
- [x] Documentation updated if user-facing behavior changed. `harness/CONTEXT.md`; the CLI's design gallery gains the reflow, denied, and hookless exhibits, which is its equivalent surface.

Design rationale lives in the two OpenSpec changes rather than here: `harness/openspec/changes/add-tool-call-detail/` (8 decisions with alternatives) and `cli/openspec/changes/render-tool-call-detail/` (7). Both pass `openspec validate --strict`.

Methods and provenance are untouched. The sandbox line is answered because files under `harness/src/sandbox/` and `harness/src/tools/workspace/` appear in the diff, and a reviewer scanning for that should get an answer rather than a deleted section.

- [x] **Sandbox:** the security model is preserved. Every sandbox-adjacent change here is display-only — `activityForTool` now resolves its phrase through the tool's own hook instead of a name table, and `sandbox-step.ts` threads that resolver in. Nothing touches the exec protocol, HMAC signing, the egress firewall, capabilities, mounts, or `allowedWritePrefix`. `execute_command`'s hook reads only `command` and deliberately never `env`, which carries secrets by design.

One behaviour change worth a reviewer's eye: with the name table deleted, a sandbox step's live activity line reads `write_file scripts/run.py` rather than the old hand-written `Writing file run.py`. The name leads deliberately — that phrase renders with no tool name beside it, so a bare detail would make a write and an edit of one path identical. The prose was nicer; it was also the unchecked mapping being removed, and it cannot be kept without keeping the table.


vivere-dally:
@claude review

vivere-dally:
@krajla — one more breaking change worth adding to the **Breaking** section, which currently lists only `tool-finished` dropping `isError`.

`contentToCortexMessages`'s second parameter changed shape:

```ts
// before
contentToCortexMessages(messages, resolveCard)
// after
contentToCortexMessages(messages, { resolveCard, resolveDetail })
```

It is re-exported from the barrel (`harness/src/index.ts:158`), so it is part of the published surface, not internal wiring — any embedder passing a bare `ToolCardResolver` has to migrate. Both in-repo call sites are updated in this PR, and the change fails at compile time rather than silently, so it is low-risk. It just belongs in the list so a consumer reading the release notes sees both.

The options bag itself is the right call — two optional function-typed parameters in positional form is exactly the call site that does the wrong thing when they are swapped.

krajla:
The breaking changes are ok, the exposed functions are called inside the harness. I also made a PR #273 which will help greatly with code maintenance in how we store display data instead of re-creating it at read time. 
I ask you to rebase this branch onto that one and fix any conflicts.

krajla (approved):

PR #288, opened 2026-07-31
fix(harness,sandbox): drop lineage reads of paths that were never there

vivere-dally:
## What & why

A customer's `bulk-transcriptomics-agent` step ran for twelve minutes, finished its work, and was then killed by its own bookkeeping:

```
[reconcile-manifest] cannot attest input — not present at reconcile
  path:      /{rid}/runs/{run}/T1S1/scripts/hashtable_class_helper.pxi
  source:    upstream   refStepId: T1S1
  throwSite: input-enoent
```

`hashtable_class_helper.pxi` is a **pandas Cython source file**. No such file has ever existed in that directory, and the step never tried to read one. Three links produce it:

1. **A script under the upstream step's `scripts/` directory** puts that directory on `sys.path` — inside the analysis mount, and inside a *declared dependency's* subtree.
2. **The script died with an uncaught pandas exception.** A Cython frame's `co_filename` is *relative* (the name in pandas' `include` directive), and CPython's `_Py_FindSourceFile` resolves a relative frame filename by **opening** `<entry>/<basename>` for every `sys.path` entry until one succeeds. All of them ENOENT.
3. **`sitecustomize.py` hooks the `open` audit event**, which CPython raises *before* the open is attempted (`sitecustomize.py:88`), so every failed probe is reported exactly like a real read and `abspath`'d against cwd (`:51`).

Reproduced against the shipped hook, unmodified — a script whose only fault is `df["missing_column"]` emits:

```
{"p": ".../runs/RUN/T1S1/scripts/hashtable_class_helper.pxi", "layer": "python", "op": "read"}
{"p": ".../runs/RUN/T1S1/scripts/index.pyx",                  "layer": "python", "op": "read"}
```

The path classifies `upstream`, which is content-attested, so `fillInputHashesFromDisk` stat'd it, got ENOENT, and threw — taking the run with it through the parent's fail-fast cascade, and *masking* the script error the agent had already recovered from.

### Why the throw was wrong

This is the third instance of one shape. Both prior fixes (`/{resourceId}/..`, out-of-mount reads) explicitly retained the ENOENT throw as "genuine drift". **That premise is false: the capture layers report *attempted* operations, not completed ones** — the Python audit hook fires ahead of the open, R's `trace()` at call entry over a `normalizePath(mustWork = FALSE)` name (`Rprofile.site:32`). Only LD_PRELOAD gets this right, gated on `fd >= 0` (`provtrack.c:225`). An absent path is therefore at least as likely to mean *nothing was read* as *an artifact vanished*, and neither is attestable. Nothing about failing recovers the edge: the step has already run and its outputs are already on disk.

### The change

- **Harness** (`reconcile-manifest.ts:201`) — an input absent at reconcile is **dropped** from lineage (warn naming the ref, its resolved `hostPath`, and `dropSite: "input-enoent"`; counted with `reason: "missing"`), joining the directory and out-of-tree drops. Dropping upholds "never register a hashless lineage edge" exactly as the throw did. Fail-fast is retained for a `stat` that fails any **other** way — the file is there and cannot be read, which says something is wrong with the tree itself.
- **Sandbox** (`provenance.go:299`) — `recordOp`, where all four layers converge, drops a `read` report whose path is not present, so a probe never reaches the frame. Only absence drops it; a `stat` failing any other way keeps the report rather than losing a real edge to a transient error. Writes and deletes are exempt, since a write is reported before the file it creates exists.
- **Specs** — deltas on `artifact-manifest`, `sandbox-provenance-tracking`, and `exec-provenance-lineage`, archived into the main specs, plus the **Purpose** sections deltas do not carry (two still claimed an unhashable input is terminal).

The two fixes are independent by design: the harness one ships via npm and holds on any sandbox image, which matters because the image is `workflow_dispatch`-only and `:latest`-tagged.

### Shipping

- The sandbox half reaches a host only when it re-pulls — `lib-store.yml` still needs a dispatch (the one unchecked task in the archived change).
- The harness half needs a harness release (`0.14.2` → next) and a cli dep bump; `cli/package.json:25` pins the published version.

### Not in scope

- The inotify `IN_OPEN`-as-read classification (carried over from the prior change).
- The `stat`→hash TOCTOU: same shape, but the sandbox is destroyed before reconcile so there is no writer, and it cannot be failed deterministically in a test — untested defensive code was the worse trade.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally. Harness suite **1349 pass** against a **1348 baseline on the unchanged tree**; the 181 failures are pre-existing Postgres-testcontainer failures on this machine, identical before and after. `tsc` clean, eslint clean on changed files, `gofmt`/`go vet` clean, Go tests green, 68 openspec items validate `--strict`.
- [x] Tests added or updated for the change. 4 new/rewritten harness tests, including the customer's exact shape end-to-end through `feedExecFrame` — **verified all 4 fail against the old throw**. 3 new Go tests — **verified the drop test fails without the fix**. The existing watch-dir tests now inject a presence probe so they keep testing the bound rather than passing on absence.
- [x] Documentation updated if user-facing behavior changed (specs; no user-facing surface changes).

- [x] **Methods:** n/a — no analytical method changes.
- [x] **Sandbox:** the security model is unchanged. No change to the workload uid, capabilities, `no-new-privileges`, egress policy, exec authentication, mount read-only-ness, write confinement, or resource limits. The sandbox-side change is one `stat` inside `recordOp`, on a path already bounded to a watch dir, and it only ever *removes* a report from the frame.
- [x] **Provenance:** prior analyses remain reproducible, and the recorded graph gets strictly more accurate — the only edges removed are ones naming files that do not exist, which could never have been content-attested. Every drop is observable: a warn record with the ref and `dropSite` host-side, a debug line sandbox-side, and `cortex.artifact.reconcile.input_dropped{reason="missing"}`.

### DIFF
--- a/harness/openspec/changes/archive/2026-07-31-drop-phantom-lineage-reads/.openspec.yaml
+++ b/harness/openspec/changes/archive/2026-07-31-drop-phantom-lineage-reads/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-31
--- a/harness/openspec/changes/archive/2026-07-31-drop-phantom-lineage-reads/proposal.md
+++ b/harness/openspec/changes/archive/2026-07-31-drop-phantom-lineage-reads/proposal.md
@@ -1,0 +1,87 @@
+## Why
+
+A customer's `bulk-transcriptomics-agent` step ran for twelve minutes, finished
+its work, and was then killed by its own bookkeeping:
+
+```
+[reconcile-manifest] cannot attest input — not present at reconcile
+  path:      /{rid}/runs/{run}/T1S1/scripts/hashtable_class_helper.pxi
+  source:    upstream   refStepId: T1S1
+  hostPath:  .../analyses/stroke/runs/{run}/T1S1/scripts/hashtable_class_helper.pxi
+  throwSite: input-enoent
+```
+
+`hashtable_class_helper.pxi` is a pandas Cython source file. No such file has
+ever existed in that directory, and the step never tried to read one. What
+happened is CPython's traceback printer, and it is reproducible in three lines:
+
+- A script the step ran lives in `T1S1/scripts/`, so `sys.path[0]` is that
+  directory — inside the analysis mount, and inside a *declared dependency's*
+  subtree.
+- The script died with an uncaught `KeyError` from pandas. Displaying that
+  traceback means showing the source line of every frame, and a Cython frame's
+  `co_filename` is relative (`hashtable_class_helper.pxi` — the name in the
+  `include` directive). CPython's `_Py_FindSourceFile` resolves a relative frame
+  filename by *opening* `<entry>/<basename>` for **every** `sys.path` entry until
+  one succeeds.
+- `sitecustomize.py`'s audit hook fires on the `open` **audit event**, which
+  CPython raises *before* the open is attempted. Every failed probe is therefore
+  reported exactly like a successful read, `os.path.abspath`-ed against the
+  process cwd.
+
+Verified against the shipped hook, unmodified — a script that only crashes with
+`df["missing_column"]` emits:
+
+```
+{"p": ".../runs/RUN/T1S1/scripts/hashtable_class_helper.pxi", "layer": "python", "op": "read"}
+{"p": ".../runs/RUN/T1S1/scripts/index.pyx",                  "layer": "python", "op": "read"}
+```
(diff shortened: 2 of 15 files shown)


vivere-dally:
@claude review

radutanasa (approved):

PR #291, opened 2026-08-03
feat(scripts): add a prototype of the two-container package store

vivere-dally:
## What & why

The sandbox image contains every package. The image is large. A user cannot add a package to it.

This change adds a prototype. The prototype divides the work between two containers.

- **The provisioner** has network access and compilers. It mounts only the package store. It writes the packages to disk. Then it stops.
- **The sandbox** does not change. It has no network access. It runs as uid 1000 and drops all capabilities. It reads only the files that the provisioner wrote.

The sandbox still has no network access. That constraint stays true.

The prototype is separate. No code in `cli` or `harness` uses it. Its purpose is to test the design before we write product code.

## How the store works

One directory on the host contains all packages. Each package version gets one directory. A hash of the content gives that directory its name. Ten analyses that use scanpy share one copy on disk.

Each analysis also gets a directory of symbolic links. The links point into the store. An analysis thus sees only the packages that it needs. The code calls this directory a farm.

## Why the harness does not change

`harness/src/sandbox/docker-client.ts:314` mounts a host directory read-only at `/mnt/libs`. `docker-client.ts:126` permits the mount when `current/` is a directory that contains the files `packages.txt` and `meta.json`.

The store and the farms go below that same mount. The links point to `/mnt/libs/store/…`. This path exists in both containers. The harness therefore needs no change to test the design.

## Test results

The tests run on linux/arm64. They use the `sandbox-base` image and 50 packages for scanpy. All 15 checks pass.

The checks include the conditions that are most likely to fail:

- Compiled C extensions load through the links.
- Vendored shared libraries resolve through `$ORIGIN`.
- `importlib.metadata` reports the correct versions.
- Two packages that share a top-level name merge correctly.
- The sandbox cannot open a network connection.
- The store and the farm are read-only.

| Measurement | Result |
|-|-|
| Sandbox with scanpy, this design | 3.1 GB |
| Sandbox with scanpy, the image we ship now | 11.4 GB |
| A second scanpy analysis | 8 MB |

## Findings

Two findings apply to the images that we ship now.

1. **numba cannot read its cache from a read-only store.** numba writes a test file to select a cache directory. A read-only directory fails this test. numba then compiles the code again. The `sandbox-python` image contains 25 cache files and uses none of them. It compiles 24 entries again at run time.
2. **An import does not fill the numba cache.** numba compiles a function at the first call. An import of numba, matplotlib and scanpy gives 0 cache files. A true scanpy workload gives 23 cache files.

Two findings apply to this design.

3. The cache preparation must use the `/mnt/libs/current` path. The numba cache key contains the path of the source file.
4. The numba cache key also contains the argument types. Prepare the cache with a workload that is similar to the analysis.

`scripts/store-prototype/README.md` gives the full details.

## Not covered

- conda. Its binaries contain the build path. The prototype does not test conda.
- The R packages and the Node packages.
- Two analyses at the same time. One pointer selects one farm.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). **This change adds no TypeScript.** It adds one shell script, two Python files and one Dockerfile, all below `scripts/`. I verified the shell syntax with `bash -n` and the Python syntax with `py_compile`.
- [x] Tests added or updated for the change. `acceptance.py` holds 15 checks. It runs inside the sandbox through `inflexa-store test`. It is not yet part of CI.
- [ ] Commits are **signed off** for the DCO (`git commit -s`). **The commit has no `Signed-off-by` trailer.** It needs an amend and a force push.

- [x] **Methods:** the prototype pins its base image by digest. It pins the `uv` version. It uses the same base image as `sandbox-base`, because the sandbox loads what the provisioner compiles. A different base gives a different ABI.
- [x] **Sandbox:** the security model does not change. This PR adds no code to any path that starts a sandbox. The prototype starts its own sandbox with the same posture: uid 1000, all capabilities dropped, `no-new-privileges`, and a read-only store. It uses `--network none`, which is stricter than the firewall inside the container. The provisioner is the one container with network access. It mounts only the package store, so it holds no user data.
- [x] **Provenance:** prior analyses stay reproducible. No shipped code path changes.


vivere-dally:
@claude review

harness/src/sandbox/create-sandbox.ts
```diff
@@ -70,6 +70,20 @@ export interface CreateSandboxClientConfig {
...
+     * keep their meaning. On Docker the location is an absolute host path; on
+     * Kubernetes it is a subPath under the store PVC. The backend resolves the
+     * provider at each `createSandbox` call, and only when a store is configured.
+     *
+     * A provider that names no farm refuses that one sandbox with the
+     * `farm_unavailable` state, and it affects no other surface. Unset keeps the
+     * single mount of the store root, thus an embedder migrates in its own
+     * release.
+     */
+    resolveAnalysisFarm?: ResolveAnalysisFarm;
```
vivere-dally: Does the managed solution need the farm concept?

vivere-dally: Yes for the concept, no for per-analysis composition. And managed keeps its behavior.

**Why the concept is host-agnostic.** The farm is the adapter between a content-addressed pool and an interpreter. The pool is `store/<name>-<version>-<hash>/`, and no Python or R interpreter imports from hash-named directories. Managed runs the same image, which bakes `/mnt/libs/current` into the `.pth` file (`images/sandbox-base/Dockerfile:412`) and into `R_LIBS_SITE`. Thus managed has the same need. If managed dropped the farm, `/mnt/libs/current` would mean "the one shared library set" there and "the closure of this analysis" under the CLI. One path with two meanings is what the boundary rule forbids.

**Managed does not suffer, and the compatibility is structural.** An embedder that sets no provider gets the identical Job spec:

- `k8s-client.ts:252` pushes the farm mount only `if (farm !== undefined && plan.farmPath)`.
- With no provider, `farmProvider` is `undefined` (`:525`), thus `resolvedFarm` is `ok(undefined)`.
- The refusal at `:542` is gated behind `farmProvider &&`, thus it never fires for an embedder that did not opt in.
- The store PVC mount at `:241` is untouched.
- `k8s-client.test.ts:511` pins it: the libs volume gives `["/mnt/libs"]` and nothing else.

The env did not move either. `FARM_CONTAINER_PATH` is the same `/mnt/libs/current` literal that the code carried before, thus `R_LIBS_SITE` and the baked `.pth` resolve as they did (`mount-plan.test.ts:56`).

**The seam is a function, thus managed chooses the policy.** It can return a distinct subPath for each analysis, or `farms/catalog` for every analysis, which is one environment-wide farm. It can also set no provider at all.

**One seam is real, and it is the store layout and not the harness.** The provisioner no longer writes `current`. A PVC that the published artifact populates carries `farms/catalog` and no pointer. Thus a managed deployment that sets no provider would mount `/mnt/libs/current` and find nothing. The single-mount fallback keeps the harness behavior, but the layout moved under it. The fix is one line: return `farms/catalog` from the provider for each analysis.

**On the API.** The seam is harder to read than it must be, because `undefined` carries two meanings: the field is unset means "the store root serves", and the return is undefined means "refuse this sandbox". A named union for the source would make the three policies explicit, and it would put the managed choice in the type rather than in a comment. I can raise that as a separate change.

vivere-dally: The API part landed in this PR rather than in a separate change — `7303ab75`.

`FarmSource` names the three deployments, thus the type answers "what does managed get" without a trace through the backend:

```ts
type FarmSource =
    | { kind: "store-root" }                                   // one mount, never a nested one
    | { kind: "fixed"; location: FarmLocation }                // one farm for every analysis
    | { kind: "per-analysis"; resolve: ResolveAnalysisFarm };  // a farm for each analysis
```

A managed deployment that mounts the published artifact names one line:

```ts
farmSource: { kind: "fixed", location: "farms/catalog" }
```

It composes nothing, and `k8s-client.test.ts` exercises it: two analyses, two sandboxes, one subPath.

Three other things came with it:

- **`undefined` no longer carries two meanings.** A provider gives `FarmResolution`, which is `{ kind: "farm", location }` or `{ kind: "unavailable", reason }`. An unset field and an absent return no longer share a spelling.
- **A refusal says why.** `farm_unavailable` carries the reason of the embedder, and `describeSandboxError` prints it. The CLI passes the failure of the composition into it.
- **One mount path for each backend.** `farmProviderOf` turns `fixed` into a provider that ignores its analysis id, thus managed and the CLI exercise the same mount code and not two branches.

An embedder that names no source keeps the single mount of the store root. The compatibility of a managed deployment is unchanged, and it is now visible in the type instead of in a comment.

vivere-dally:
## Open gap: the R track cannot be extended after the catalog build

Recorded so that it is not lost. It is out of scope for the current farm work, by decision.

### What holds today

`inflexa store add` acquires Python only:

- `runStoreAdd` fixes the ecosystem at the call site — `parseLibStoreFlightSpec(raw, "python")` (`cli/src/modules/libs/store.ts:792`).
- `provisionPackages` passes each spec as a bare positional argument (`store.ts:378`). The provisioner routes those through `reject_off_index` into uv (`images/sandbox-provisioner/provision.py:1682`).
- The CLI never passes `--r-manifest`.

### Why it is not a small addition

The provisioner refuses the invocation that one R package would take (`provision.py:1694-1695`):

> `--r-manifest needs --farm <name>; an R install farms each package it stores`

The R path takes a whole manifest and a farm, and it runs pak over the full track (`provision_r`). No acquire-one-R-package operation exists in the system. Three parts are necessary:

1. pak driven for one package, and not a lockfile install of the whole track.
2. The farm requirement satisfied or removed, because an acquisition run builds no farm.
3. A placement rule for the subtree. Today `r/cran`, `r/bioconductor`, and `r/github` come from a reachability sort over the whole install. Thus one package alone has no placement.

### What does work

A farm can still LINK any R package that the catalog holds. `DepsNode.track` carries `"r"`, an R node records its inner directory name, and the link pass writes the R subtree links (`cli/src/modules/libs/composition.ts:882-901`).

### The consequence to design for

An R request that the pool cannot answer must refuse with its own reason. That reason is: the catalog does not carry the package, and this store cannot acquire an R package. A generic not-found is wrong here. Without that distinction an agent retries a request that can never succeed.

vivere-dally:
Rebuilt in https://github.com/inflexa-ai/inflexa/pull/453

PR #292, opened 2026-08-03
feat(harness)!: link a thread to the one that spawned it, and give it a type

### LINKED ISSUE #235: Multi-session data model: parent/child linkage + thread type
**Parent:** #224

## Goal

The original core of #224, landed on the single-homed model: a thread records **what kind of session it is** and **which thread spawned it**, so a report session can live as a child of the analysis conversation that spawned it.

## Scope

- **Columns** (additive, both halves of the established migration pattern): `thread_type TEXT NOT NULL DEFAULT 'conversation'` and `parent_thread_id TEXT REFERENCES cortex_analysis_threads(thread_id)` (first FK on the table; the `ADD COLUMN ... REFERENCES` form is precedented), plus a partial index for child listing.
- **Store surface**: `createThread` accepts `type`/`parentThreadId`; `Thread` exposes them; `listThreads` gains `type` and `parentThreadId` filters — this covers both "list a session's children" and "the report sessions of this analysis" without new methods.
- **Integrity**: a parent must exist and belong to the same analysis — enforced in `createThread`. This is data integrity, not authorization, so the store's no-scope-checks rule stands.
- **Creation semantics**: lazy first-turn creation always yields a default `conversation` thread with no parent; typed/child threads are created **eagerly by the deliberate spawn action**. `createThread`'s never-overwrite idempotency protects spawned rows from a racing lazy create.
- **Lifecycle across the edge**: archive/delete of a parent cascades to its children, defined against the lifecycle sibling issue's verbs (single-level semantics for now).
- **Specs**: deltas to `harness-thread-store` plus a new capability spec for the multi-session model.

## Open questions

- **Session-per-report vs session-per-version** — #221 reads both ways ("a report lives in its own child session" vs "the report versions of this analysis"). Affects whether a version link belongs on the thread row. Settle before writing the spec delta.
- **`parent_seq` checkpoint pin** (the parent's `messages.seq` at spawn — the frozen-at-T anchor #223 assumes): cheap and well-defined now (`messages` PK is `(thread_id, seq)`); deferring keeps #223 unconstrained. Leaning: add it nullable now.

## Out of scope

- Type→agent selection (sibling issue).
- The Report Builder agent and its tools (#225); CLI navigation UX (separate follow-on).

vivere-dally:
## What & why

A report must leave the analysis conversation and run in its own session. The thread model cannot express that today: every thread under an analysis is the same kind of thing, and no row records which thread spawned it.

`cortex_analysis_threads` gains three columns. `thread_type` records what kind of session a thread is. `parent_thread_id` records which thread spawned it. `parent_seq` records the point in that parent's transcript where the spawn happened.

Two pieces of work wait on this column set: the seam that selects an agent from a thread's type, and the Report Builder agent that runs in the child session.

Closes #235

## The data model

An analysis conversation `X` spawns a report session `Y1`. More analysis work follows in `X`. A second request spawns `Y2`.

```
X   thread_type='conversation'  parent_thread_id=NULL  parent_seq=NULL
├── Y1  thread_type='report'    parent_thread_id=X     parent_seq=40
└── Y2  thread_type='report'    parent_thread_id=X     parent_seq=210
```

`Y1` and `Y2` are flat children of `X`. Neither refers to the other.

One session holds one report version. Version lineage stays on the artifact side, where `previewId` and `baseVersion` already carry it. A single `parent_thread_id` therefore always points at one place, and `parent_seq` indexes that same place.

`thread_type` is a closed set of two values, `conversation` and `report`. Nothing here creates a `report` thread — the spawn action arrives with the Report Builder work. The set is closed because a thread's type selects the agent that runs it, and that selection can only be exhaustive over a known membership.

## Warning: the database cascade cannot do the delete

`messages` carries no foreign key to `cortex_analysis_threads`. A cascade removes a descendant's metadata row, but it cannot reach that descendant's transcript. Those message rows then belong to no analysis, and no later reclamation of any scope finds them.

`purgeThread` therefore walks the subtree and deletes each transcript explicitly, to the depth the cascade reaches. `ON DELETE CASCADE` remains behind it as a backstop.

`purgeAnalysis` needs no code change, but its statement order is now load-bearing. It deletes messages before thread rows. Reverse the two and the cascade empties the thread table first, so the message delete finds no rows to reclaim. A new requirement in the `analysis-purge` spec records this.

## Archive reaches the subtree, unarchive reaches one row

`archiveThread` stamps the named thread and every descendant. `unarchiveThread` clears the stamp on the named thread alone.

The asymmetry is deliberate. A symmetric cascade restores a child that the user archived on its own beforehand. The schema would then need a fourth column to record which action set each tombstone. With the asymmetry, no row carries that distinction, and a user recovers any archived thread the same way — by naming it.

## Integrity rules

`createThread` enforces the two rules a constraint cannot express: a parent belongs to the child's analysis, and a parent and an anchor arrive together. A parent that names no row goes to the foreign key, which trips `constraint_violation` without the round trip an existence query costs.

Every rule is conditional on the caller supplying a value. A create that names neither a parent nor an anchor reaches none of them, so first-turn conversation creation keeps the path it has today.

## Breaking change

`Thread` gains three required fields, so a consumer that builds one from a literal must supply them. `createThread`'s error channel widens to `DbError | ThreadInputError`. Both new types are re-exported from the package barrel.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally — see "What ran" below, which states this precisely.
- [x] Documentation updated if user-facing behavior changed. The OpenSpec deltas carry it; `harness/CONTEXT.md` names no thread-store verb, so nothing there went stale.
- [x] Commits follow Conventional Commits.
- [x] Commits are signed off for the DCO.

## What ran

- `tsc -p tsconfig.json` — exit 0.
- `bunx eslint` over the five changed files — exit 0. A full `bun run lint` fails on `scripts/smoke.mjs`, which this branch does not touch and which fails the same way on `main`.
- `bun test` over `thread-store`, `thread-history`, `purge-analysis`, and `chat-turn` — 113 pass, 0 fail. **The full suite did not run.**

Two behaviours were confirmed by mutation rather than by reading:

1. Cripple the recursive subtree walk to one level, and 5 tests fail across archive, unarchive, and purge.
2. Move the message delete out of `purgeThread`'s transaction, and both partway-failure tests fail.

The schema statements were also run against `pgvector/pgvector:pg18`. The self-referencing `ADD COLUMN IF NOT EXISTS` form is accepted, and a re-run is a no-op.


vivere-dally:
@claude review

radutanasa (approved):

PR #293, opened 2026-08-03
Remove the test preload that leaks file descriptors, and pin the Bun version

### LINKED ISSUE #282: Test suite peaks at ~9.5 GB RSS: re-check whether `--isolate` is still needed, and bound the memory either way
## The problem

A full `bun run test` in `cli/` peaks at roughly **9.5 GB of RSS**. On a developer machine running anything else — a second checkout, a browser, the Podman VM — that is enough to push the system into swap. Two concurrent runs made a laptop unusable and killed both runs by `SIGTERM`/`SIGKILL`, which also produced phantom test failures that looked real until the runs were repeated in isolation.

That is not a defensible amount of memory for a test suite, and it is the thing this issue exists to fix.

## Measurements

Taken on macOS, `bun 1.3.14`, in `cli/`, with `/usr/bin/time -l`:

| command | wall time | peak RSS |
|-|-|-|
| `bun test --isolate src/tui/commands.test.ts` (one file) | 2.1 s | 476 MB |
| `bun run typecheck` | 3.3 s | 919 MB |
| `bun run lint` | 8.4 s | 2.1 GB |
| `bun run test` (full suite, 141 files) | ~220 s | **~9.5 GB** |

## Why it happens

`bun run test` is `bun test --isolate`. Every test file runs in a fresh realm inside **one long-lived process**, and native resources (opentui, tree-sitter, sqlite) are not fully reclaimed between files, so RSS climbs monotonically for the whole run. One file costs 476 MB; 141 of them in a single process is the 9.5 GB.

`--isolate` is not there by preference. It works around a Bun-on-macOS bug where a child process's pipe output vanishes after certain module graphs have loaded earlier in the run — dozens of spawn-dependent tests (the provenance bridge, `spawnInflexa` process bounds, the embedding launch drain) fail with empty captured output under a plain `bun test`. `--isolate` avoids the corruption entirely. Linux CI is unaffected by the original bug.

So the memory cost is the price of a workaround, and whether we still need to pay it has not been re-checked since the workaround was adopted.

Worth ruling out first: every one of the 33 render-test files disposes its renderer (`renderer.destroy()` in a `finally`), so this is **not** the known undisposed-`testRender` hazard.

## Required outcome

One of these two, and in either case the RAM ceiling is the point:

1. **The pipe bug is fixed in the current (or a newer) Bun → drop `--isolate`.** Re-run the full suite without it on `1.3.14` and confirm the spawn-dependent tests pass. If they do, remove the flag: one shared realm instead of 141, and a workaround we no longer maintain. Note we are **already on the newest published Bun** (`1.3.14` locally and on npm), so "upgrade" here means verifying the current version rather than moving to a new one — and re-verifying whenever Bun does release.
2. **The bug still reproduces → keep `--isolate`, but the memory must still be measured and bounded.** Upgrading Bun does not automatically fix this, so an upgrade is not a substitute for measuring. If `--isolate` stays, shard the run so no single process accumulates all 141 files.

Either way, **we must not end up back at a ridiculous amount of RAM for running tests.** Suggested acceptance: record peak RSS for the full suite in this issue, and treat a multi-gigabyte figure as a defect rather than a fact of life. If the growth turns out to be unbounded accumulation across isolates, that is worth reporting upstream to Bun regardless of which option we take.

## Related: CI does not pin Bun

Every workflow uses `oven-sh/setup-bun@v2.2.0` with **no `bun-version` input**, so CI resolves whatever is latest at run time while a developer machine sits wherever it was last upgraded. They agree today. The day they do not, a green CI stops meaning anything locally and a Bun regression arrives unannounced. Pinning the version — and bumping it deliberately — is a small change that also makes the experiment above reproducible.

## Interim working rule

Until this is resolved: run the specific test files you changed (`bun test --isolate <file>` — seconds, sub-gigabyte), keep `typecheck` and `lint` as the cheap local gates, and treat a full-suite run as a single deliberate act, never something run concurrently with another. CI runs the full suite on Linux anyway.

vivere-dally:
## What & why

The test suite used the command `bun test --isolate`. A note in `cli/CLAUDE.md` said that this flag prevented a Bun defect on macOS. It said that the defect made the output of a child process empty. That note was not correct.

The true cause is in `cli/bunfig.toml`. The `[test].preload` list contained the item `@opentui/solid/runtime-plugin-support`. This item registers a catch-all `onResolve` hook. The hook opens directory handles, but it does not close them. Thus one test process opens more than 21,400 file descriptors.

macOS has a limit with the name `OPEN_MAX`. The value of this limit is 10240. If the number of a file descriptor is more than this limit, `Bun.spawn` gives the child process `/dev/null` in place of a socketpair. The child process then operates correctly and exits with code 0. But the parent process reads zero bytes. Thus seven tests that start a child process failed. No error message showed the cause.

The flag `--isolate` did not correct this defect. It only hid the defect. The flag removes the realm of each test file after that file. This operation releases the file descriptors of that file.

Closes #282

## The changes

- Remove `@opentui/solid/runtime-plugin-support` from `[test].preload` in `cli/bunfig.toml`.
- Set the `test` script in `cli/package.json` to `bun test`.
- Correct the notes in `cli/CLAUDE.md`, `cli/src/test_support/preload.ts`, and `cli/src/tui/hooks/conversation.usage_recorder.test.ts`. These notes described the incorrect cause.
- Add a `.bun-version` file. All seven `setup-bun` steps read this one file. Before this change, each workflow installed the most recent Bun at run time.

## Results

| Command | Result | Time | Peak memory | Highest fd |
|-|-|-|-|-|
| `bun test --isolate` | pass | 221 s | 5.70 GiB | — |
| `bun test --isolate --shard=i/8` | pass | 114 s | 2.13 GiB | — |
| `bun test` (this PR) | pass | 58 s | 0.73 GiB | 2307 |

The suite runs 2652 tests in 156 files: 2651 pass, 1 skip, 0 fail. These counts agree with the counts from the `--isolate` run. A comparison of the junit results shows the same set of tests. The only difference is that the seven spawn tests now pass.

All 38 render test files pass without the preload item. They run the same 412 tests and make the same 1994 assertions.

The highest file descriptor number is now 2307, and the limit is 10240. Thus the margin is 4.4 times.

Local runs and CI runs now use the same command.

## Notes

- Bun issue #32067 records the silent replacement of the pipe with `/dev/null`. Bun PR #32068 corrects it. No Bun release contains that correction.
- The test `src/tui/hooks/run_completion.test.ts` is not reliable. It failed 4 times in 12 runs with `--isolate`, and 0 times in 6 runs without `--isolate`. This defect is not new, and this PR does not correct it.
- This PR contains two commits. The first commit pins the Bun version. The second commit corrects the test preload. The two commits are related, because a pinned Bun version makes the measurements above repeatable.

## Type of change

- [x] Bug fix
- [x] Documentation

## Checklist

- [ ] Tests added or updated for the change. No new test is necessary. Seven existing tests changed from fail to pass.


vivere-dally:
@claude review

radutanasa (approved):

PR #297, opened 2026-08-03
feat(harness): add citation resolution

radutanasa:
## Summary

- add a host-agnostic CitationResolver with deterministic normalization, source planning, aggregation, comparison, verdicts, bounded batching, caching, and rate limiting
- expose citation verification through the conversation agent, literature reviewer, sandbox allowlist, and shared runtime assembly
- consolidate arXiv, Semantic Scholar, and PubMed wire access into shared authority clients used by both discovery tools and citation adapters
- add opt-in external contract coverage and complete OpenSpec artifacts for citation resolution

## Review follow-up

Two rounds of review feedback are folded in:

- a source's rate limiter wrapped the fetch it replaced, so a request's timeout ran while it waited its turn — a paced source reported its own queue as upstream timeouts. The HTTP layer now takes a `schedule` admission seam and arms the timeout inside it.
- author comparison fell through to the surname, so `John Smith` verified against `Jane Smith`. The rule is now surname-position free and requires given-name agreement where both sides supply one.
- Crossref reported "DOI is not registered by Crossref" whenever the registration agency was merely undetermined. It now separates that gap (`unavailable`) from an actual finding of other-agency ownership (`not_applicable`) — which also closes a path where a DOI neither authority had checked could reach `not_found` on complete coverage.
- a pinned `free_text` kind no longer leaks an extracted identifier into planning and candidate scoring.
- plus: consistent DOI URL encoding, per-source containment of a throwing or inconsistent client, lookup-identity caching with per-caller comparisons, the full `CitationSourceClient` contract exported, and the Semantic Scholar key injected through `BioToolKeys` rather than read from the environment inside the tool.

## Validation

From `harness/`:

- `bun run test:full` — 1980 pass, 6 skip, 1 fail
- `tsc -p tsconfig.json --noEmit`
- `bun run lint`
- `openspec validate add-citation-resolution --strict`
- `git diff --check`

The single failure is `configured-provider.barrel.test.ts` → "recognizes claude-opus-5, clamping an over-large ceiling to its real limit" (expects 128000, gets 200000). It is pre-existing and unrelated to this change — it reproduces on `main` and on this branch with the citation work stashed.

The earlier note that `test:full` could not start for want of a container runtime no longer applies; it has since been run, and OpenSpec task 8.3 is checked accordingly.


radutanasa:
@claude review

harness/src/agents/sandbox/shared.ts
```diff
@@ -155,13 +159,17 @@ export interface SandboxAgentPromptOptions {
 function resolveSandboxTools(deps: SandboxAgentDeps, tools: readonly SandboxToolName[]): Tool[] {
     const ncbi = createNcbiTools(deps.bioKeys);
     const chemDb = createChemDbTools(deps.bioKeys, { ...(deps.logger ? { logger: deps.logger } : {}) });
-    const registry: Record<SandboxToolName, Tool> = {
+    if (tools.includes("resolve_citation") && deps.citationResolver === undefined) {
+        throw new Error('createSandboxAgent: SandboxToolName "resolve_citation" requires a CitationResolver dependency.');
+    }
+    const registry: Record<SandboxToolName, Tool | undefined> = {
```
vivere-dally, line 165: Can we use `neverthrow` here?

harness/src/lib/cache.ts
```diff
@@ -0,0 +1,38 @@
+interface CacheEntry<Value> {
+    readonly value: Value;
+    readonly expiresAt: number;
+}
+
+export class BoundedTtlCache<Value> {
```
vivere-dally, line 6: It's interesting that we have a cache impl, shouldn't we use a well-known lib here? I'm fine with it, but I'd like some more serious tests if we decide on own impl

harness/src/citations/compare.ts
```diff
@@ -0,0 +1,80 @@
...
+    CitationCandidateCluster,
+    CitationConflict,
+    CitationField,
+    CitationFieldComparison,
+    CitationInput,
+    CitationRecord,
+    CitationSource,
+} from "./types.js";
+
+export const CITATION_COMPARISON_RULE_VERSION = "citation-compare-v2";
```
vivere-dally: why v2?

harness/src/citations/normalize.ts
```diff
@@ -0,0 +1,197 @@
+import { CitationInputSchema, type CitationInput, type NormalizedCitation, type NormalizedCitationMetadata, type UnsupportedWorkKind } from "./types.js";
+
+const DOI_PATTERN = /(?:https?:\/\/(?:dx\.)?doi\.org\/|doi\s*:\s*)?(10\.\d{4,9}\/[^\s<>"']+)/i;
+const PMID_PATTERN = /^(?:pmid\s*:\s*)?(\d{1,9})$/i;
+const ARXIV_PATTERN = /^(?:https?:\/\/arxiv\.org\/(?:abs|pdf)\/|arxiv\s*:\s*)?((?:\d{4}\.\d{4,5}|[a-z-]+(?:\.[a-z-]+)?\/\d{7})(?:v\d+)?)(?:\.pdf)?$/i;
```
vivere-dally, line 10: ih, I guess we have to?

harness/src/lib/rate-limit.ts
vivere-dally, line 1: Again, some core `lib` kind of implementation that is dumped here. Maybe use a dependency, or move this to a `lib/` dir - and ensure tests are rigorous

vivere-dally (changes requested):

cli/src/modules/harness/config.ts
```diff
@@ -165,6 +167,9 @@ function defaultsWith(cfg: z.infer<typeof harnessConfigSchema> | undefined, conf
             epaCcte: cfg?.bioKeys?.epaCcte ?? "",
             ncbi: cfg?.bioKeys?.ncbi,
             github: cfg?.bioKeys?.github,
+            // The harness no longer reads this variable itself (no ambient reads below the
+            // composition root), so the host resolves it here and keeps `SEMANTIC_SCHOLAR_API_KEY` working.
+            semanticScholar: cfg?.bioKeys?.semanticScholar ?? readEnvCredentialVar("SEMANTIC_SCHOLAR_API_KEY"),
```
krajla: This one config entry doesnt deserve the prose more than the others. Also let's inject it the same without reading env at this stage.

krajla (commented):
Just one thing besides what Stefan flagged.

harness/src/lib/rate-limit.ts
radutanasa: Moved — `rate-limit.ts` is now `src/lib/rate-limit.ts` (and `cache.ts` alongside it in `src/lib/`). You were right that it had already outgrown `citations/`: `literature/sources/http.test.ts` was reaching into `citations/` for the scheduler its own module consumes, which is the smell you picked up on.

I left it hand-rolled rather than taking a dependency. The harness is published, so a dep here lands in every consumer's tree, and the limiter is ~90 lines doing something the off-the-shelf ones mostly do not: it paces *and* bounds concurrency *and* is cancellation-aware throughout (a waiter aborted mid-queue is spliced out of the semaphore rather than left to be resolved). Happy to swap it if you have a library in mind that covers those three.

Tests broadened per your ask — slot released when the operation throws (so a failure cannot wedge the limiter), an already-aborted signal refused before a slot is taken, no credit banked while idle, per-operation return values preserved under interleaving, invalid configuration rejected, and `sleepWithSignal`'s three paths.

harness/src/lib/cache.ts
radutanasa: Same move — `cache.ts` now lives in `src/lib/`.

Kept as-is for the same published-package reason, and because the surface is small and specific: an LRU where the TTL is per `set` (positive and negative entries get different lifetimes) over an injected clock so the tests are deterministic. `lru-cache` would cover it, but it is a dependency every consumer inherits for ~35 lines.

Tests broadened as you asked — reads count as uses so recency follows gets and not only sets, re-setting a key refreshes in place without growing the map, the exact expiry boundary (held at t-1, gone at t), an expired entry reclaimed on read rather than left occupying the bound, non-positive TTLs and a zero maximum storing nothing, the bound held under sustained insertion, and an invalid maximum rejected. 2 cases → 11.

harness/src/citations/compare.ts
radutanasa: Good catch — reverted to v1.

I bumped it in this PR because the author-comparison rule changed, but that was wrong: v1 has never shipped. The released package (`@inflexa-ai/harness@0.16.0`) has no `citations/` at all — I checked the published tarball, not just `node_modules`. The version labels *published* comparison semantics for a consumer reading `ruleVersion` off a field comparison, so a revision made before the first release is not a new version, and shipping straight to v2 only raises the question you just asked. Added a comment on the constant saying when to bump it.

harness/src/citations/normalize.ts
radutanasa: Pretty much, yes. Each registry defines its own identifier syntax and a citation is prose, so the identifier has to be found inside the string rather than read off a field — there is no parser to reach for.

Added a comment explaining the part that actually is surprising: the DOI suffix is deliberately permissive (`[^\s<>"']+`) because the registry allows nearly any character in it, so there is no delimiter to stop at — which is why `trimDoiPunctuation` exists to walk back the sentence punctuation and brackets a greedy match swallows. PMID and arXiv anchor instead, having fixed shapes. All three are covered by `normalize.test.ts`.

cli/src/modules/harness/config.ts
radutanasa: Both done — the comment is gone and the entry is now just `semanticScholar: cfg?.bioKeys?.semanticScholar`, identical in shape to `ncbi` and `github`.

One consequence worth flagging since it is a user-visible change rather than a refactor: today `SEMANTIC_SCHOLAR_API_KEY` works because the *harness* reads it ambiently inside the tool, and this PR removes that read (it is below the composition root). The env fallback here was covering that gap. Without it, once the harness release lands, anyone relying on the variable has to move the key to `harness.bioKeys.semanticScholar` in `config.json`. Consistent, and I think the right call — just noting it so it can be mentioned in the release notes rather than surprising someone.

harness/src/agents/sandbox/shared.ts
radutanasa: I would rather leave this one as a throw, and I think the surrounding code agrees.

`resolveSandboxTools` already throws ~30 lines below for the same class of problem — an `AgentMeta` naming a `SandboxToolName` with no implementation. Both are composition-time invariant violations: an agent declared a tool whose dependency was never wired, which no runtime handling can recover from and no caller can act on. Per `lib/result.ts`, `err` is reserved for what would otherwise be thrown as an *unexpected upstream failure*; a static wiring bug is neither upstream nor recoverable, and failing loudly at construction is what stops it reaching a step.

Converting only mine would leave the same function throwing for one wiring bug and returning a `Result` for the other. Converting both would change `createSandboxAgent`'s signature to `Result<AgentDefinition, …>` and ripple through all 16 sandbox agents plus `assembleCoreRuntime` — worth doing deliberately if you want that boundary to speak `Result`, but as its own change rather than smuggled in here. Happy to open one if you would like it.

vivere-dally (approved):

Issue #236, opened 2026-07-27
Thread type selects the agent: the type→agent resolution seam


vivere-dally:
**Parent:** #224

## Goal

A thread's **type selects the agent that runs it** — resolved by the harness, not the embedder, so a managed deployment gets the same behavior with zero CLI involvement.

## Today (verified)

The harness has no agent registry. `prepareChatTurn` never names an agent; the embedder passes the single `runtime.conversationAgent` into every turn. The only id→roster machinery anywhere is the sandbox catalog (plan-step agents only), and the current report-builder is constructed inline per invocation inside `runReportIteration` with closure-bound tools — deliberately outside any catalog.

## Scope

- A **type-keyed agent resolution at the composition root** (`assembleCoreRuntime`): one registered type today (`conversation` → the existing conversation agent), and a registration point where #225's Report Builder plugs in by registering — not by rewiring the CLI.
- The embedder's turn engine asks the harness for the thread's agent instead of hardcoding `conversationAgent`.
- Shape the seam so a type's agent carries its own tool roster and limits (`AgentDefinition` is already `{id, systemPrompt, model, tools, maxIterations}`).
- Spec delta recording the selection rule (type → agent is harness-owned).

## Out of scope

- The Report Builder agent itself, its roster, and the report-session runtime (#225). Note for #225: today's report-builder binds tools in a per-preview closure; registering at assembly forces that tool-binding redesign.

## Depends on

The data-model sibling issue (the `thread_type` column). #225 consumes this.

PR #298, opened 2026-08-03
feat(harness)!: select the agent that runs a thread from the thread's type

vivere-dally:
## What & why

A thread's type now selects the agent that runs it, and the harness owns that selection. `assembleCoreRuntime` builds a type-keyed registry of the agents it assembles and exposes resolution as `CoreRuntime.agents` — `forThread(type)` is the only way to reach an assembled agent by thread type. `conversation` resolves to the conversation agent; the registry is the single registration point where a future typed agent (the Report Builder) plugs in at assembly, with no embedder change.

Part of #236 — the harness half. The embedder's turn-engine wiring is CLI-side work outside this PR.

## The seam's three properties

**Resolution returns assembled singletons.** `forThread` hands back the same `AgentDefinition` object on every call. An embedder may wrap providers in delegating handles it re-points at idle, and the assembled tools capture those handles at construction — a per-call reconstruction would orphan them. The rejected factory shape is also what forces the Report Builder's tool-binding redesign: per-thread state moves to call-time `ToolContext`, not construction-time closures.

**An unregistered type refuses on the `Result` channel, and the channel is permanent.** `report` is a valid `ThreadType` with no registered agent yet; `forThread("report")` returns `{ type: "unregistered_thread_type", threadType: "report" }`, never a throw. The signature does not narrow once every current member registers — otherwise each future `ThreadType` member would have to arrive with its agent in the same commit.

**`prepareChatTurn` surfaces the thread's type.** The `ok` result carries `threadType`, read from the row the ownership check already loads — zero extra queries. A thread the call creates reports `conversation`, the store default, even when the best-effort create fails non-fatally.

## Breaking change

`CoreRuntime.conversationAgent` is removed; consumers resolve via `agents.forThread("conversation")`. `PrepareChatTurnResult`'s `ok` variant widens by one required field, so a seam fake built from a literal must supply it. `ThreadAgentResolver` and `UnregisteredThreadType` are re-exported from the package barrel.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally — see "What ran" below, which states this precisely.
- [x] Documentation updated if user-facing behavior changed. The OpenSpec specs carry it (a new `thread-agent-resolution` capability spec, synced); no harness doc names the removed field, verified by grep.
- [x] Commits follow Conventional Commits.
- [x] Commits are signed off for the DCO.

## What ran

- `tsc -p tsconfig.json` — exit 0.
- `bunx eslint` over the six changed files — exit 0. The full `bun run lint` did not run.
- `bun test` over `assemble`, `boot`, and `chat-turn` — 13 pass, 0 fail, against `pgvector/pgvector:pg18`. **The full suite did not run.**

One deviation, recorded in the archived change's `design.md`: the resolver tests drive the extracted pure `createThreadAgentResolver`, not the real `assembleCoreRuntime` — `DBOS.registerWorkflow` refuses duplicate and post-launch registration, so the real call is single-shot per process and untestable as an offline unit. The registry literal stays inline in `assembleCoreRuntime` under the typechecker. The `report` refusal is likewise driven directly in a unit test: no production path creates a `report` thread yet.

### DIFF
--- a/harness/openspec/changes/archive/2026-08-03-add-thread-agent-resolution/.openspec.yaml
+++ b/harness/openspec/changes/archive/2026-08-03-add-thread-agent-resolution/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-03
--- a/harness/openspec/changes/archive/2026-08-03-add-thread-agent-resolution/design.md
+++ b/harness/openspec/changes/archive/2026-08-03-add-thread-agent-resolution/design.md
@@ -1,0 +1,68 @@
+# Design: add-thread-agent-resolution
+
+## Context
+
+`assembleCoreRuntime` builds exactly one agent and returns it as `CoreRuntime.conversationAgent` (`src/runtime/assemble.ts:103-111`). Two consumers hold that field: the boot-time display backfill, which needs the conversation tool roster (`src/runtime/boot.ts:108`), and the embedder, which passes the field into every turn's `runAgent`. `prepareChatTurn` loads the thread row for its ownership check (`src/app/chat-turn.ts:57`) and discards `threadType`. The type itself landed in #292: a closed two-member set, `conversation` and `report` (`src/memory/thread-store.ts:66-67`), closed precisely because type→agent selection must be exhaustive over a known membership.
+
+Constraints from the surrounding system:
+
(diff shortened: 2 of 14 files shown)


vivere-dally:
@claude review

radutanasa (approved):

PR #301, opened 2026-08-04
refactor(harness): share the durable LLM step and run finalisation

radutanasa:
## What & why

Two mechanisms in the harness were written once and needed twice, and both encode subtleties that are invisible in a fresh implementation. This lands both moves on their own, so a behaviour-preserving refactor of two existing workflows is reviewed as a refactor rather than as a diff buried inside a feature.

**`runLlmStep` / `structuredLlmCall`** move from `workflows/target-assessment/lib/` to `workflows/lib/`, unchanged. They own the per-call `DBOS.runStep` cache slot, the forced-`submit`-tool structured output, and the billing-gateway 402 → marker → suspend/self-cancel choreography that makes an insufficient-funds pause resumable. They lived under `target-assessment/` only because they were written there.

**`collectAndComplete`'s generic half becomes `finaliseRun`.** What is shared is the *order*, because every constraint in it is a bug found once:

1. a checkpointed clock read and the caller's terminal provenance **before** the status write — a host polling `cortex_runs.status` may shut down the instant it leaves `running`, and emitting after the write races that shutdown;
2. the status write before any ledger note, so a reader seeing a terminal status also sees how it got there;
3. the pending-row sweep and the budget suspend as exact complements of **the branch the caller took, never the status it wrote** — the resumable 402 pause also writes `"canceled"`, yet its `pending` rows must survive for the resumed workflow;
4. charge close, then authorization revoke;
5. exactly one terminal stream part, last.

Each side effect keeps its own named step and the log-don't-roll-back rule: a finished run is not undone by a failed bookkeeping note.

`executeAnalysis` keeps what is genuinely its own — the status derivation, the synthesis-outcome note, and building the terminal part (whose artifact count is its own durable step). Net −150 lines there.

### Durable identities deliberately preserved

- **`BUDGET_EXCEEDED_TOPIC = "ta-budget-exceeded"`** — a topic is persisted with its message in DBOS's notification table, so renaming it would strand a marker sent by a workflow already in flight.
- **`Symbol.for("ta.budget-exceeded")`** — a registry identity other code pins.
- **Every durable step name** — built at the target-assessment call sites (`ta-decision:…:${attempt}`, `ta-synth:…:${attempt}:N`), never inside the moved modules, so no cache key moves and an in-flight workflow replays into the same slots.

The `ta-` spellings are identities, not statements about who may call. One thing *does* change: the log namespace `ta-llm-step` → `llm-step`, since `named()` is the module prefix and a shared module must not mislabel the records of everything that will call it. Nothing asserts it.

### Verification — evidence, not assertion

- The 15 durable step names extracted from `HEAD` and from the new pair of files **diff empty**.
- The emitted part-type multiset and `deriveFinalStatus` are **byte-identical** to `HEAD`.
- `src/workflows` is **118 pass / 0 fail** before and after. The only edit to a test file is one import specifier — no assertion touched.
- `tsc -p tsconfig.json` and `eslint .` clean.

### Known pre-existing failure (not from this PR)

`bun run test:full` reports **2003 pass, 6 skip, 1 fail**. The failure is `src/providers/configured-provider.barrel.test.ts` → *"recognizes claude-opus-5, clamping an over-large ceiling to its real limit"* (expected `128000`, got `200000`).

Confirmed pre-existing by stashing this branch's changes and re-running at `HEAD` — identical failure. Cause is external drift: the installed `@ai-sdk/anthropic` is 4.0.25 against the `^4.0.20` floor, and its model table no longer clamps `claude-opus-5` to 128k. The test's own comment says it guards that dependency floor. Fixing it means deciding whether opus-5's real ceiling is now 200k (update the expectation) or the SDK regressed (pin the dependency) — a separate change either way.

### Follow-on

First of a five-PR series; the specs for the rest are on branches stacked on this one. The consumers of these two extractions are `add-manuscript-reference-review` (the finaliser) and `add-manuscript-editorial-review` (the LLM step).

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `tsc -p tsconfig.json`, `bun run test:full` — see the pre-existing failure noted above).
- [x] Tests added or updated for the change. *No new tests: the existing `executeAnalysis` and `executeTargetAssessment` suites passing unmodified **is** the acceptance criterion for a behaviour-preserving extraction, and step-name/part/status equality is verified against `HEAD` directly.*
- [x] Documentation updated if user-facing behavior changed. *No user-facing behaviour change; the OpenSpec deltas record that the mechanism is now shared.*

### DIFF
--- a/harness/openspec/changes/share-durable-workflow-mechanics/.openspec.yaml
+++ b/harness/openspec/changes/share-durable-workflow-mechanics/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-04
--- a/harness/openspec/changes/share-durable-workflow-mechanics/proposal.md
+++ b/harness/openspec/changes/share-durable-workflow-mechanics/proposal.md
@@ -1,0 +1,79 @@
+# Make the two durable-workflow mechanics shareable before a third workflow needs them
+
+## Why
+
+Two mechanisms in the harness are written once and needed twice, and both encode
+subtleties that are invisible in a fresh implementation:
+
+**Run finalisation.** `collectAndComplete` (`src/workflows/execute-analysis.ts`)
+settles every step row, sweeps never-started rows to `skipped`, writes the run's
+terminal status, closes the run charge, revokes owned authorization, and emits
+exactly one terminal event — each in its own named `DBOS.runStep`, none rolling
+back the ones that already succeeded, with the terminal event emitted *before*
+the status write so no reader can observe a terminal run without it, and with the
+402 pause branch selected structurally rather than inferred from the written
+status. Every clause there is a bug that was found once. A second workflow that
+writes its own version does not inherit any of them.
+
+**Durable model calls.** `runLlmStep` and `structuredLlmCall`
+(`src/workflows/target-assessment/lib/`) own the per-call `DBOS.runStep` cache
+slot, the forced-`submit`-tool structured-output pattern, and the
+billing-gateway 402 → self-addressed marker → suspend/self-cancel choreography
+that makes an insufficient-funds pause resumable. They live under
+`target-assessment/` for no reason other than being written there first; nothing
+in them is target-assessment-specific.
+
+`add-manuscript-reference-review` is the second consumer of the finaliser and
+`add-manuscript-editorial-review` is the second consumer of the LLM step. This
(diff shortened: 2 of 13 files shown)


radutanasa:
@claude review

vivere-dally:
DFTM

PR #324, opened 2026-08-05
feat(prov-kernel): add the provenance format kernel package

krajla:
## What

A new top-level subsystem `prov-kernel/`, published as `@inflexa-ai/prov-kernel` (0.2.0). It carries the Inflexa PROV dialect and nothing else:

- the document model: QName derivation, tsprov statement builders, unify options, an injectable digest
- the chain-hash and Ed25519 sign/verify primitives, and the signed-sidecar schema
- the actor/ref value types the builders accept
- `SPEC.md`: the wire rules, precise enough for a non-TypeScript writer
- a golden-fixture test that guards cross-recorder drift

Deliberately absent: any event vocabulary, event reducer, recorder lifecycle, or harness bridge. The package has no dependency on `@inflexa-ai/harness`.

## Why

This replaces #304. The harness must not own the tracked vocabulary: the harness observes (collector + provenance hooks), this kernel represents (dialect, document, integrity), and each host decides (its recorder, its event set, its sink and signer). The CLI keeps its recorder; a managed deployment writes its own; both consume this kernel. A future Go writer implements against `SPEC.md`.

## CI

- `test (prov-kernel)` and `lint (prov-kernel)` jobs, mirroring the harness jobs
- `release-prov-kernel.yml`: registry-gated, idempotent npm publish via trusted publishing (OIDC), `prov-kernel-v*` tag
- `version-guard.yml` now guards the third manifest

**Before the first release lands**: npm needs a trusted-publisher entry for `@inflexa-ai/prov-kernel` bound to this repo and `.github/workflows/release-prov-kernel.yml`, like the harness one.

## Gates

`bun run typecheck`, `bun run lint`, `bun test` (25 pass / 0 fail), `bun run build` + `node scripts/smoke.mjs`, `openspec validate --all --strict`, actionlint on all touched workflows — all green.

## Follow-ups (separate changes)

- Cortex: its own recorder + Nexus CAS sink, consuming this kernel
- cli: optionally rebase its local `document.ts`/`signing.ts` copies onto the kernel


radutanasa (dismissed):

prov-kernel/src/document.ts
```diff
@@ -0,0 +1,570 @@
...
+        appendCreation,
+        appendInputAdded,
+        appendInputRemoved,
+        appendRunStarted,
+        appendRunCompleted,
+        appendStepCompleted,
+        appendCommandExecuted,
+        appendFileWritten,
+        appendInputUsed,
+    };
```
vivere-dally, line 600: This is a bit weird. Have you considered swallowing that big switch statement from the CLI repo? I understand this as that you're exporting the functions required to register the prov events.

krajla: It was an over correction in the direction of the consumer owning the format. But i think its better to move the switch here.

krajla: Check latest

vivere-dally (approved):

PR #325, opened 2026-08-05
refactor(cli): home the dev-only command surfaces, and guard the boundary

### LINKED ISSUE #56: Home the dev-only command surfaces under modules/harness/dev/
From the PR #45 review (thread: https://github.com/inflexa-ai/inf-cli/pull/45#discussion_r3551050522): make "these are local-DX / agent-E2E surfaces" legible in the tree, complementing the single dev-channel registration gate.

**Movable as-is:** `cli/src/modules/harness/chat.ts` (the clack REPL) and `run.ts` — their only importer is the dev-gated registry (`cli/src/cli/index.ts:138-175`, behind `devCommandsEnabled()`).

**Needs a split first:** `profile.ts` is not dev-only —
- `describeBootError` + `ensureSandboxImage` feed the product TUI boot path (`src/tui/app.launch.tsx:14`, `src/tui/hooks/boot.ts:5`)
- `seedProfileLedger` feeds the parity trigger (`src/modules/harness/profile_trigger.ts:17`)
- `run.ts`/`chat.ts` lean on `resolveSingleAnalysis`, `formatElapsed`, `readNewestWorkflowStep`, `withStatusPool` and friends

**Proposed cut:** move `chat.ts` + `run.ts` into `modules/harness/dev/`, extract the `runProfile`/`runProfileStatus` command actions there too, and leave the shared boot/profile helpers where the product imports them. Update all importers per the no-shims rule.

vivere-dally:
## What & why

The CLI gates its dev/E2E commands at registration — `src/cli/index.ts`, behind `devCommandsEnabled()` — and the `dev-commands` spec records that gate. The tree said nothing. A reader who opened `src/modules/harness/` found `chat.ts`, `run.ts`, and `profile.ts` beside the product modules. Nothing there showed that a release build carries none of the three.

Two files blocked a plain move, because each mixed a shared half with a dev half:

- `profile.ts` (498 lines) held three helpers the product imports — `describeBootError` (the TUI boot path), `ensureSandboxImage`, and `seedProfileLedger` (the parity trigger) — beside the `inflexa profile` command actions.
- `chat_printer.ts` (441 lines) held six readers the TUI consumes beside the stdout printer that only `inflexa chat` builds. The `chat-command` spec claimed the whole file for the dev command, which was half wrong.

So the work lands in two ordered phases, one commit each, and phase 2 is a pure file move.

**Phase 1 — each shared helper goes to the module that owns its subject.**

| Helper | New home | Why there |
| --- | --- | --- |
| `describeBootError` | `harness/runtime.ts` | It is a total switch over the `HarnessBootError` union that file declares. The `never` guard makes a forgotten variant a compile error, and that guard only earns its keep while the mapping cannot drift from the union. A variant and its message are now one edit, not two. |
| `ensureSandboxImage` | `libs/pull.ts` | That module owns the sandbox image. `lib-store-provisioning`, the spec for that module, already held the requirement named after this function. The inline presence lookup collapses into the private `imagePresent` it duplicated. |
| `seedProfileLedger` | `harness/profile_trigger.ts` | Its only product consumer. It cannot go into `dev/`, because a product file must not import one. |

The `Spinner` alias was spelled twice over the same clack return value. One name now lives in `lib/cli.ts`, the repository's `@clack/prompts` boundary. `eventDepth` joins `tui/hooks/conversation.ts`, its one caller.

**Phase 2 — the dev surfaces move under `src/modules/harness/dev/`.**

`chat.ts` and `run.ts` move unchanged. The `profile` command actions become `dev/profile.ts`. The wait and status readers become `dev/status.ts`, because both dev commands call them and nothing else does. The TUI activity panel refuses those readers on purpose: the step cache records a step only when it returns, so its newest row names whatever finished last, never the work in flight.

The printer folds **into** `dev/chat.ts` rather than into a file of its own. `createChatPrinter` has one caller, and `cli/CLAUDE.md` forbids a separate file for a single-caller helper. `chat_printer.ts` keeps the five readers that two surfaces share.

`src/modules/harness/profile.ts` is gone. No file is left with a name that describes none of its residents.

**The boundary is now enforced, not only documented.** A `no-restricted-imports` rule fails the build when a product file statically imports `dev/`. It exempts the files under `dev/` and the registration gate.

Two things about that rule are worth a reviewer's eye:

- My first pattern was `**/modules/harness/dev/*`, which reads as correct and matches **nothing** from inside `modules/harness/`, where a sibling writes `./dev/chat.ts`. The rule matches the import *string*, not a resolved path. It is now `**/dev/*`, confirmed against a sibling form, a deep form, and dev-to-dev traffic.
- It covers a **static import only**. `no-restricted-imports` never visits an `ImportExpression`, so a dynamic `await import("./dev/…")` from a product file still passes. Covering it wants `no-restricted-syntax`, which replaces rather than extends per config block. Threading that through the `src/cli/**` block would cost the `.action()` ban its scope. That form is the registry's own, and a product file has no reason to write it. Thus the gap stays a review item, and the rule's comment says so.

Behavior does not change anywhere. This is a location and ownership change, plus one new lint rule.

The last commit rewrites `chat-command` and `analysis-run-launch` into STE. Every `### Requirement:` and `#### Scenario:` heading stays byte-for-byte identical. Those are identifiers that a delta matches on at archive time. Thus the counts are unchanged, at 6/21 and 5/13.

Closes #56

## Type of change

- [ ] Bug fix
- [x] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). — 2711 pass, 0 fail, across 159 files.
- [x] Tests added or updated for the change. — the existing suites split so each block sits beside its subject (`profile.test.ts` → `runtime.test.ts` + `dev/status.test.ts`, and `chat_printer.test.ts` → the readers stay while the printer blocks move to `dev/chat_printer.test.ts`). No test was lost: the count is identical before and after. **The new lint rule has no automated test.** I proved it with three throwaway probe files — a sibling import, a deep import, and legal dev-to-dev traffic — then deleted them.
- [x] Documentation updated if user-facing behavior changed. — no user-facing behavior changed. `cli/CLAUDE.md` gains "The `dev/` subdirectory", and three spec Purposes are corrected.

## What else was run

- `bun run docs:gen` emits 47 files and 23 nav entries, with no dev command leaked into the release surface.
- All three dev commands load and reach their action bodies from a source run. `profile --status` and `run --status` hit `resolveSingleAnalysis`, and `chat` hits its TTY guard. This is the one that matters: `agent_policy_tree.test.ts` walks the registry without running a handler, so it never reaches the three lazy imports. Only `tsc` and a real invocation do.
- `openspec validate --specs` passes 86/86, and the change itself validates.


vivere-dally:
@claude review

krajla (dismissed):

Issue #310, opened 2026-08-04
Report verification: the mechanical gate and a visual examination in the builder loop


vivere-dally:
**Parent:** #221

## Goal

Make the Report Builder judge its own work. #221 dropped the separate orchestrator.
Thus the judgment of the report falls to this agent and to a deterministic gate.

## Scope

- **The gate is a precondition of a finished version.** Run the mechanical validation
  before the version is recorded. The validation covers the content grammar, the
  presence of a binding, the resolution of each reference, and the match of each
  assertion.
- **Nothing half-built ships.** A version that the gate did not accept is never
  recorded. A failure removes the new version.
- **Eyes on the page.** Do a visual examination of the rendered page. The headless
  Chrome path in `harness/src/tools/report/preview-snapshot.ts` is prior art.
- **A failure names its block.** Report each failure against the `id` of the block
  that holds it. Thus a reviewer sees which block broke, and not only that the report
  broke.

## Why this issue matters

The mechanical gate is the guarantee against fabrication. A claim that binds to a
number that no artifact holds fails the gate. A claim that binds to a real
coordinate, but states a different number, also fails the gate. Neither failure needs
a model.

## Out of scope

- The semantic judgment of whether the prose follows from the value. That judgment
  needs a model, and it is advisory.

## Merge posture

**Additive and dormant.** The gate runs in the loop of the new agent, and that agent
is dormant until #314. Thus this work merges to `main` alone. Do not add the gate to
the old path.

## Depends on

- #305 — the authoring tools
- #307 — the renderer

vivere-dally:
A delta from the design discussion on #306. The gate owns the value validation. This comment records how the gate reads a value, and why.

## The validation has two tiers

- **The structural validation.** It reads the snapshot only, and it opens no file. It answers two questions. Does the path exist at T? Does the hash match? This validation is instant, thus each authoring operation in #305 can run it.
- **The value validation.** It reads the artifact, and it compares the cell to the assertion. It runs one time, at the gate, for each reference in the document together.

Thus the costly read is not per edit, and it is not per cell. It is one batch, one time for each version.

## The value validation runs in the sandbox

The harness writes one resolution script. The sandbox runs the script, and it gives back the resolved values for the whole document. The harness then compares each value to its assertion.

The harness does not parse a data file in its own process. The reasons:

- The sandbox holds the real data tools. A large table, a parquet file, and a chunked read each work already.
- No new dependency and no new parser enter the harness.
[...]

Issue #308, opened 2026-08-04
Report versions: store a version as a document, a pinned snapshot, and a time anchor


vivere-dally:
**Parent:** #221

## Goal

Make a report version a checkpoint. Store a version as three parts:

- the block document
- the snapshot that its references are pinned to
- the parent thread, and the `parent_seq` value that anchors the version in time

## Scope

- **The store.** Decide where a version lives, and how a key identifies it.
- **Reuse across versions.** Version 2 reads version 1. It carries a block forward by
  the `id` of that block. Thus a new version keeps the parts that the user liked, and
  it is not a rebuild from zero.
- **Resolution against the pinned snapshot.** A stored version resolves each
  reference against its own snapshot. Thus its numbers never drift.
- **Lifecycle.** Define what happens to a version when a person deletes the analysis
  or the parent thread. Refer to the purge behavior in #234.

## Why this issue matters

The user does more investigation in the analysis session, then asks for a new report.
The new version must show the later state. The earlier version must keep the state
that it recorded. Thus a version is a checkpoint, and it is not a live view.

## Out of scope

- The mint of a snapshot (#306).
- The context that crosses from the analysis session (#223).
- The lineage bundle of a version (#318). That work computes a new thing from the
  document and the snapshot that this issue stores.

## Merge posture

**Additive and dormant.** The store has no caller. Give it its own namespace on disk
and in the database. Do not write into the `previews/` tree that the old path owns.
Thus the two systems never share mutable state, and this work merges to `main` alone.

## Depends on

- #306 — the snapshot and the production resolver
- #235 — `parent_thread_id` and `parent_seq` are in the thread model already

vivere-dally:
Two lifecycle decisions from the design discussion behind #223.

## A version stays open for amendment

A user can return to an earlier report session and amend its version. Version 2 does not close version 1. Thus the store must let any version accept a change, not only the newest one.

## The knowledge cap

A session must know the facts and the results of the parent conversation only up to its own `parent_seq`. The cap holds for the life of the session, not only at the spawn. Thus a late amendment to version 1 must not use a fact from after its anchor.

The cap binds three surfaces:

- **The data.** A version resolves each reference against its own pinned snapshot. This rule is in the scope already.
- **The transcript.** If a tool reads the parent messages, the read must stop at the `parent_seq` of the session. It must not read the latest state.
[...]

Issue #225, opened 2026-07-24
Report Builder agent + report-session runtime


vivere-dally:
**Parent:** #221

> **Scope note.** This issue was a large one. The work between the type contract and
> a conversation with a user now has its own issues: the authoring tools (#305), the
> snapshot and the resolver (#306), the renderer (#307), the versions (#308), the
> spawn of a session (#309), and the verification (#310). Thus this issue is the
> agent only.

## Goal

Build the Report Builder agent. It is the single agent that owns a report session,
and it converses with the user.

## What the agent is

- **One agent, and it is interactive.** The user talks to it to shape the report. A
  small change, such as "use a pie chart in place of a donut chart for X", is a
  change in the same version. It is not a rebuild.
- **It sees the analysis. It cannot run the analysis.** It carries the read tools of
  the analysis: the file tree, the data profile, the inspection of a run, and the
  workspace search. Thus it goes deeper than a single brief. It cannot start a run,
  and it cannot change an analysis.
- **It authors blocks, and each claim is grounded.** It composes the report from the
  typed blocks with the authoring tools (#305). Each claim binds by reference to a
  real artifact.

## Scope

- The definition of the agent, and its registration at the type to agent seam (#236).
- The tool roster:
  - the read tools of the analysis, which exist already
  - the block authoring tools (#305)
  - the render and the preview (#307)
  - access to the design system
- The loop of the agent, and its prompt.
- The rule that no tool starts a run or changes an analysis.

## Out of scope

- The authoring tools themselves (#305).
- The renderer (#307).
- The store of a version (#308).
- The spawn of the session that the agent runs in (#309).
- The gate and the visual examination (#310).
- The context that the agent receives (#223).

## Depends on

- #222 — the contract that it authors against
- #305, #307, #309 — the tools, the renderer, and the session
- #223 — how it receives context

## Merge posture

**Additive and dormant.** Register the agent against the `report` thread type. No
code spawns such a thread until #314. Thus the registration changes no behavior, and
this work merges to `main` alone.

## Why the agent sits off the analysis conversation

The roster of this agent is large, and its context is its own. Both stay off the
analysis agent. Thus the report work never taxes the agent whose task is the
analysis. This is the reason for the whole restructure.

vivere-dally:
A delta from the #223 research. One candidate tool for the roster, deferred.

[...]

PR #327, opened 2026-08-06
feat(harness): add the block authoring tools for a report draft

### LINKED ISSUE #305: Block authoring tools: compose a report document and change it
**Parent:** #221

## Goal

Give an agent the tools to compose a typed block document, and to change one. The
contract in #222 defines what a valid document is. No code lets an agent build one.

## Why this issue connects the contract to the product

The contract landed as types (`harness/src/contracts/report-blocks.ts`) and a
mechanical validator (`harness/src/report-model/validate.ts`). It has no producer.
This issue is the join between two states:

- A valid report is definable.
- An agent can author a report.

This issue also makes an in-place change real. Each block carries a stable `id`. Thus
a request such as "use a pie chart in place of a donut chart for X" is one field
change on one block. It is not a rebuild of the page.

## Scope

- The operations that change a block document: add a block, change a block by its
  `id`, remove a block, and move a block.
- Each operation validates the result before the change lands. The validation covers
  the content grammar, the presence of a binding, and the resolution of a reference.
- A read-back operation. The agent inspects the current document. It does not keep
  the full tree in its context.
- A refused operation returns typed data. It does not throw. Thus the agent reads the
  reason, then tries again.

## Out of scope

- The agent that calls these tools (#225).
- The render of a document into a page. This is a different issue.
- The persistence of a document and its versions. This is a different issue.

## Merge posture

**Additive and dormant.** These tools reach no agent roster here. Thus this work
merges to `main` alone, and no user sees a change. The old report path stays live.
#314 holds the change that a user sees.

## Depends on

- #222 — the block and grounding contract

vivere-dally:
## What & why

The block contract and its validator landed with no producer. A valid report was definable, but no code let an agent compose one or change one. This change is the join between the two states: an agent can now author a typed block document, and it can change one block in place through its stable id. "Use a pie chart in place of a donut chart for X" is one field change on one block, and not a rebuild of the page.

A document under composition cannot obey the completeness rules of the contract, because authoring starts from nothing. Thus the change introduces a draft. **An edit refuses a fault, and the finish refuses a gap.** Each operation refuses what can never become valid: bad grammar, a duplicate id, an unresolvable reference, an unknown target. The min-1 completeness rules gate one time, in the finish operation.

### What landed

- The draft grammar (`src/report-model/draft.ts`). It composes from the contract atoms, and only the section relaxes to permit emptiness. Both unions spread one `ATOM_BLOCK_SCHEMAS` tuple from the contract, thus a ninth block kind reaches the draft the moment that the contract declares it.
- The pure operations (`src/report-model/draft-operations.ts`): add, change, remove, move, and the title. Each of the four block operations returns `Result<DraftDocument, DraftRefusal>`, and the refusal reason set is closed at nine members. A refused operation leaves the draft untouched.
- The read surface (`src/report-model/draft-read.ts`): an outline (id, kind, depth, clipped label) and one block by id. Every kind clips, the clip counts code points, and a clipped label carries a marker. An atom reads in full, and a section gives its title and the id of each child. The agent holds the outline, and it does not keep the full tree in its context.
- The finish (`src/report-model/draft-finish.ts`): the full document schema, the id rule, and the structural tier over every reference. On a pass it gives the valid `ReportDocument` value, and on a fail it gives typed gaps. It carries each advisory warning in either outcome, and it scans prose for a free numeral.
- The tool layer (`src/tools/report-authoring/authoring-tools.ts`): eight tools through one factory — `add_block`, `change_block`, `remove_block`, `move_block`, `set_title`, `read_outline`, `read_block`, and `finish_draft`. Each declares the inline execution mode, because the draft is closure memory with no durable backing. A refusal returns in the ok channel as `{ applied: false, refusal }`, and a landed operation returns the child order of each container that it changed.
- One walk of a block tree (`src/report-model/block-walk.ts`). The validator, the finish, and the operations held three copies of the same block-kind switch. A ninth kind could miss one copy in silence.

### Three decisions

**A destination is an anchor, and not an index.** Add and move name a parent and a place (`start`, `end`, `before`, `after`). An index breaks when the tree changes under the agent, but an anchor id stays stable. An anchor implies its parent, and a conflict refuses. A section cannot move into its own subtree.

**The change stamps the target id.** The operation stamps the target id on the payload, thus the id of the payload does not matter and an id mismatch is unrepresentable. A kind change ("turn this table into a chart") is one call.

**Each edit runs the structural tier only.** Per-operation validation resolves each incoming reference against the pinned snapshot, and it opens no file. The value tier stays a seam, and the report gate of #310 runs it one time for each version. Thus an edit stays instant while a document grows.

### Two notes for a reviewer

**This work is dormant.** `src/index.ts` exports no part of it, no agent roster names the tools, and the old report path stays live and untouched.

**The shared contract changes.** `src/contracts/report-blocks.ts` exports the seven atom schemas. It adds the `ATOM_BLOCK_SCHEMAS` tuple and the `AtomBlock` type, and both unions derive from them. It also tightens `ReportDocumentSchema.title` to `z.string().min(1)`.

That title rule is what makes an untitled draft a gap at the finish. The design puts each completeness rule in the document schema. Nothing in the tree produces a `ReportDocument` with an empty title.

**The draft holder is per factory.** `createReportAuthoringTools` closes over one draft and one snapshot. Two factories hold two independent drafts, and the tools read no ambient state. The persistence of a draft belongs to #308, and the agent that calls the tools belongs to #225.

### How it was verified

- 86 tests pass across the five changed areas, and the typecheck, the lint, and `openspec validate --strict` are clean.
- An independent verification pass mapped each of the 10 requirements and all 26 scenarios of the new `report-authoring` spec to code and tests, with no critical finding.
- The one warning of that pass — the same-section move resolves its destination again after the removal shifts the sibling indices — got dedicated order-asserting tests.

### The review follow-ups

Two commits answer a review of this branch. They change the surface, thus the text above describes the state after them.

`e4b4949` corrects 20 findings. The load-bearing ones:

- The block payload published the empty JSON schema. Thus the model got no grammar for the eight kinds, and it had to find each field through one refusal at a time. The payload now publishes the whole draft grammar, and its runtime parse stays permissive so that a malformed block still gives a typed refusal.
- Zod 4 treats a bare `z.unknown()` as a required key. Thus `change_block` rejected the section retitle call that its own description asks for, and no agent could retitle a section.
- No operation could set the title of the document, and the finish accepted an empty one. `set_title` and the `min(1)` rule close that.
- The duplicate-id scan read the whole draft. Thus one duplicate that the draft already carried refused every later operation, even an operation on an untouched branch. The scan now reads the payload of the operation.
- A destination that named two places refused with `unknown-target`, which sent the agent to look for an id that was already correct. It refuses with `conflicting-destination`.
- The outline clipped a text and a claim only, it cut by UTF-16 unit, and it left a clipped label unmarked. A title, a caption, and a note rode unclipped, and an astral character could split into a lone surrogate.

`9049cc6` fixes the cost of a landing. Each landed operation returned the outline of the whole draft, thus authoring n blocks cost n-squared outline entries of agent context. A report of 10 sections and 100 atoms spent 341275 bytes of tool results, against a final outline of 6103 bytes. A landing now carries the child order of each container that it changed, and the same report costs 39825 bytes. `read_outline` stays the whole-draft view, one call away.

Closes #305

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed areas pass, which is 86 tests in `src/report-model/` and `src/tools/report-authoring/`. The full suite did not run locally, because it starts one Postgres container for each file.


vivere-dally:
@claude review

krajla (approved):

Issue #329, opened 2026-08-07
Explore richer chart customization for the report blocks


vivere-dally:
## Context

#307 gives the new report renderer one fixed derivation rule for each chart type. The rules came from a study of echart's flexibility and extensibility, which sets the bar for per-type customization.

To stay deterministic and grounded, #307 took three narrowing decisions:

- The renderer computes no aggregate. A pie or heatmap table arrives pre-aggregated, with one row for each category or cell.
- The histogram bin count is automatic: the larger of the Sturges count and the Freedman-Diaconis count. There is no bin knob.
- The chart encoding carries column names only (`x`, `y`, `group`, `value`). There is no operator field.

## The gap list

The reference dialog exposes knobs that have no home in the block contract today:

- An aggregation operator (`{ col, agg }` with `SUM`, `COUNT`, `COUNT_DISTINCT`, `MEAN`, `MAX`, `MIN`, `AVG`) for pie, histogram, and heatmap.
- A histogram bin-count knob (a 3 to 30 slider in the reference).
- The generalized histogram mode: bin one column, then aggregate a second column over each bin.
- Four chart types outside the contract: violin, volcano, manhattan, and MA. Each carries its own knobs, such as threshold sliders and a control-perturbation pair.

## What to explore

- Do report authors need these knobs, or does the pre-aggregated rule cover the real cases?
- If a knob earns its place, the change lands in the block contract (`ChartTypeSchema` and `ChartEncodingSchema` from #222), and the renderer grows with it.
- The volcano and manhattan types are genomics-native. Weigh them first, because our domain asks for them.

## Relations

- Inspired by the derivation study of #307 and the echart's research.
- This does not block #221. The rebuild delivers with the seven fixed types.

Refs #221, #307.

Issue #221, opened 2026-07-24
Rebuild report generation from scratch: a dedicated, interactive, grounded report-building session


vivere-dally:
> **Supersedes #197.** This from-scratch rewrite replaces the incremental improvement plan in #197, which is now closed. #197 remains the best diagnosis of the current subsystem and its gaps. The work moves here.

## Summary

Rebuild the harness **report-generation subsystem from scratch**. The current system works, but its *ceiling* is the problem: it produces simplistic, weak reports. This is not an incremental-patch effort. [#197](https://github.com/inflexa-ai/inflexa/issues/197) diagnosed the current system in depth, and it is the prior art behind this work. But the goal here is a new design, and not a fix list.

> **Status:** vision / tracker. This is a living document. The structure and subissues below are a starting point to be refined together — expect them to change.

## Why rewrite, not patch

Report generation is a **core part of the product**, and today it under-delivers. The reasons are structural, not cosmetic:

- **One brief, one turn.** The analysis agent composes a report brief as a side-quest while doing analysis. Report craft gets a fraction of one turn's attention, and the report's ceiling is whatever fit in that moment.
- **A blind, under-powered builder.** The builder cannot see the analysis, cannot explore the data, and (in OSS) cannot see the page it wrote. It can only assemble a fixed, narrow set of section types from one skill pack.
- **No human-in-the-loop on the rendered artifact.** Refining a report means diffing briefs, not looking at the report and saying "go deeper here."
- **No grounding.** Nothing ties a number in the report to a real analysis artifact. An agent transcribes every figure from its memory.

## The vision

Report-building **leaves the analysis conversation** and becomes its own **child session** owned by a single **Report Builder agent**.

- **A dedicated, interactive session.** The user talks to the Report Builder to shape and amend the report. "Use a pie chart instead of a donut for X" is an in-place amend within the same version, not a full rebuild.
- **The Report Builder can see the analysis.** It gets the analysis-*read* tools (full file tree, data profiling, run inspection, workspace search) — but **cannot start analysis runs** or do anything analysis-related. This is what raises the richness ceiling: it can go deeper than any one-shot brief.
- **This keeps the analysis agent clean.** The expanded report capability (many tools, its own context) lives on the child session, off the analysis agent, so it never taxes the agent whose job is the analysis.
[...]

PR #330, opened 2026-08-07
feat(harness): render a block document into a deterministic HTML page

### LINKED ISSUE #307: Deterministic renderer: turn a block document into an HTML page with no agent-written markup
**Parent:** #221

## Goal

Render a block document into an HTML page with a deterministic function. No agent
writes markup.

## Why this issue removes a class of defect

Today the builder agent writes a Nunjucks template. The gate then examines the result
for an unrendered marker, an asset that does not resolve, and an empty page
(`harness/src/tools/report/submit-report.ts`). Each of those checks exists because an
agent can write bad markup.

Make the layout a pure function of the block tree. Then there is no template for an
agent to get wrong, and the full class of defect goes away. An agent makes the
decisions about meaning. The renderer makes the decisions about layout.

## Scope

- A renderer. Its inputs are the block document and the values that resolution gives.
  Its output is an HTML page.
- Each value in the page comes from the resolution of a reference. No value comes
  from prose. Thus a number on the page is always a value that the renderer read from
  an artifact.
- A rendered form for each of the eight block kinds. Reuse or replace the component
  set in `templates/report-html`.
- A deterministic output. The same document and the same snapshot give the same
  bytes.

## Out of scope

- The visual quality and the design system. A different issue covers that work. This
  issue covers the mechanism only.

## Merge posture

**Additive and dormant.** The renderer is a function with no caller. The old render
path (`harness/src/execution/report-render.ts`) stays live and untouched.

**One rule protects the old path.** Give the new renderer its own directory for its
templates and its assets. The old path reads `templates/report-html` at run time. If
the new renderer shares that directory, then #311 changes the output of a report that
a user makes today. A separate directory is cheap now, and expensive later.

## Depends on

- #222 — the block contract
- #305 — the document shape that the authoring tools produce

vivere-dally:
## What & why

Today the builder agent writes a Nunjucks template, and the gate examines the result for symptoms: an empty page, an unrendered marker, and an asset reference that does not resolve. Each check exists because an agent can write bad markup. This change adds the renderer of the report rebuild: a block document becomes one HTML page through a pure function, and no agent writes markup. Thus the whole defect class becomes unrepresentable, and the gate has nothing left to police.

### What landed

- `renderReportPage(document, values)` in `src/report-render/`. The inputs are the valid `ReportDocument` from `finish_draft` and the resolved values keyed by block id. The output is one HTML string on the `Result` channel. A missing or mis-shaped value returns collected typed problems, and the render never throws for an expected fault.
- A rendered form for each of the eight block kinds. The forms: nested section headings with a left sidebar, text, claims with evidence markers, metric stat cards, full tables, figures, citations, and charts. A shared ledger numbers the evidence by first appearance, and it dedupes the reference list.
- The chart derivation for the seven chart types, each option normalized through `normalizeEchartSpec`, thus the `echart-layout` discipline applies without a second implementation.
- The page is one self-sufficient string. The theme CSS inlines in a `<style>` block, and the scripts and the fonts ride the pinned CDN references with SRI hashes. No local asset reference exists in the output.
- Escaping is always on, and the renderer owns it. The two raw JSON sinks (the chart options and the theme) replace each `<` in the serialized text, thus a `</script` inside a data cell cannot close the element.

### Three decisions

**The renderer computes no aggregate.** A pie or heatmap table arrives pre-aggregated, one row for each category or cell, and a repeated category refuses. Thus every plotted number stays traceable to a cell of the pinned evidence. The one sanctioned computation is presentation math: the histogram bins with the automatic rule (the larger of the Sturges count and the Freedman-Diaconis count), and the box plot computes type-7 quantiles with Tukey fences. The derivation rules mirror our dashboard chart dialog, and #329 tracks the knobs that stay deferred.

**The caller adapts the values.** The renderer takes a closed `RenderValue` union, not `ResolvedValue`. A file echo of path and hash is not showable. Thus the caller computes the figure source (a data URI, or a URL that the host serves). The renderer holds no policy about where image bytes live.

**Determinism is a rule, not an aspiration.** No clock, no random value, no locale call, code-unit sorts, and first-appearance ordering everywhere. The same document and the same values give the same bytes, and a byte test pins it.

### Two notes for a reviewer

**This work is dormant.** `src/index.ts` exports none of it, no roster names it, and the old render path (`src/execution/report-render.ts`, `templates/report-html/`) stays live and untouched. The renderer never reads the template directory, and its constants are copies with byte-identical SRI hashes.

**The value tier stays a seam.** The renderer trusts its value map. #310 realizes the value tier and feeds it, and #308 stores the rendered string as part of a version.

### How it was verified

- 54 tests pass across the six test files, and the typecheck and the lint are clean.
- An independent verification pass mapped all 9 requirements and 22 scenarios of the new `report-render` spec to code and tests. It recomputed the histogram edges and the box summary by hand from the fixtures. It traced every interpolation site for an unescaped sink, and it swept for nondeterminism (clock, random, locale, Set and Map iteration). No critical and no warning finding.
- The OpenSpec specs are in this pull request, synced and archived.

Closes #307

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed area pass, which is 54 tests in `src/report-render/`. The full suite did not run locally, because it starts one Postgres container for each file.


harness/src/report-render/escape.ts
vivere-dally, line 1: Is it wise to do this ourselves? Isn't there a library that can be used that is battle-tested?

harness/src/report-render/page.ts
vivere-dally, line 1: Is there a library that can validate the html & css that we have inline as raw strings are valid?
Validating would assure us that future updates are not breaking the rendering

harness/src/report-render/escape.ts
vivere-dally: The battle-tested option here is `escape-html`, and its whole body is the same five entity replacements as this file. Thus a swap adds a runtime dependency to a published library, and no new behavior. The risky half of escaping is where it applies, and no library owns that — the tests pin each interpolation site. The two dangerous sinks are JSON inside a script element, and they take the `\u003c` rule. That rule stays hand-rolled under any library, because entity escaping corrupts the JSON. Thus the file stays small, owned, and fully traced.

harness/src/report-render/script-json.ts
vivere-dally, line 1: Please make this a JSON extension, similar to how it's done in `cli`

vivere-dally: The extension pattern fits the cli, and it would be wrong here, because the harness is a published library. A global `JSON` method mutates the runtime of every consumer, the `declare global` leaks into the published types, and the method exists only when the side-effect module loads — and a library has no entry point to anchor that load. The same posture already keeps a logging library out of the harness. The cli doc also reserves an extension for a cross-cutting helper, and `scriptJson` has two call sites in one subsystem. The narrow door is also the point: the `\u003c` rule is correct only inside a script element, and a global method invites it where it is wrong. Thus the helper stays a module function beside its invariant.

krajla:
Doesn't this remove the need for the templates and allow us to delete them?

vivere-dally:
> Doesn't this remove the need for the templates and allow us to delete them?

@krajla That is correct. Removal will be done in a later issue. #221 centralizes all of them, and there are one or two issues that are preparing the system for production use.

The reasoning: we want to keep merging pieces of the system to main - without actually using them - until we do the "switch". It was either this, or a separate long-running feature branch, that had to be constantly rebased with main.

krajla (approved):

Issue #318, opened 2026-08-04
Report lineage: derive a provenance bundle for a report version


vivere-dally:
**Parent:** #221

## Goal

Give each report version a lineage bundle. For each claim, each figure, and each
table, the bundle answers one question: which data produced this, and how?

## The two halves exist already

The lineage has two halves. Each half is in the tree today.

- **The last hop.** A `Reference` binds a claim to one artifact. It pins a `path` and
  a content `hash` (#222).
- **Each hop before it.** A `ProvenanceRecord` records how the harness made an
  artifact (`harness/src/provenance/types.ts:60-77`). It holds `outputPath`,
  `outputHash`, `producer`, `scriptPath`, `stepId`, `runId`, and an `inputs` array.
  Each entry in `inputs` holds a `path`, a `hash`, and a `source`.

**The two halves join on one key.** A reference pins `path` and `hash`. A provenance
record keys on `outputPath` and `outputHash`. These are the same coordinates. Thus no
adapter is necessary, and no new identifier is necessary.

## The walk

Start at a claim. Then do these steps again for each input:

1. Read the `path` and the `hash` of the reference.
2. Find the provenance record with the same `outputPath` and the same `outputHash`.
3. Record the producer: the script, the step, and the run.
4. Take each entry in `inputs`, and do the walk again.

The walk stops at an entry with a `source` of `data`
(`harness/src/provenance/types.ts:37`). That entry is a raw input file, and it carries
a `fileId`.

One chain then holds each of these parts:

- the sentence in the report
- the cell that the sentence cites
- the file that holds the cell
- the script that wrote the file, with its step and its run
- the data that the script read

## Scope

- **Compute the bundle. Do not author it.** The two inputs are the block document and
  the pinned snapshot. Both exist at the moment that a version is cut (#308). Thus no
  person maintains the bundle, and the bundle cannot drift away from the report.
- **Store the bundle with the version.** The bundle is a function of a frozen version.
  Thus the bundle is as frozen as the version.
- **Give the bundle a shape.** Decide how it records one chain, and how a reader finds
  the chain of one block by the `id` of that block.

## The questions to answer

- **The size of the closure.** A late artifact in a deep pipeline reaches many
  upstream files. Set a policy for the depth, or a rule that summarizes. A bundle that
  no person opens has no value.
- **The analysis scope.** A `Reference` carries no `analysis_id`. But `cortex_artifacts`
[...]

Issue #309, opened 2026-08-04
Spawn a report session: the harness operation that makes the child thread


vivere-dally:
**Parent:** #221

> **Scope note.** This issue holds the mechanism only. The conversation agent does
> not gain the operation here. That change is the switch, and #314 holds it. The
> reason for the split is in the merge posture below.

## Goal

Give the harness the operation that makes a child report thread. No code makes a
thread of the type `report` today.

## The parts that exist already

The data model and the agent lookup are in place. Two changes landed:

- `thread_type`, `parent_thread_id`, and `parent_seq` are columns on
  `cortex_analysis_threads` (`harness/src/state/init.ts:283-292`), from #235.
- `ThreadAgentResolver.forThread(type)` selects the agent from the type of a thread
  (`harness/src/runtime/assemble.ts:121-138`), from #236.

`report` is a permitted thread type. No agent is bound to it, and no code path makes
such a thread. This issue adds the operation.

## Scope

- **The spawn operation.** It makes a child thread. The child carries a `thread_type`
  of `report`, the `parent_thread_id` of the analysis thread, and a `parent_seq`
  value.
- **The time anchor.** Set `parent_seq` to the `messages.seq` value of the parent at
  the moment of the spawn. This value anchors the version in time.
- **The list of children.** Give an operation that answers one question: which report
  versions belong to this analysis?
- **Harness-first.** A managed deployment gets this behavior with no CLI user
  interface.

## Out of scope

- The tool that lets the conversation agent call the spawn (#314).
- The agent that runs in the session (#225).
- The navigation between the two sessions in the CLI (#312).
- The context that crosses the link (#223).

## Merge posture

**Additive and dormant.** The operation has no caller. Thus this work merges to
`main` alone, and no user sees a change. The old report path stays live.

## Depends on

- #235 — the parent and child columns
- #236 — the type to agent seam

vivere-dally:
A delta from the #223 research. The spawn operation carries the context transfer, in three parts.

- **The payload.** The spawn takes a brief as its argument. The conversation agent authors it at the moment of the ask. It holds the objective, the audience, the angle, the exclusions, and open questions for the user. Keep the brief below approximately 2000 tokens.
- **The memory copy.** At the spawn, copy the working-memory render into the child context. The working memory is one mutable row per analysis (`cortex_working_memory`). A live read later sees state past the anchor, and that breaks the knowledge cap on #308.
[...]

Issue #306, opened 2026-08-04
Report snapshot: mint a pinned artifact set and resolve a reference against real storage


vivere-dally:
**Parent:** #221

## Goal

Make the grounding real. Today the `ReferenceResolver` seam has one realization, and
it reads an in-memory fixture map (`harness/src/report-model/fixture-resolver.ts`).
No code resolves a reference against a real analysis artifact. Thus the guarantee
against fabrication is a demonstration, and not yet a product behavior.

## Scope

- **Mint a `ReportSnapshot` for an analysis at a point in time.** The snapshot holds
  the set of pairs of a path and a content hash. The data comes from
  `cortex_artifacts`. The snapshot also holds what a locator uses to read a cell.
- **Add a production `ReferenceResolver`.** It reads the workspace tree and the
  ledger. It returns the same typed outcomes that the fixture resolver returns:
  `artifact-missing`, `hash-mismatch`, `locator-out-of-range`, `ambiguous-match`, and
  `assertion-failed`.
- **Decide how to read a tabular artifact for one locator.** A full parse of a large
  table is expensive. Set a bound on the cost.

## Why this issue matters

This is the mechanism that freezes a report version at a point in time. A version
resolves each number against the artifacts as they were when the version was cut.
Thus a version 1 report shows the same numbers forever, and a later run does not
change them.

## Out of scope

- The storage of a snapshot with a report version. This is a different issue.
- The narrative part of the context transfer (#223).

## Merge posture

**Additive and dormant.** The new resolver has no caller. The fixture resolver keeps
its place in the tests. Thus this work merges to `main` alone, and no user sees a
change.

## Depends on

- #222 — the reference shape and the resolver seam

vivere-dally:
A delta from the #223 research, in two parts.

**Mint the snapshot at the spawn of the report session, not at the version cut.** Then `parent_seq` and the snapshot describe the same moment, and the version of #308 records one consistent T. The session amends its version against that one moment, per the knowledge cap on #308. #309 holds the spawn operation.

**One storage fact records the why of the copy.** The registry (`cortex_artifacts`) keeps one row per path, and an upsert replaces the hash in place. Thus a pin alone cannot re-read the past, and the snapshot must copy what resolution needs. The scope of this issue does this already.

vivere-dally:
A correction to the scope in the body above. The body says "Add a production `ReferenceResolver`. It reads the workspace tree and the ledger." The design discussion moved that work.

## What this issue holds now

[...]

PR #331, opened 2026-08-07
feat(harness): store a report version as a document, a snapshot, and an anchor

### LINKED ISSUE #308: Report versions: store a version as a document, a pinned snapshot, and a time anchor

vivere-dally:
## What & why

A report version is a checkpoint, but nothing recorded one. The producers landed earlier: `finishDraft` gives the valid document (#305), `mintReportSnapshot` gives the pinned artifact set (#306), and the report thread carries the parent anchor (#235). This change is the store between them. It records the triple — the document, the snapshot, and the anchor — and it computes none of them.

The lifecycle rule carries the most weight. **A thread is a scaffold, and a version is a deliverable.** The user deletes a report chat, and a shipped report must survive that. Thus the table has no foreign key to `cortex_analysis_threads`, and the anchor is denormalized onto the row. The versions of an analysis die only in `purgeAnalysis`.

### What landed

- `cortex_report_versions` in the state DDL: the document and the snapshot as JSONB, the anchor columns, a per-thread ordinal with a unique index, and a parent version link with `ON DELETE SET NULL`.
- `createReportVersionStore` (`src/state/report-versions.ts`): `record`, `getVersion`, `getLatestVersion`, and `listVersions`. The store surface holds no update path, thus a recorded version is immutable and a correction is a new version.
- The record parses the document against `ReportDocumentSchema` before the insert, and it refuses a malformed value as typed data with no row. The error union distinguishes the malformed document, the foreign parent, the corrupt row, and the database fault.
- The purge entry in `ANALYSIS_KEYED_DELETES`, ahead of the `cortex_analysis_state` delete, with the foreign-key cascade as the backstop only.
- The `analysis-purge` spec footprint now lists the table, and the new `report-versions` spec records the seven requirements.

### Three decisions

**The snapshot is per version, and the store never mints.** A new version must show the later state of the analysis, and an old version must never drift. Thus the caller mints at the anchor of the version, and the store stores the value as given. A later ledger write changes nothing in a recorded version.

**The ordinal computes inside the insert.** `COALESCE(MAX(version_number), 0) + 1` over the thread, under the unique index on `(thread_id, version_number)`. A lost race retries the insert one time with a fresh id, and a second loss surfaces the typed database error. The retry test drives a real lost race with an uncommitted competing transaction, not a sleep.

**The parent must belong to the same analysis.** The lineage bundle of #318 joins claims to provenance through the parent link, and a cross-analysis parent would corrupt that chain. The record refuses the mismatch as typed data, and the self foreign key refuses an unknown parent id.

### Two notes for a reviewer

**This work is dormant.** `src/index.ts` exports no part of it, nothing outside `src/state/` imports the store, and the old report path stays untouched. #309 (the session spawn) becomes the first caller.

**The reads parse on read.** A read gives the parsed `ReportDocument` and `ReportSnapshot` values, thus the stored bytes never flow onward unchecked. A row that fails the parse reads as a typed `corrupt_version` error that names the version id and the failed part.

### How it was verified

- The store and the purge suites run against a real Postgres, beside a clean typecheck and a clean lint.
- An independent verification pass mapped all 8 requirements and all 17 scenarios of the two specs to code and tests. All 7 design decisions hold, and there is no critical finding.
- The purge rig pins the table through the information schema, thus a future analysis-keyed table cannot silently escape the purge assertion.

Closes #308

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed areas pass against a real Postgres (`src/state/report-versions.test.ts`, `src/state/purge-analysis.test.ts`). The full suite did not run locally, because it starts one Postgres container for each file.


vivere-dally:
@claude review

krajla (dismissed):

krajla (approved):

PR #332, opened 2026-08-07
refactor(cli): adopt the prov-kernel dialect and delete the local copies

krajla:
## What

The CLI's provenance implementation moves onto the published `@inflexa-ai/prov-kernel@^0.3.0` and deletes its local copies of the dialect: `document.ts` 583→56, `signing.ts` 274→176 (key-file lifecycle only), `verify.ts` 265→124 (storage reads + command wiring over kernel verify/sidecar), `types/prov.ts` 243→31 (kernel re-exports + identity conventions), and the recorder's nine-case event switch becomes one `applyProvEvent` call. Net **−1008 source lines** in the prov area (−538 including the new fixture and compatibility tests). The recorder lifecycle (SQLite snapshots, dirty tracking, single-flight flush, rehydration), bus delivery, lineage read-side, and CLI commands are unchanged.

## Continuity (existing local documents stay stable)

- **Digest**: the kernel model is constructed with the CLI's historical `Bun.hash(s).toString(36)` digest injected; a test pins QNames against literals captured from the pre-kernel implementation.
- **User agents**: the kernel keys user actors on `id`; the CLI passes `id` = the email it always keyed by (plus the optional `email` attribute), so agent QNames are byte-identical.
- **Compatibility fixture**: generated by the pre-kernel code before deletion. The kernel `verifyProvenance` accepts it, `loadDocument` rehydrates it, and replaying the same events dedupes into it (structurally equal modulo serializer blank-node numbering on anonymous lifecycle relations).
- **Accepted forward deltas** (documented in the openspec design): empty command args no longer emit `inflexa:args: ""`; the sidecar gains an unset optional `kid`; a verify-result wording change. None affect stored bytes.

## Notes

- OpenSpec change `adopt-prov-kernel` validates strict; the one `--all` failure (`usage-breakdown` spec) is pre-existing on main and untouched here.
- The full-suite failures on this branch are byte-identical to origin/main's (stash-verified): the published harness 0.17.0 lacks `normalizeDetail`/`DETAIL_MAX_LENGTH` — the documented red-on-main window that PR CI's local-harness link closes.
- Independent of inflexa#326; the only shared cli file there is `run_deps.test.ts`, untouched here.


cli/src/modules/prov/lineage.ts
vivere-dally, line 1: Maybe this can be also part of the prov-kernel package.

cli/src/modules/prov/verify.ts
```diff
@@ -1,121 +1,32 @@
...
+import {
+    buildSidecar as buildSidecarWithSigner,
+    createKeypairSigner,
+    formatVerifyResult,
+    sidecarSchema,
+    verifyProvenance,
+    verifySidecar,
+    type ProvSigningError,
+    type Sidecar,
+} from "@inflexa-ai/prov-kernel";
```
vivere-dally, line 12: The sidecar name is terrible, I had to ponder for a while to remember what it is. Maybe make a issue for me to do some refactoring there or you could do it if you have the energy for it

vivere-dally (commented):
I'd wait for your answers first, before I'll approve or not.

In addition, I'd like to ask you to test that this works:
1. Run an analysis (full run)
2. Export the prov document via the TUI
3. Verify them
4. Try to alter the file and verify again - it should fail
5. Check the lineage of any file from that run

I know that tests may be passing, but they've been updated as well. An e2e manual test would be good

cli/src/modules/prov/lineage.ts
krajla: I'll have a look why it didn't consider this "generic" enough

krajla: Moved the lineage code to the kernel, this makes the CLI and lumen a consumer since they interpret the provencence.

cli/src/modules/prov/verify.ts
krajla: Renamed to attestation

krajla:
@vivere-dally Ran an analysis locally and the provenance registered and exports and verifies correctly.

vivere-dally (approved):

PR #336, opened 2026-08-07
feat(harness): enforce one report version for each thread

vivere-dally:
## What & why

The policy on #221 binds one report version to each report session, and until now only prose held it. The version store still carried the multi-version machinery: the per-thread ordinal, the max-plus-one computation, and the race retry. A dormant capability invites use without the vision, and nothing published carries the table. Thus this change deletes the machinery and makes the database enforce the rule.

The `version_number` column and its unique index leave the DDL, and a named UNIQUE constraint on `thread_id` replaces them. The record writes no ordinal, and a second record for the same thread refuses as the typed reason `thread_already_holds_version`, mapped from the constraint name. Two concurrent first records race, and the loser gets the same refusal, because the thread holds a version by then.

The read surface now teaches the enforced shape. `getThreadVersion` replaces `getLatestVersion`, because "latest" implies a set. `listVersions` dies, because a list of at most one element teaches a false shape. `getVersion` by id and the cross-session `parentVersionId` link stay unchanged.

### Notes for a reviewer

- The store stays dormant: nothing calls `record`, and `src/index.ts` exports none of it.
- A development database made before this change holds the old table shape, and the init DDL does not alter an existing table. Drop `cortex_report_versions` there one time.
- After the merge, the policy comment on #221 gets a correction, because its "stays in place and unused" sentence goes stale with this change.

An independent verification pass mapped the added, modified, and removed requirements to code and tests with zero findings. The second-record test asserts the surviving row count against a real Postgres, and a violation of a different constraint stays a plain database error.

Serves the policy locked on #221.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed area pass against a real Postgres. The full suite did not run locally, because it starts one container for each file.


radutanasa (approved):

vivere-dally:
@claude review

Issue #314, opened 2026-08-04
Switch the report capability of the conversation agent from the brief tools to a report session


vivere-dally:
**Parent:** #221

## Goal

Change the report capability of the conversation agent. It stops the offer of the old
brief tools. It starts the offer of the report session.

This is the one change that a user sees. Each other issue in the rebuild is additive,
and it changes no behavior.

## Why this change is separate

Two report paths in one roster are a defect. The old pair `plan_report` and
`submit_report` builds a report in the analysis conversation. The new path spawns a
report session (#309). If both are in the roster at the same time, the agent selects
between them, and the selection is not predictable.

Thus the removal from the roster and the addition to the roster are one atomic
change. They cannot land in two parts.

## Scope

- Remove `plan_report` and `submit_report` from the roster of the conversation agent.
- Add the tool that spawns a report session, and give it the operation from #309.
- Bind the Report Builder agent (#225) to the `report` thread type at the seam from
  #236.
- Update the prompt of the conversation agent. It must describe one report path, and
  not two.

## The code stays

This change removes nothing from the source. It changes a roster and a prompt only.
Thus a revert is one commit. #313 removes the old code later, after this change
proves itself.

## Merge posture

**This is the switch. It is the only user-visible change in the rebuild.**

Land it after each dormant part is on `main` and each part works. Two options control
the risk:

1. Land the switch as its own small pull request. A revert restores the old roster.
2. Add a narrow experimental entry point first. The old path stays the default. Then
   the flip is a one-line change of that default.

Option 2 is better if you want to use the new path on `main` before the flip. Keep
the decision in the harness. An embedder must not select which capability exists.

## Depends on

- #305, #306, #307, #308 — the parts that the new path uses
- #309 — the spawn operation
- #225 — the agent that the session runs
- #310 — the gate that a version must pass

vivere-dally:
## One more flip precondition

The report page loads its runtime from sibling `assets/` files (#311, PR #353). In the compiled binary the copy source exists only after #312 packs the assets and binds `resolveReportPageAsset` — refer to the comment there. Thus the switch must not flip before that binding ships. Without it, `preview_report` fails with `write-failed` on a user machine, and the report session cannot produce a page.

PR #340, opened 2026-08-10
feat(harness): add the report-session agent over a durable session state

### LINKED ISSUE #225: Report Builder agent + report-session runtime

vivere-dally:
## What & why

No agent served a `report` thread, and `agents.forThread("report")` refused. This change adds the Report Builder agent — the single interactive owner of a report session — and the runtime that carries its state. The agent reads the analysis, composes the report from typed blocks, and grounds each claim through a reference. It cannot start a run, and it cannot change an analysis. The resolution seam returns singleton agents, and chat runs one replica for each turn. Thus the per-session state lives in one durable row, and not in process memory.

### What landed

- One row for each report thread (`cortex_report_session_state`) holds the in-progress document and the pinned snapshot. The snapshot mints one time, thus a post-anchor artifact never joins the evidence set.
- `createReportAuthoringTools` takes a session-state gateway in place of the closure holder, and a landed document persists before the tool reports `applied: true`.
- The session runtime realizes the gateway, and its idempotent state operation mints at the first use.
- The `preview_report` tool finishes the draft, resolves, renders, and writes the page into `report-sessions/{threadId}/`. The result carries the page path, because `PreviewPublisher` mints hosted access only, and it carries no page.
- The agent `report-session`, its prompt, and its registration under the `report` thread type.

### Two notes for a reviewer

- The work is dormant. No production path spawns or serves a `report` thread until #314. The old report path stays live and untouched, with the old `report-builder` id in place.
- One spec obligation is deferred on purpose. The serving path must run the state operation at the turn start, and no serving path exists yet. Each tool call anchors the mint through `gateway.load`, and #314 owes the turn-start wiring with its test.

Closes #225

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed areas pass, across `src/state/`, `src/tools/`, `src/report-render/`, `src/app/`, `src/agents/`, and `src/runtime/`. The full suite did not run locally, because it starts one Postgres container for each file.


vivere-dally:
@claude review

radutanasa (approved):

Issue #346, opened 2026-08-10
Sandbox lease: a warm machine that runs queued commands with durable completion


vivere-dally:
**Parent:** #221
**Blocks:** #310

## Goal

Give the harness a warm sandbox machine that waits for commands. A caller leases the machine, submits a script with its input, and gets a typed result. The machine stays alive through a quiet window, and then it closes.

## The vision

- The machine runs in a wait-for-commands mode, and no agent runs inside it.
- A queue of command items. An item is a script with its input. The machine pops an item, runs it, and gives the result back.
- An idle window of 30 to 60 seconds. A machine with an empty queue closes when the window passes. A later command gets a fresh machine.
- Durability through DBOS. Each command is one durable record. A command completes only when the sandbox gives its response. If the machine dies mid-command, the record stays open, and the command runs again on a fresh machine.

## The rails exist

The exec protocol already separates the machine lifetime from the command lifetime. `POST /exec` is idempotent on `execId`, thus one machine takes many commands (the harness-sandbox-exec spec). `awaitExec` polls in durable steps, thus a host restart loses no command. The missing part: a machine that outlives one command. A lease manager gives the warm machine, and an idle reaper closes it.

## Scope

- The lease manager, the idle reaper, and the durable command record.
- The submit of a caller script with its input, and the typed result of the run.
- The confinement model stays as it is.

## Out of scope

- The report resolver that consumes the lease. That consumer is #310, and it submits only a fixed harness-authored script.
- An agent loop inside the machine. A leased machine runs scripts, not a model.

## Merge posture

Additive and dormant. The lease gets its own spec and its own change. No production caller reaches it until #310 lands on top.

vivere-dally:
Gave up on the idea after talking with @krajla - there are questions like who pays for the machine. It can be done, but it's too complex for our current use case. Might come back to it later

PR #347, opened 2026-08-11
feat(harness): gate a report version on the value tier and the eyes

### LINKED ISSUE #310: Report verification: the mechanical gate and a visual examination in the builder loop

vivere-dally:
## What & why

No code realized the value tier, and nothing gated a recorded version. A claim could bind a real coordinate and state a wrong number. This change adds #310: the resolver, the gate, and the eyes.

### What landed

- The production resolver, three layers behind the one seam. The identity compare refuses before any parse. The host fast path parses CSV, TSV, JSON, and parquet (`hyparquet`) under a 16 MiB cap. Every doubt falls through to the sandbox arm.
- The extraction workflow on the data-profile rails. The fixed Python script rides as the argv of one exec, and no agent runs in the container. One submission covers one pass.
- The record tool: the full gate runs before the append-only store, each failure names its block, and only the gated document value records.
- The look-before-record rule: the preview stamps the rendered hash, and the eyes copy it in SQL. A never-seen or stale page cannot record. The judgment stays advisory.
- The `examine_page` tool: a `file://` navigation of headless Chrome, with the screenshot, the console errors, and the failed requests.
- The wiring: a per-analysis resolver factory, and a new prompt section with its anti-pattern entry.

### Three notes for a reviewer

**A warm lease lost to a one-shot run.** #346 explored a warm machine, and the payer question closed it. The extraction run bills and dies like a profile run.

**One embedder-facing type change.** `CoreRuntimeDeps` drops `reportReferenceResolver`, and `reportHostReadCapBytes` replaces it. The cli meets this at the pin bump.

**The Python script never ran in a real container.** The tests pin its argv and its output protocol. The package-boundary route executes it later.

### How it was verified

An independent examination mapped each spec requirement and each scenario to code and tests, and its warnings got tests. The gates pass against a local Postgres.

Closes #310

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint and typecheck pass locally. The tests of the changed areas pass. The full suite did not run locally, because it starts one Postgres container for each file.


harness/src/report-model/draft-hash.ts
vivere-dally, line 1: Is this `canonicalize` necessary? How many callers does it have? Just `computeDraftHash`? Can't we use zod or some other library?

harness/src/report-model/production-resolver.ts
vivere-dally, line 1: Shouldn't we better use a library for csv/tsv? Something like https://github.com/data-forge/data-forge-ts or if not that, https://github.com/adaltas/node-csv or papa parse?
Also, should we really trust the ext of the files? What about mime types?

harness/src/report-model/production-resolver.ts
```diff
@@ -0,0 +1,784 @@
...
+ * time, and fills the cache. `resolve` then answers from the cache. A `resolve` with no prior `prepare`
+ * still answers, at the cost of a per-reference read.
+ */
+export function createProductionResolver(deps: ProductionResolverDeps): ReferenceResolver {
+    const { workspaceRoot, analysisId, extractionArm } = deps;
+    const cap = deps.cap ?? DEFAULT_CAP_BYTES;
+
+    // The per-pass caches. `prepare` clears and refills them, thus a later pass reads the artifacts again.
+    const idCache = new Map<string, Identity>();
+    const cellsCache = new Map<string, Cells>();
```
vivere-dally, line 677: Here we should use a ttl cache.
If we're in the managed deployment, this will grow forever, no?

harness/src/report-model/draft-hash.ts
vivere-dally: It is load-bearing, with one caller. The `document` column is JSONB (`init.ts:189`), and Postgres jsonb stores keys in its own physical order. Thus a draft can stamp before a round trip and compare after one, and without the sort the rule breaks with false stale-look refusals. `computeDraftHash` is the one caller, and the two stamp sites call it (`preview-report.ts:322`, `record-version.ts:146`). Zod parses, and it does not canonicalize. A stable-stringify dependency does exactly these twelve lines, and the draft is JSON-safe by construction because zod parsed it. The call: keep the twelve lines, and take no dependency for them.

harness/src/report-model/production-resolver.ts
vivere-dally: The strictness is the feature. Papa Parse and node-csv are permissive by design: they guess a dialect. A host guess that differs from pandas is the two-semantics risk that the design excludes, thus the host parser refuses on any doubt and falls through to the sandbox, where pandas reads the file with the engine that wrote it. The extension is a routing hint only, and never a correctness input. A mislabeled file fails its parse, and the fall-through absorbs it — a wrong route degrades to the safe arm, never to a wrong value. A mime type does not exist on a disk file, and a content sniff is one more guess. One concession: if the fall-through volume proves high in practice, a host library widens coverage at divergence risk. Measure first.

harness/src/report-model/production-resolver.ts
vivere-dally: No growth exists here. The cache lives in the resolver closure, and the tools construct a resolver for each call (`record-version.ts:131`, the same in the preview). The instance dies with the tool call, thus the cache lifetime is one document pass, on the managed deployment too. A TTL would never fire before the map dies. The header names it "a per-pass cache" (`production-resolver.ts:497`).

harness/src/tools/report-session/examine-page.ts
```diff
@@ -0,0 +1,204 @@
...
+            // stays valid, thus the tool gives it back either way.
+            const stamped = await deps.gateway.stampSeen(threadId);
+            if (stamped.outcome !== "stamped") {
+                const detail = stamped.outcome === "failed" ? stamped.detail : "the session state row is absent";
+                logger.warn("the seen hash did not stamp", { threadId, analysisId, detail });
+            }
+
+            return ok({
+                outcome: "examined",
+                screenshotBase64: captured.screenshotBase64,
```
vivere-dally: The examined outcome returns the screenshot as base64 inside the JSON tool result, but the loop serializes every tool output as JSON text with no image content path. Thus the model receives 50-200KB of base64 text that it cannot see, on every look that the prompt orders. The verify loop gains no visual signal, and repeated looks flood the context window. The result must reach the model as an image content block, or the tool must return a path plus the console faults only.

harness/src/tasks/extract-values.ts
```diff
@@ -0,0 +1,267 @@
...
+
+/**
+ * Build the sandbox exec for one extraction pass. The command runs the fixed script through `python3 -c`.
+ * The working directory is the analysis mount, thus a relative request path reads the correct file. The
+ * request list rides in one environment variable.
+ *
+ * The function is pure, thus a test asserts the command shape without a sandbox.
+ */
+export function buildExtractionExec(analysisId: string, requests: readonly ExtractionRequest[], execId: string): SubmitExecBody {
+    const scriptRequests = requests.map((request) => ({ path: request.path, format: detectExtractionFormat(request.path) }));
```
vivere-dally: buildExtractionExec drops the pinned hash from ExtractionRequest, and the script never receives or checks one. If the file changes between the streamed sha256 of the identity layer and the pandas read, the extraction returns drifted bytes, the gate matches asserts against them, and a version records values that do not come from the pinned hash. This is the fabrication class that the gate exists to block. The script must re-hash the bytes and refuse a mismatch.

harness/src/report-model/production-resolver.ts
```diff
@@ -0,0 +1,784 @@
...
+        const locator = reference.locator;
+        let selectedRow: Row;
+        if (locator.row !== undefined) {
+            if (locator.row < 0 || locator.row >= rows.length) {
+                return fail(reference, "locator-out-of-range", `row ${locator.row} is outside the ${rows.length} rows`);
+            }
+            selectedRow = rows[locator.row];
+        } else if (locator.rowFilter !== undefined) {
+            const filter = locator.rowFilter;
+            const matches = rows.filter((row) => row[filter.column] === filter.value);
```
vivere-dally: rowFilter matches with strict equality while every CSV and TSV cell is a string. Thus a numeric filter value never matches a text-backed artifact, and the reference fails locator-out-of-range for a row that exists. valuesMatch coerces numerically for the assert, and the row selector does not, thus the outcome flips by artifact format. The filter compare must use the same numeric coercion as the assert rules.

harness/src/report-model/production-resolver.ts
```diff
@@ -0,0 +1,784 @@
...
+ * A parquet integer column decodes as a `bigint`, thus the coercion narrows it to a number. A `null` or an
+ * absent value is not a cell, thus it drops and the row omits the key. The fixture row omits an empty cell
+ * the same way. An object or an array is structural doubt, thus it drops too and the format falls through
+ * where the caller detects the loss.
+ */
+function coerceCell(value: unknown): string | number | undefined {
+    if (value === null || value === undefined) {
+        return undefined;
+    }
+    if (typeof value === "number" || typeof value === "string") {
```
vivere-dally: coerceCell passes NaN through, because typeof NaN is number, and selectCell guards only null and undefined. A parquet NaN with no authored assert resolves as a valid scalar, the gate passes, and the page renders NaN. The extraction arm drops NaN to None, thus the two arms of one seam disagree on the same bytes. coerceCell must drop a non-finite number, the same as the script.

harness/src/tools/report-session/examine-page.ts
```diff
@@ -0,0 +1,204 @@
...
+                captured = await capture(pathToFileURL(pagePath).href);
+            } catch (cause) {
+                logger.warn("the page capture failed", { threadId, analysisId, ...defaultErrorFields(cause) });
+                return ok({ outcome: "capture-failed", detail: cause instanceof Error ? cause.message : String(cause) });
+            }
+
+            // The look counts only when the seen hash lands. A failed stamp is a transient store fault, and
+            // the record gate refuses a never-seen page, thus a later look re-stamps the marker. The picture
+            // stays valid, thus the tool gives it back either way.
+            const stamped = await deps.gateway.stampSeen(threadId);
```
vivere-dally, line 160: When the rendered stamp of the preview failed, stampSeen copies a NULL or stale rendered_document_hash, and the tool still reports examined. The record then refuses with never-seen or stale-look, and nothing tells the agent that the fix is a new preview. The agent can oscillate examine and record until the iteration cap. The eyes must refuse, or the outcome must name the absent rendered stamp.

harness/src/tools/report-session/preview-report.ts
```diff
@@ -268,11 +274,12 @@ export function createPreviewReportTool(deps: PreviewReportToolDeps): Tool<Previ
             }
             const document = finished.document;
 
-            if (deps.resolver === undefined) {
+            if (deps.makeResolver === undefined) {
                 return ok({ outcome: "resolver-unavailable" });
             }
+            const resolver = deps.makeResolver({ analysisId, auth: ctx.session.auth });
```
vivere-dally: deps.makeResolver runs unguarded here and in record-version.ts:131, and the assemble factory calls resolveWorkspaceRoot, which throws on an unresolvable root. The throw escapes execute, and the loop maps it to a generic error text with no typed arm. examine_page guards this exact call, and the two sibling tools must guard it the same way.

harness/src/tasks/extract-values-script.ts
```diff
@@ -0,0 +1,147 @@
...
+def read_frame(path, fmt):
+    # pyarrow reads the parquet columnar format.
+    if fmt == "parquet":
+        return pd.read_parquet(path, engine="pyarrow")
+    if fmt == "tsv":
+        return pd.read_csv(path, sep="\\t", dtype=str)
+    if fmt == "json":
+        return pd.read_json(path)
+    # The default reads a comma-delimited text file.
+    return pd.read_csv(path, dtype=str)
```
vivere-dally: pd.read_csv with the default keep_default_na turns an empty string and the literal NA, NaN, null, and N/A into NaN, and rows_of drops the key. The host parser keeps each of these as a value. Thus one reference into a column that holds the literal NA passes under the cap and fails over it, and the verdict depends on file size, not content. The read must set keep_default_na=False.

harness/src/report-model/production-resolver.ts
```diff
@@ -0,0 +1,784 @@
...
+    for (const name of header) {
+        if (seen.has(name)) {
+            return undefined;
+        }
+        seen.add(name);
+    }
+    const rows: Row[] = [];
+    for (let r = 1; r < records.length; r += 1) {
+        const fields = records[r];
+        if (fields.length !== header.length) {
```
vivere-dally, line 408: A blank line anywhere in a multi-column CSV becomes the ragged record [""], and the field-count check fails the whole file. Every preview and record on an otherwise-clean small CSV then pays an ephemeral container and a workflow to read one cell, and a sandbox fault turns the file into unreadable-artifact. The parser can skip a fully empty record without a dialect guess.

harness/src/tools/report-session/record-version.ts
```diff
@@ -0,0 +1,184 @@
...
+            if (!finished.valid) {
+                return ok({ outcome: "gaps", gaps: finished.gaps });
+            }
+
+            if (deps.makeResolver === undefined) {
+                return ok({ outcome: "resolver-unavailable" });
+            }
+            const resolver = deps.makeResolver({ analysisId, auth: ctx.session.auth });
+
+            const validation = await validateReport(finished.document, snapshot, resolver);
```
vivere-dally, line 159: The full validation, with the prepare pass that can start a container and a minutes-long exec, runs before the cheap look-before-record hash compare at lines 146-151. A premature record pays the whole sandbox cost and then refuses with never-seen or stale-look. The hash compare must run first, because it is one equality over state that the tool already loaded.

harness/src/tools/report-session/examine-page.ts
```diff
@@ -0,0 +1,204 @@
...
+            });
+
+            await page.goto(url, { waitUntil: "networkidle2", timeout: NAV_TIMEOUT_MS });
+
+            // The theme-ready event fires after the chart theme registers. The `__inflexaThemeReady`
+            // sentinel guards a page that already dispatched the event before this wait, thus a plain
+            // listener does not block forever. The callback runs in the browser context.
+            await page
+                .evaluate(
+                    new Function(
```
vivere-dally: The capture waits up to 8 seconds for inflexa-theme-ready, and the session page from report-render never dispatches that event. The event belongs to the old agent-authored path. Every look burns the full timeout and gains no signal. Dispatch the event from the render skeleton, or key the wait on what the page exposes.

harness/src/tools/report-session/examine-page.ts
vivere-dally: Fixed. The screenshot now rides as an image content block in the tool result. The Anthropic wire renders it, an OpenAI-compatible wire drops the block with a logged warn, and the JSON text carries no base64.

harness/src/tasks/extract-values.ts
vivere-dally: Fixed. The exec input carries the pinned hash for each request. The script streams a sha256 over the bytes before any read, and a drift refuses as a typed hash-mismatch error.

harness/src/report-model/production-resolver.ts
vivere-dally: Fixed. The filter compare now uses the shared numeric coercion of the assert rules, and the fixture and the production realization call the same function.

harness/src/report-model/production-resolver.ts
vivere-dally: Fixed. coerceCell drops a non-finite number, thus a NaN or an Infinity cell reads as an absent value, the same as the extraction script.

harness/src/tools/report-session/examine-page.ts
vivere-dally: Fixed. The seen-stamp store operation reports whether a rendered hash existed. When none exists, the eyes return the typed missed-stamp outcome that directs a new preview.

harness/src/tools/report-session/preview-report.ts
vivere-dally: Fixed. Both tools guard the resolver construction and return the typed root-unresolvable outcome, the same pattern as examine_page.

harness/src/tasks/extract-values-script.ts
vivere-dally: Fixed. Each pandas text read sets keep_default_na=False with no na_values, thus a literal NA and an empty string survive as text, in parity with the host parser.

harness/src/report-model/production-resolver.ts
vivere-dally: Fixed. The record walk skips a fully empty record before the field-count compare, thus a blank line parses clean and a genuinely ragged record still falls through.

harness/src/tools/report-session/record-version.ts
vivere-dally: Fixed. The never-seen and stale-look compares run first on loaded state. The resolver constructs and the validation runs only after the look holds.

harness/src/tools/report-session/examine-page.ts
vivere-dally: Fixed at the page. The chart bootstrap sets the sentinel and dispatches inflexa-theme-ready when it completes, and immediately when no chart exists. The capture now returns on a real signal.

harness/src/tools/report-session/examine-page.ts
vivere-dally: Follow-up in #348

vivere-dally:
@claude review

krajla (dismissed):

radutanasa (approved):

PR #349, opened 2026-08-11
feat: add a configurable request timeout for slow local models

vivere-dally:
## What & why

A slow local model (for example, turbo-fieldfare on a 16 GB M4 Mac mini) thinks longer than the transport limits of the runtimes. The Bun fetch of the CLI cuts each request after 300 seconds of silence, and no config option controlled it. The cut classified as a retryable connection error, thus the user saw silent retries and then a failure. This PR adds one config surface that lets such a model complete chat, the data profiler, and every other provider path.

Notes for the reviewer:

- The harness owns the bound. `AiSdkProviderConfig` accepts `requestTimeoutMs` and `maxRetries`. A fetch wrapper aborts an attempt when a silent interval — the wait for the response start, or a gap between body chunks — exceeds the value. The failure classifies as a retryable provider timeout.
- `ChatProvider` advertises the limit. The ad hoc router and the planner derive their deadlines from it, with their constants as floors.
- The CLI wires the two values from `models.connection`, in both modes. Its transport lift rides the untyped Bun init key `timeout: false`. An empirical probe on Bun 1.3.14 drove that choice. A numeric `timeout` and a long abort signal do not lift the 300-second idle cut. Only `timeout: false` does, and the cast site records this.
- An absent field changes nothing, byte for byte.
- The specs are synced, and both OpenSpec changes are archived in this branch.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist


cli/src/modules/harness/runtime.ts
```diff
@@ -569,6 +569,41 @@ export function buildAuthInjectingFetch(source: CredentialSource, underlying: In
...
+export function buildProviderFetch(
+    requestTimeoutMs: number | undefined,
+    source: CredentialSource | undefined,
+    underlying: InjectingFetch = fetch,
+): InjectingFetch | undefined {
+    const timeoutFetch = requestTimeoutMs !== undefined ? buildTimeoutLiftingFetch(underlying) : undefined;
+    if (source !== undefined) return buildAuthInjectingFetch(source, timeoutFetch ?? underlying);
+    return timeoutFetch;
+}
+
```
vivere-dally: I don't believe this is a good solution. Why are we injecting the timeouts into the plain fetch? These are not needed.
The AI-sdk will handle retrying and timeouts...

vivere-dally:
@claude review

krajla (approved):

PR #431, opened 2026-08-19
docs(harness): propose the citation-grounding ledger

radutanasa:
## What & why

This pull request adds the OpenSpec change `citation-grounding-ledger`, documents only. The run synthesizer writes model-generated PMIDs into durable findings. The only guard is a prompt rule plus a format validation (`harness/src/execution/run-synthesis.ts:184`). The change proposes a citation ledger: the literature-reviewer tool records each PMID that its inner tool results carried, and `submit_synthesis` rejects a PMID outside the ledger.

The gate is deterministic, and it makes no network call. The change also gives the voluntary `resolve_citation` tool to the analogical reasoner and to the target-assessment critique.

Notes for the reviewer:

- The ledger source is the inner tool results of the reviewer, not its prose report. Thus a PMID that the reviewer invents stays out of the ledger.
- The gate rejects a PMID that only a step summary carries, and this is deliberate. The synthesis prompt already forbids that path, and the gate now enforces it.
- The sandbox agent metas and the report session stay out of scope. The report session already binds each citation to pinned evidence.

## Type of change

- [ ] Bug fix
- [x] Documentation

## Checklist

### DIFF
--- a/harness/openspec/changes/citation-grounding-ledger/.openspec.yaml
+++ b/harness/openspec/changes/citation-grounding-ledger/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-08-19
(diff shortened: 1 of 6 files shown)


vivere-dally:
DFTM

PR #452, opened 2026-08-24
feat(data-profile): menu-driven profiling — detected sets, declared groups, dimensions

krajla:
Implements the OpenSpec change `data-profile-menu-and-dimensions` (archived in this PR together with its predecessor `data-profile-orientation`; the decision ledger is the archived design.md).

## What changes

- **Scanner is a five-stage pipeline** (`input-scan/`): quarantine → name markers → sibling-directory clustering → full-path template mining (opaque-ID + date token classes, suffix-chain slots, MDL-bounded) → assembly with variance carried, not split. Sets can span directories; companions attach to their data file; per-member completeness is computed.
- **The agent authors groups by operating on a bounded menu** — `use` / `split` / `merge` / `group`. No pattern typing, no declared counts (both are schema-unrepresentable). Membership, counts, and the partition are resolved deterministically at submit, with one full-resubmit repair round and a visible `unclassified` sweep; final-round contested files sweep with a `partition.contested` record.
- **Axes become dimensions with corroborating observations**: slot observations (template-keyed bindings, cardinality/values computed), column/document observations with citations, `checked:{matched,of}` only when measured, no canonical cardinality, optional nesting. A shipped vocabulary (roles, categories with defaults, probe list) feeds both prompt and validation from one source.
- **Prompt rewritten** around the substrate test, category defaults, and the five-probe list with explicit outcomes — an empty answer is a correct answer.
- **Incremental absorption**: the persisted recipe is template-keyed (including dimension bindings and the unclassified sweep), so a byte-identical tree re-absorbs fully with no sandbox and no LLM; only structurally new files wake the agent, over the delta.
- **Projection**: three index tiers (`input-group`, `input-dimension`, annotated members under legacy `input`) with replace-not-merge semantics; orientation renders the whole structure (census reserved, prose the only clamp target, mandatory elision markers); `inspect_data_profile` gains the `groups` scope with legacy rows served labelled.
- **Decoder split**: prefix-sufficient readouts run in harness TS over the workspace read seam; only Parquet/HDF5-family (plus PDF/DOCX) decode in the sandbox via a linted, pytest-covered Python asset. The inline `String.raw` blob is gone.
- **Schema renames** `kinds`→`groups`, `axes`→`dimensions` with legacy optional-on-read; no version field.

## Verification

- Full suite green (3655 pass / 0 fail), typecheck/lint/prettier clean, ruff + pytest wired into CI, `openspec validate --strict` green.
- Two independent adversarial review passes; all findings fixed except one declined with a measured rationale (JSON-schema `$defs` dedup — needs a global `defineTool` change and `$ref` support is not contractual across providers).
- Host-side end-to-end validation against a real large delivery tree ran in a private workspace (per the change's data-hygiene rule, only the conclusion is recorded): full partition with an empty `unclassified`, temp artifacts quarantined, cross-directory identity slots detected and cross-linked, bounded menu/submission, no repair round needed.
- Property tests assert the partition invariant over generated trees and operation lists.

## Follow-ups (recorded, not in this PR)

- Lumen: contracts bump for the new record shape.
- Cortex: harness version bump; ticket for the input materializer's partial-download temp files.
- Marker-unit co-clustering hardening; the predecessor change's three live-model validations (see its archived tasks.md).


krajla:
@claude review

radutanasa (dismissed):

Issue #69, opened 2026-07-10
K8s workspace mounts + preview paths after the workspace-root seam (PR #65)


vivere-dally:
Two questions from reviewing #65, both about surfaces this repo cannot see. cc @krajla

### 1. Did anything downstream consume `workspace/mount-strategy.ts`?

It had no importers here — `buildMountPaths`, `buildPodMounts`, and `buildDockerMounts` were all unreferenced, and `sandbox/mount-plan.ts` is the live mount model for both backends. I deleted the module and removed the `workspace-profiles` requirement that specified it.

If a managed deployment imported it (`@inflexa-ai/harness/workspace/mount-strategy.js`), that import breaks. `buildMountPlan(coords, stores)` returns the container-side paths both backends use; `buildSessionSubPaths(coords, workspaceSubPath)` returns the K8s `subPath`s.

### 2. K8s `subPath` no longer assumes `{pvcRoot}/{analysisId}`

`mount-plan.ts` hardcoded `sessionSubPathRO = analysisId`, while `createSandbox` pre-creates the writable step tree at `resolveWorkspaceRoot(analysisId)/runs/{runId}/{stepId}`. Those two agree only if the resolver returns `{pvcRoot}/{analysisId}` — structurally guaranteed before #65 (both derived from `sessionsBasePath`), an unenforced convention after it, with nothing in the type system or the tests to catch a divergence.

It is now `relative(sessionPvcRoot, resolveWorkspaceRoot(analysisId))`, so **`K8sClientConfig` requires a new `sessionPvcRoot` whenever `sessionPvc` is set** — the absolute mountpoint of that PVC on the Cortex pod's own filesystem. Without it, `createSandbox` throws on the first step rather than silently mounting a same-named sibling directory. A root resolving outside `sessionPvcRoot` also throws.

**Q:** is there a managed deployment wiring `SANDBOX_BACKEND=k8s` that needs this config added?

### 3. Previews moved on disk

They now live at `{resolveWorkspaceRoot(analysisId)}/previews/{previewId}/…`, not `{pvc}/previews/{analysisId}/{previewId}/…`.

[...]

PR #453, opened 2026-08-25
feat: the local package store — the pool, the per-analysis farms, and the catalog build

vivere-dally:
## What & why

The runtime image baked every R and Python library. Thus the image was many gigabytes, and an analysis could not gain a package after the build. This PR moves them into a local package store: one content-addressed pool on the host, one symlink farm for each analysis, and a dependency graph. The sandbox mounts the store read-only at `/mnt/libs`.

What this adds:

- **Harness**: the required `farmSource` seam resolves one farm per analysis at each sandbox creation. The `extendAnalysisFarm` seam gives each sandbox agent `link_packages`. The farm `inflexa.lock` is the mount gate and the inventory. A pool miss returns classified: in flight, failed with its recorded reason, or absent.
- **CLI**: the store commands, the detached transfers, the sandbox gate, and the acquisition flights with the pending set. The catalog transfer runs under an unpacking watch with a phase readout, thus a frozen child settles as failed instead of a forever-running row. A request keeps its raw spelling beside its canonical identity, thus `GO.db` reaches pak verbatim. `store download --foreground` serves a one-shot container, for the managed refresher (platform-charts #69).
- **Images**: `sandbox-base` bakes no library. The provisioner builds the catalog and acquires under the pins of the pool. The store publishes to GHCR as an OCI artifact, with a non-gating acceptance run.
- **Migration**: creation makes the empty farm eagerly, and the heal composes the full farm from the catalog closure.

Notes for the reviewer:

- The OpenSpec trees of `cli` and `harness` carry the requirements: the `package-store-*` specs and the archived changes.
- `docs/local-package-store.md` is the consolidated working record: the vision, the 38 decisions, the spike harvest, the research verdicts, and the CI postmortems.
- `.github/workflows/package-store-pipeline.yml` is the one-dispatch driver of the three store builds, and it stays after the merge.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved, and this PR makes no loosening. The store mounts read-only, and the egress rules do not change.


krajla (approved):
LGTM 👍🏻

PR #464, opened 2026-08-27
feat(harness): ground plans in a cited knowledge corpus (knowledge plane, phase 1)

radutanasa:
## What & why

Method selection had no ground: the planner never saw the skill packs, a plan step carried no citation, and validation covered structure only. This change lands Phase 1 of the knowledge plane, per the OpenSpec change `harness/openspec/changes/knowledge-plane-phase-1`. It adds a `KnowledgeBase` seam with a file-backed realization, and a cited rule corpus at `knowledge/` (15 records, each with a DOI or a PMID). It adds a knowledge brief in the planner seed, and knowledge tools for the planner and the sandbox agents. A `grounding` field on the plan step and a grounded gate in `submit_plan` complete the contract. The gate gives each plan a citable decision chain, and it catches an unacknowledged applicable `reject` rule before a run starts.

Notes for the reviewer, before the diff:

- The gate enforces acknowledgment, not method compliance, because a Phase-1 step carries no typed method. Only `reject` severity blocks, and only when the rule mechanically applies. `warn` and `note` advise.
- Nothing binds a corpus in this PR. The CLI wiring is a later change in `cli/`, thus the shipped CLI behavior does not change here. An install with no corpus follows the exact pre-change paths.
- The corpus conditions are a closed, evaluable set. A condition over a fact the profile lacks gives `not_evaluable`, and it never blocks — the gate stays honest about what it can test.
- Two schema-drift guards fired as designed, and the resolutions are deliberate: `grounding` renders in the step briefing, and the sandbox always-on tool count grew by the two knowledge tools.
- Task 5.4 (one end-to-end grounded rejection loop through `harness:verify`) stays open in `tasks.md` — it needs a live model provider.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] **Methods:** the method is stated, relevant literature cited where appropriate, and tool/library versions are pinned.
- [x] **Sandbox:** the security model is preserved (the knowledge tools run host-side, and no container surface changes).
- [x] **Provenance:** prior analyses remain reproducible (the `grounding` field is optional on the persistence schema, and historical plans parse unchanged).


radutanasa:
Closed. Phase 0 replaced this Phase 1 design. The knowledge content is now in a private repository, and the harness seam comes in a new pull request.

PR #467, opened 2026-08-27
feat: report lineage — from a claim in the page to the raw data

### LINKED ISSUE #318: Report lineage: derive a provenance bundle for a report version

vivere-dally:
## What & why

Each grounded block of a report page now opens a lineage popover that walks from the pinned artifact back to the raw data. The provenance stays where it lives. The harness emits report observation events, and it stages the signed analysis document into the page assets. The cli records the events into that same document. The tsprov browser bundle walks it in the page, offline. This closes the last open piece of the report rebuild: a reader can ask "where did this number come from" and get the chain, content-addressed at every hop.

Reviewer notes, before the diff:

- The harness stays free of every provenance API. A lint rule bans tsprov imports, and the bundle rides the page as static bytes only.
- The recorder maps the report event family in the host, before the kernel dispatch. Thus the kernel exhaustiveness guard keeps its force for the core family.
- The two seams bind at the composition root, and each is optional alone. Absence is a normal condition, and the page degrades to the honest last hop.
- The popover rail renders from the walk edges, never from the flat node set. Off-chain files collapse behind one count row, and the design fixture plus a headless-Chrome drive pin the anatomy.
- The harness spec changes are synced and archived. The cli change `record-report-provenance` is implemented, and it stays active in its tree.

Closes #318

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] **Sandbox:** the security model is preserved, and any loosening is called out and justified. This PR touches no sandbox code.


vivere-dally:
Note to self, the implementation is not ok - needs to be changed. But we're in a PoC phase now. The report-prov files can go, it should use the single & existing provenance recorder

vivere-dally:
@claude review

cli/src/modules/prov/prov.ts
```diff
@@ -111,19 +111,244 @@ function isProvEvent(event: StampedEvent): event is StampedProvEvent {
...
+/** What a mapped act mints, so its arm can hang the edges only that act needs. */
+type ReportAction = { actionQn: string; agentQn: string; time: string };
+
+/**
+ * Narrow a prov member to the report family. The two type strings ARE the discriminant, so the test
+ * corresponds exactly to the template type above — that equivalence is what makes the predicate
+ * sound, the same argument {@link isProvEvent} rests on.
+ */
+function isReportEvent(event: StampedProvEvent): event is StampedReportEvent {
+    return event.type === "prov.session_created" || event.type.startsWith("prov.report_");
```
vivere-dally: Why do we consider a `session_created` to be a report event? It shouldn't. We now record all session creation. It is a plain prov event.
But why do we even need to differentiate if a prov event is for a report or not? I don't think this makes sense

cli/src/modules/prov/prov.ts
```diff
@@ -111,19 +111,244 @@ function isProvEvent(event: StampedEvent): event is StampedProvEvent {
...
+/**
+ * Map one report member onto the analysis's live document.
+ *
+ * Every member lands one typed action activity, so each record says what was done and when, in the
+ * shape the lifecycle family already uses. Two members mint an entity beside it: a created report
+ * session mints the report itself, and a recorded version mints that version.
+ *
+ * A throw here reaches the recorder's shared guard, exactly as a kernel builder throw does.
+ */
+function appendReportRecords(doc: ProvDocument, event: StampedReportEvent): void {
```
vivere-dally: why isn't this part of the main switch statement in prov kernel?

cli/src/types/prov.ts
```diff
@@ -29,3 +29,88 @@ export type {
...
+    threadId: string;
+    /** `report` sessions get an `inflexa:Report` entity; a `conversation` is the session alone. */
+    kind: "conversation" | "report";
+    /** The thread the session was started from, absent on a root session. */
+    parentThreadId?: string;
+};
+
+/** One block operation on a report document — the payload of the four `prov.report_block_*` members. */
+export type ProvReportBlockRef = {
+    /** The report session the block belongs to. */
```
vivere-dally: don't we need a block kind too here?

cli/src/lib/bus.ts
```diff
@@ -68,6 +68,57 @@ function eventFields(event: StampedEvent): Record<string, unknown> {
...
+        case "prov.report_block_added":
+        case "prov.report_block_changed":
+        case "prov.report_block_removed":
+        case "prov.report_block_moved":
+            return {
+                analysisId: event.analysisId,
+                actorKind: event.actor.kind,
+                threadId: event.block.threadId,
+                blockId: event.block.blockId,
+                model: event.model,
```
vivere-dally, line 93: don't we need to know the kind too? If the block was added or changed or removed?

cli/src/modules/harness/report_bridge.ts
vivere-dally, line 1: the comments are absurdly verbose in this file. Again, why do we need to differentiate between report prov events and plain prov events?

harness/src/agents/conversation-agent.ts
```diff
@@ -162,6 +163,12 @@ export interface ConversationAgentDeps extends EnvironmentStorePaths {
      * at a page reads that same answer.
      */
     readonly eyes?: AcquireEyes;
+    /**
+     * The observation seam of a report session. `start_report_session` hands it to
+     * its spawn, which emits the creation of the session at the moment that the
+     * child lands. No other tool of this roster emits on it.
+     */
+    readonly emitReportObservation?: EmitReportObservation;
```
vivere-dally: we should reuse the existing prov seam...
again, create session is just a plain prov event. Similar to analysis creation

harness/src/tools/report-observation.ts
vivere-dally, line 1: I don't agree of this file. It should be part of the existing prov seam...

harness/src/tools/report-provenance.ts
vivere-dally, line 1: I don't agree of this file. It should be part of the existing prov seam...

cli/src/modules/prov/prov.ts
vivere-dally: The distinction is gone. Every prov member now walks one path: the bus, then toKernelEvent, then applyProvEvent. Kernel 0.6.0 carries the report family in its core union.

cli/src/modules/prov/prov.ts
vivere-dally: It is part of the kernel switch now. Kernel 0.6.0 maps the nine members in applyProvEvent, with first-declaration guards, and the host switch is deleted.

cli/src/types/prov.ts
vivere-dally: Added. The four block members carry blockKind end to end: the seam, the bus, and the record attribute inflexa:blockKind.

cli/src/lib/bus.ts
vivere-dally: The act rides event.type (prov.report_block_added and its siblings), and the bus tap logs that field. The block kind now rides the projection too.

cli/src/modules/harness/report_bridge.ts
vivere-dally: The file is gone. The mapper folded into prov_bridge.ts with trimmed comments, and no report-versus-plain split remains.

harness/src/agents/conversation-agent.ts
vivere-dally: Done. One ProvenanceSeam carries the run emit, the session emit, and the document read. The boot installs one object, and the turn reads it.

harness/src/tools/report-observation.ts
vivere-dally: The file is deleted. Its vocabulary lives in harness/src/provenance/seam.ts, as the session half of the one ProvenanceSeam.

harness/src/tools/report-provenance.ts
vivere-dally: The file is deleted. The document read is the readExport member of the one ProvenanceSeam, realized in prov_bridge.ts.

krajla (approved):

PR #468, opened 2026-08-28
fix(harness): bind the sandbox thread pools to the cpu quota

radutanasa:
## What & why

Each sandbox backend turns the step resources into a hard limit: `NanoCpus` and
`Memory` on Docker, `requests` and `limits` on K8s. But a cgroup quota is
invisible to the runtimes inside it. `parallel::detectCores()` and the OpenBLAS
pool read the core count of the host from `/proc`, not the quota.

Thus a step with 2 CPUs on a host of 12 cores forks 12 workers into a cgroup of
2 CPUs and 8 GB. Then the memory runs out. sandbox-server shares that cgroup,
thus it does not answer `/exec` again. The container now receives the quota that
the cgroup already enforces: `OMP_NUM_THREADS`, `OPENBLAS_NUM_THREADS`, and
`MC_CORES`, from the clamped `spec.cpu`.

Notes for the reviewer:

- Each backend composes its env from one record, not from a list. A duplicate
  key resolves at the kubelet or at the libc that reads `environ`, not in this
  code. One record makes the precedence explicit, thus the `extraEnv` of an
  embedder keeps the last word over the derived count.
- The three variables cover the R and Python stack of the image. OpenBLAS keeps
  a private pool that ignores `OMP_NUM_THREADS`, and R reads `MC_CORES` at
  startup into `options(mc.cores)`, which is what `mclapply` and BiocParallel
  size from.
- No spec states this behavior yet. `openspec/specs/docker-sandbox-provider/spec.md`
  states the `NanoCpus` and `Memory` translation only.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved. The change adds three env
  values. It grants no capability, no mount, and no egress, and it does not
  change a resource limit.
- [x] **Provenance:** a prior analysis stays reproducible. But a parallel step
  now runs with fewer workers, thus a method whose result depends on the worker
  count can vary.

### DIFF
--- a/harness/src/sandbox/docker-client.test.ts
+++ b/harness/src/sandbox/docker-client.test.ts
@@ -291,6 +291,47 @@ describe("docker createSandbox — transport modes", () => {
         expect(registered).toEqual([{ runId: "run-1", stepId: "step-a", sandboxId: ref.sandboxId }]);
     });
 
+    test("the cpu request is published as the thread count of every parallel runtime", async () => {
+        const { docker, created } = stubDocker();
+        const ops = createDockerSandboxOps({
+            image: "sandbox-base:latest",
+            cortexBaseUrl: "https://x",
+            resolveWorkspaceRoot: (id) => join("/sessions", id),
+            docker,
+            fetch: okFetch,
+            registerSandbox: async () => {},
+        });
+
+        (await ops.createSandbox(META, mintSandboxIdentity("run-1")))._unsafeUnwrap();
+
+        // Without these the runtimes size their pools from the host's core count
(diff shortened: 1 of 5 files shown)


radutanasa:
Closed. A probe of the sandbox image shows that no environment value reaches the failure path.

- `MC_CORES` raises the default worker count of `mclapply` from 2 to the cpu request.
- `parallel::detectCores()` reads `/sys/devices/system/cpu/online`, thus it reports the cores of the host under any cpu limit.

A new design puts the shim in the image, where the container reads its own cgroup quota.

PR #469, opened 2026-08-28
fix(sandbox): publish the cpu quota of the container to the runtimes in it

radutanasa:
## What & why

The harness gives each container a hard CPU quota. A cgroup quota is invisible
to the runtimes inside it, because `/proc` and `/sys` still describe the
machine. Thus `nproc`, `os.cpu_count()`, and `parallel::detectCores()` all
report the cores of the host under any quota.

Code that sizes a worker pool from that number forks one worker for each core of
the host. Each fork copies its working set. Thus the forks exhaust the memory
limit of the container before the CPU limit throttles them, and the workload
takes sandbox-server down with it.

sandbox-server now reads the quota of the cgroup that it is in, and it publishes
that number to each command. The value comes from the kernel, not from the
request that made the container, thus the two can never disagree.

Notes for the reviewer:

- An environment value alone reaches neither language. `detectCores()` reads
  `/sys/devices/system/cpu/online`, and `os.cpu_count()` reads the machine. Thus
  the image carries one shim for each: an R profile on `R_PROFILE_USER`, and a
  Python module that a `.pth` line imports.
- `MC_CORES` and `options(mc.cores)` stay unset. R defaults `mclapply` to 2
  workers, and that default is R's own, not a count of the host. A value there
  would raise the fork count.
- Each published name is a default. A value on the container, and a value in the
  exec request, both keep the last word over it.
- Under a quota of 2 CPUs a probe of the image gives 2 for `detectCores()`, for
  `mclapply` with `mc.cores = detectCores()`, for `os.cpu_count()`, and for
  `mp.Pool()`. With no quota the same probe gives the numbers of the host.
- `go vet` and `go test ./...` pass in `images/sandbox-base/server`. The two
  shim files have no CI coverage, and the probe above is the only record of
  them.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally. This change touches no TypeScript, so the Go suite of the image is what ran.

- [x] **Sandbox:** the security model is preserved. The change adds two read-only
  shim files and some environment names. It grants no capability, no mount, and
  no egress, and it changes no resource limit.
- [x] **Provenance:** the provenance hooks are untouched. The R shim rides on
  `R_PROFILE_USER`, and the provenance profile keeps `R_PROFILE`. A probe of the
  image shows that both profiles load. But a parallel step now runs with fewer
  workers, thus a method whose result depends on the worker count can vary.

### DIFF
--- a/images/sandbox-base/Dockerfile
+++ b/images/sandbox-base/Dockerfile
@@ -277,6 +277,10 @@ COPY --from=provtrack-builder /provtrack.so /opt/provenance/provtrack.so
 COPY images/sandbox-base/provenance/sitecustomize.py /opt/provenance/sitecustomize.py
 COPY images/sandbox-base/provenance/Rprofile.site /opt/provenance/Rprofile.site
 
+# CPU-quota shim for R. sandbox-server points R_PROFILE_USER at this file and
+# publishes the quota as INFLEXA_CPU_LIMIT; see server/cpulimit.go.
+COPY images/sandbox-base/cpu-limit.R /opt/cpu-limit.R
+
 # Default to the unprivileged workload user. Callback mode and K8s run as uid
 # 1000 throughout. Docker poll mode overrides `User` to root at create time so
 # the entrypoint can install the egress firewall, then `setpriv` drops back to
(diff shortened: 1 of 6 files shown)


krajla:
Can't the harness add the envs at runtime instead of the sandbox from inside determining the sizes?

krajla:
https://github.com/inflexa-ai/inflexa/pull/472

PR #471, opened 2026-08-28
feat(harness): complete the data profile of an analysis with no input files

krajla:
## What & why

An analysis with no input files had no profile state that says "done". `triggerDataProfile` refused an unseeded analysis and an empty manifest, so the ledger row stayed at `pending` or NULL, and a consumer read it as "not profiled". 

A profile describes files, and an analysis with no files has nothing to describe. Thus the profile is now complete at once: the new ledger operation `completeEmptyDataProfile` stamps the row `completed` with no result, no workflow id, and the seed `[]`. The trigger takes this route when the manifest is empty and the seed names no file, and it returns the new result `"completed"`. No claim runs, and no workflow starts.

Notes for the reviewer:

- The CAS of `completeEmptyDataProfile` carries the negation of the `SEEDED` claim conjunct. Thus a claim into `running` and this stamp can never both win on one row. A seed upsert that lands between the pre-read and the stamp is never hidden behind a finished profile.
- A `completed` row with the seed `[]` is not claimable for a rerun. A later seed upsert that names files makes it claimable, the same as any completed row.
- An empty manifest against a seed that names files stays a refusal. The unseeded refusal now applies only to a manifest that names files.
- `DataProfileTriggerResult` gains the member `"completed"`. The three exhaustive switches in the CLI carry the case. The parity ladder and the force path settle an empty set before they trigger, so they map it to `already_profiled`. The dev `inflexa profile` command reports the completion and exits. The CLI's empty-set branch still calls `clearDataProfile`, so the CLI keeps its "not profiled" route in this PR.
- The spec delta lives in `openspec/changes/complete-empty-data-profile`, and `openspec validate` passes.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

https://claude.ai/code/session_01QgKBpFt9m9dZ6ebY9hAY38

### DIFF
--- a/cli/src/modules/harness/dev/profile.ts
+++ b/cli/src/modules/harness/dev/profile.ts
@@ -108,6 +108,12 @@ export async function runProfile(flags: ContextFlags): Promise<void> {
         case "already_running":
             log.info("A profile run is already in progress — watching it");
             break;
+        case "completed":
+            // The analysis has no input files, so the ledger row is complete at
+            // once with no result. No workflow ran, so there is nothing to watch.
+            log.step("No input files — the profile is complete with no result");
+            outro("Done — inspect details with `inflexa profile --status`");
+            return shutdown(0);
         case "failed": {
             // The trigger claims pending/completed rows only; a failed row needs
             // the retry claim. Mirror the managed retry route: claim, then start.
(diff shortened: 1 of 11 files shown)


krajla:
@claude review

vivere-dally (approved):

PR #472, opened 2026-08-28
fix(harness): give each sandbox runtime the cpu quota of its step

krajla:
## What & why

Each backend turns the step resources into a hard cgroup limit, but the limit is invisible to the runtimes inside it. `parallel::detectCores()` greps `/proc/cpuinfo`, and `os.cpu_count()` reads `/sys/devices/system/cpu/online`. Both describe the host. Thus a step with 2 CPUs on a host of 12 cores forks 12 workers into a cgroup of 8 GB, and the forks exhaust the memory. On Docker, `Memory` alone lets the container swap as much again, thus the storm thrashes for minutes and sandbox-server stops to answer `/exec`.

The harness now does three things at launch, on both backends. It publishes the quota as the thread count of each library that reads one from the environment (`thread-env.ts`). It covers the two kernel files with the quota (`cpu-files.ts`: a bind mount on Docker, a ConfigMap with `subPath` mounts on K8s). On Docker it sets `MemorySwap` equal to `Memory`.

Replaces #469.

Notes for the reviewer:

- On Docker, through `submitExec`: `detectCores()` 2, `os.cpu_count()` 2, no swap inside the cgroup, and a fork storm of 6 workers ends with an OOM kill in 15 s instead of a hang. On staging EKS under gVisor, the sandbox sees no cgroup at all by default. The env covers the libraries that read it. The `subPath` mount over the two files gives the quota to each runtime and each library default (pydeseq2, xgboost, WGCNA, `mp.Pool()`).
- `MC_CORES` stays unset. R defaults `mclapply` to 2 workers, and a value there raises the fork count.
- The K8s service account of the harness must have `create`, `get`, `update`, and `delete` on `configmaps` in the sandbox namespace. That is a platform-charts change, and it must land before the pin bump. A missing permission fails the launch at once, with the cause.
- The `cpuinfo` blocks describe the harness host, not the node.
- No spec states this behavior yet. `openspec/specs/docker-sandbox-provider/spec.md` states the `NanoCpus` and `Memory` translation only.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved. The change adds env values, two read-only files over two kernel paths, and a swap limit. It grants no capability, no mount of host data, and no egress, and it lowers no resource limit.
- [x] **Provenance:** an earlier analysis stays reproducible. But a parallel step now runs with fewer workers, thus a method whose result depends on the worker count can vary.

https://claude.ai/code/session_01Ji4x1TwcZRuKJvtvpFgWbE

### DIFF
--- a/harness/package.json
+++ b/harness/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@inflexa-ai/harness",
-    "version": "0.30.2",
+    "version": "0.30.3",
     "description": "Host-agnostic bioinformatics agent harness — hand-rolled agent loop, DBOS-durable workflows, sandboxed compute.",
     "license": "Apache-2.0",
     "type": "module",
--- a/harness/src/sandbox/cpu-files.test.ts
+++ b/harness/src/sandbox/cpu-files.test.ts
@@ -1,0 +1,61 @@
+/**
+ * `cpuFiles` is pure: the content of the two cpu files from a thread count and
+ * the text of a real `/proc/cpuinfo`.
+ */
+
+import { describe, expect, test } from "bun:test";
+
+import { cpuFiles } from "./cpu-files.js";
+
+function block(n: number): string {
+    return `processor\t: ${n}\nvendor_id\t: GenuineIntel\nmodel name\t: Test CPU\ncpu MHz\t\t: 2400.000\nflags\t\t: fpu vme\n`;
+}
+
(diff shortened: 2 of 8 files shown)


krajla:
https://github.com/inflexa-ai/platform-charts/pull/68

vivere-dally (approved):

PR #491, opened 2026-09-01
feat(harness): move the file-tool writes to the host filesystem

krajla:
## What & why

The sandbox gave containment to a `write_file` or `edit_file` write, but it gave no provenance value. The mutator records the `file_tool` event in the host process, and the hashes come from the host. Thus the sandbox exec of a python program added cost and a python dependency for no benefit. Now the `WorkspaceMutator` seam writes with the host filesystem after the `resolveForWrite` confinement. `execute_command` stays sandbox-gated, and the container mounts are unchanged.

Notes for the reviewer:

- The confinement of a file-tool write moves from the kernel mounts to the host path checks. The seam adds the symlink checks: a realpath comparison of the deepest ancestor that exists, a refusal of a symlinked final component, and an `O_NOFOLLOW` open as the backstop. `SECURITY.md` now scopes the containment claim to command execution.
- The loop does not wrap a workflow-mode tool body in a step. Thus the disk mutation runs in `ctx.runStep`, and a DBOS replay returns the cached outcome without a second write.
- The provenance record lands outside that step on purpose. The collector is process-local, and a recovered replay must record again.
- `edit_file` gets a regex mode for a bulk edit: `regex=true` with `expected_matches` or `replace_all=true`. On a count mismatch the tool writes nothing, and it reports the count and the line numbers.
- The conversation agent now holds `write_file` and `edit_file` through a session-scoped mutator. The write prefix is the analysis root, and the same gauntlet refuses a traversal, a foreign analysis, and a symlink escape. The chat route gives the passthrough `runStep`, so no workflow is involved.
- The prov-kernel flattens `file_written` into one event for each write, and the `generation` field selects the authority: `"command"`, `"step"`, or `"call"`. A file-tool write, in a step or in chat, is a `"call"` write. It mints a deterministic call action from the digest of the invocation id and its scope, with no timestamp. The `file_tool` arm of `ProvCommandRef` is gone — a tool call is not a command. The stored documents do not change, and the readers keep the old shape. The unchanged `inflexa:FileToolWrite` type keeps one render path in the frontend.
- Draft PRs in the consumer repos wait on the npm publish: inflexa-ai/cortex#361 (bridge + the chat-write seam consumer), and a lumen PR for the golden-fixture sync.
- The sandbox agents and the report agent are unchanged. The report agent still cannot write.
- The artifact-hash reconcile now does the same realpath check before it hashes an entry. An entry that escapes the tree is not attested, and an input that escapes the tree is dropped from the lineage. The read seam held this check already.
- The OpenSpec change is `host-side-file-tool-writes`, and this PR syncs the main specs.
- `bun run lint` reports one error in `src/providers/ai-sdk.test.ts` that exists on main. This change does not touch that file.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved for command execution (unprivileged uid-1000 workload, all capabilities dropped, `no-new-privileges`, no network egress in poll mode, signature-authenticated exec endpoints, read-only analysis tree, resource limits, no host credentials in spawned commands). The one loosening — the file-tool write leaves the container mounts — is called out and justified above.
- [x] **Provenance:** prior analyses remain reproducible (or the expected variance is documented).


cli/src/types/events.ts
```diff
@@ -135,6 +136,8 @@ export type BusEvent =
      * reaches this contract only through the pin, never silently.
      */
     | { type: "prov.session_created"; analysisId: AnalysisId; actor: ProvActor; model: ProvModelId; session: ProvSessionRef }
+    /** A file a conversation turn wrote with a file tool — a session-scoped write with no run or step to anchor to. */
+    | { type: "prov.session_file_written"; analysisId: AnalysisId; actor: ProvActor; model: ProvModelId; write: ProvSessionFileWriteRef }
```
vivere-dally: I don't think we should differentiate between a conversation file written and a file written via a Run. The actor and model should be enough

vivere-dally (changes requested):
Please ensure the agent cannot write outside its analysis/run tree

cli/src/types/events.ts
krajla: The idea was extend file_written union with more options or create a new event type. The differences were file_written by step agents have step_id and are resilient to replays of the analysis. file_written by the agent directly has a timestamp(since doesnt need to be replay resilient), and thread_id and no step.
I also feel like they should be the same event though, the file_written as it is now might be too specific.

krajla:
> Please ensure the agent cannot write outside its analysis/run tree

We took the necessary precautions, same for symlink hopping being denied.

krajla:
@claude review

krajla:
The review found a correct gap, and 70488410 holds the fix. The landing now does a realpath check of the created directory after the recursive mkdir. If the canonical path escapes the prefix, the seam refuses the write with `out_of_prefix`. The open also uses the canonical path, thus the write does not go through the raced segment again.

A new test does the write 200 times, and it puts a symlink into the missing segment at a different point in each attempt. The test finds the escape without the fix, and it passes with the fix.

A residual window remains. A process can replace a created directory with a symlink between the check and the open. To close that window in full, `openat` with `RESOLVE_BENEATH` is necessary, and Node does not supply it. The pre-planted attack at a known missing path is gone, and only a rename race against a directory that exists remains.

krajla:
@vivere-dally Removed the newer seperate session_file_written. Modified file_written to not be so closely coupled with step agents and it can now represent files written with tools outside analysis workflow runs.

vivere-dally (approved):

vivere-dally:
LGTM 📿

PR #497, opened 2026-09-02
fix(harness): forward the libs root through the sandbox client factory

vivere-dally:
The K8s backend proves the `inflexa.lock` of a farm through the host mountpoint of the libs volume. The factory did not forward that root, and an embedder composes only through the factory. Thus the gate was unreachable for every embedder. Under a fixed farm source no resolver exists, and the forwarded root carries the only gate that shape has.

A spy test pins the forward, because the field is optional and tsc cannot miss it for a caller.


vivere-dally:
@claude review

radutanasa (dismissed):

krajla (approved):

PR #501, opened 2026-09-02
feat: the managed exposure of the package store

### LINKED ISSUE #500: Package store, managed exposure: the image fragment as a label, a required-store gate, the toolchain from the client, and mountpoints at download
## Problem / motivation

`list_available_packages` merges the farm's `inflexa.lock` with the image inventory fragment, `/opt/inflexa/image-packages.txt`, the list of tools the image bakes (`images/sandbox-base/Dockerfile:441-443` assembles it from the conda and node package lists). The harness reads it as a host path, `EnvironmentStorePaths.imagePackagesFile`, with the default `/opt/inflexa/image-packages.txt` (`harness/src/tools/sandbox/list-available-packages.ts:46`).

That path exists only inside the sandbox image. The CLI covers it by running the image with the entrypoint set to `cat` and caching the output by image digest (`cli/src/modules/libs/packages.ts:43-77`). A managed embedder cannot: the Cortex pod has no engine to run the image, so every production inventory lacks the "System tools (CLI)" section, and the agents never learn that samtools, bcftools, bedtools, cellsnp-lite, mageck, tabix, vcftools, and echarts exist.

The workaround on the Cortex side, a `COPY --from=<sandbox image>` into the Cortex image at build time, bakes a second pin of the sandbox tag into a different build, and the chart can override the runtime tag. The inventory would then describe an image that is not running.

The fragment is four lines, about 110 bytes. The image already carries OCI labels (`Dockerfile:186-187`).

## Proposed solution

Two small changes, none in an embedder:

1. **The image publishes the fragment as an OCI label.** At the step that assembles `image-packages.txt`, also set a label, for example `ai.inflexa.image-packages`, holding the same content. The label is image metadata about the image, which is what labels are for, and it travels with the tag through every registry.
2. **The harness reads the label for the configured image.** A `readImagePackages(imageRef)` on the sandbox client, or beside it: on the Docker backend through engine inspect, on the K8s backend through an anonymous registry fetch of the manifest and config blob (GHCR allows it, and the pod already has registry egress for the pull). Cache per digest. Bind the result where `imagePackagesFile` is read today, so `list_available_packages` merges it with no host path involved. `imagePackagesFile` can stay as the explicit override.

The CLI then drops its `cat` extraction, and the fragment always describes the running image, on every host.

## Alternatives considered

- **Read the file from the first sandbox of that image over the exec channel and cache per image ref.** No image change, but the conversation agent has no fragment until a sandbox has run once, and the cache needs a durable home.
- **Ship the fragment inside the store artifact.** Rejected: the fragment is the image's contract, and a copy in the store duplicates it and drifts.
- **Embedder-side build-time copy.** The second pin above.

## Area

Docker sandbox, with the harness's inventory tool as the consumer.

---

## Three more asks from the managed adoption (inflexa-ai/cortex#362)

Same context, same reviewer, so they ride this issue. Each stands alone.

### 1. Let the embedder declare the store mandatory, so the mount gate refuses instead of degrading

**Today.** At sandbox create the backend reads the farm's `inflexa.lock`. On failure it degrades: `docker-client.js:255-262` logs "mounting no package store (sandbox degrades to available:false)" and still makes the container; the K8s backend does the same through `libStorePvcRoot`. On the managed service that is a run that spends tokens on steps that can import nothing, and the only signal is a warning line.

**Ask.** A config fact on `CreateSandboxClientConfig`, for example `packageStore: "required"`, under which the same gate refuses at create with a typed `SandboxError` carrying the lock path and the parse reason. `farm_unavailable` fits ("the farm source refused the analysis"); a distinct `store_unusable` reads clearer. A create refusal already reaches the embedder's failure path: the execute-plan step fails with the reason, and the data-profile ledger row is stamped failed, which Cortex saw first-hand with `container_create_failed`. Default stays degrade, so the CLI keeps its behaviour.

### 2. Read the toolchain from the sandbox client, so one declaration feeds both consumers

**Today.** `toolchainSource` is declared on the sandbox client, which uses it for the resolver env and exposes nothing back (`SandboxClient` has only `createSandbox`, `submitExec`, `awaitExec`, `isAlive`). The orient prompt reads a second copy from each agent bag, `deps.toolchainSource` in `createSandboxAgent` (`agents/sandbox/shared.js:195`). Three builders fill that bag: Cortex's step-agent bag sets it; the CLI's `buildStepAgent` in `cli/src/modules/harness/run_deps.ts` does not (the only `toolchainSource` in `cli/src` is the client literal at `runtime.ts:956`); the profiler bag built by the harness in `tasks/data-profile.js:255-271` does not, and `DataProfileDeps` has no field for it. So in the CLI today every sandbox agent runs the image toolchain while reading the legacy orient text, and on Cortex the profiler does.

**Ask.** `SandboxClient` gains `readonly toolchainSource: ToolchainSource`, the value the factory was given (`"store"` when absent), and `createSandboxAgent` resolves `deps.toolchainSource ?? deps.sandboxClient.toolchainSource`. Every bag already carries `sandboxClient`. The per-bag field can stay as an override or go. The profiler needs no new field, the CLI needs no change, and Cortex drops its redundant line.

### 3. `store download` creates the mountpoint entries

**Today.** The harness binds the farm inside the read-only store bind, and runc refuses to create a mountpoint inside a read-only mount. `ensureStoreMountpoints` (`cli/src/modules/libs/composition.ts:1449-1457`) makes the empty `farm`, `current`, and `cache` entries, but it is called only from `resolveAnalysisFarm` (`:1484`), at sandbox create. `store download` does not create them, and the published artifact strips `farm` before packing (`.github/workflows/package-store-build.yml:495`). A consumer that only downloads, the platform-charts refresher Job on the `cortex-libs` PVC and Cortex's `just setup-libs`, gets a store that refuses every nested bind. Measured 2026-09-02 on runc 1.3.4: Docker refuses the run with `make mountpoint "/mnt/libs/farm": … read-only file system`; Kubernetes fails the container as `ContainerCannotRun` with empty logs (inflexa-ai/platform-charts#69).

**Ask.** Call `ensureStoreMountpoints` at the end of the download's merge in `store_download.ts`, after `mergeStagedRoot` and before the receipt, so every consumer of `store download` gets the entries. The resolver call stays as the heal for a hand-deleted entry.

### Checklist

- [ ] Image publishes `image-packages.txt` as an OCI label; harness reads it per digest
- [ ] `packageStore: "required"` turns the lock-gate degrade into a typed refusal
- [ ] `SandboxClient.toolchainSource`, read by `createSandboxAgent` as the fallback
- [ ] `store download` creates `farm`, `current`, `cache` at the store root

vivere-dally:
## What & why

Cortex, the managed embedder, met four gaps in the package store, tracked in #500. Each is a silent degradation. This PR closes all four.

- **The image inventory rides the store.** `list_available_packages` merged the tools of the image from a fragment inside the image, which a host with no engine cannot read. The image now describes itself in one JSON record, the catalog build copies it into the store root, and the harness reads it there. A `LABEL` cannot read a file that the build makes, thus the record replaces the label.
- **An embedder can require the store.** Under `packageStore: "required"` the lock gate refuses with `farm_unusable`, and an unprovable declaration refuses at composition. The client seam throws an exported `SandboxFailure` that names the failure and its cause.
- **The toolchain has one declaration.** The sandbox client carries `toolchainSource`, and the agent composition reads it there, thus each agent gets the orient text of its toolchain.
- **The download makes the mountpoint entries.** `store download` makes `farm`, `current`, and `cache` at the store root, on each run, thus runc mounts a store that a consumer only downloaded.

Closes #500.

Reviewer notes:

- The last two commits close the review findings. The conda load check fails the build when a binary has no package version.
- The image workflow moves `sandbox-base:latest` with no ref gate, thus it did not run here. A local arm64 build passed.
- Cortex edits four lines: the required-store option, the deleted `toolchainSource` line, its client fakes, and the record path from `IMAGE_PACKAGES_FILE`.
- Two older harness spec failures under the validator are not from this branch.
- Harness: 4232 tests, 0 failures. CLI: 3071 tests, 0 failures.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved (unprivileged uid-1000 workload, all capabilities dropped, `no-new-privileges`; no network egress in poll mode — the in-container firewall on Docker, a NetworkPolicy on K8s; signature-authenticated exec endpoints; read-only analysis tree with writes confined to the step's output directory; resource limits; and no host credentials in spawned commands); any loosening is called out and justified. The image gains one build arg and one world-readable record under `/opt/inflexa`. No user, capability, mount, or egress rule changes.

No analytical method changes, and no provenance record changes.


vivere-dally:
@claude review

vivere-dally:
@claude review

krajla (approved):

PR #523, opened 2026-09-08
fix(cli): name the log file by the local day, and roll it at local midnight (closes #506)

### LINKED ISSUE #506: [possible-bug]: If an inflexa session runs overnight, it may write in the wrong log file? Investigate

vivere-dally:
## What & why

The name of the log file came from the UTC date, and one destination stayed open for the whole process. Thus a session across midnight wrote into the file of the day before. The name now carries the local calendar date, and a live process moves to the file of the new day at local midnight.

Closes #506

## Reviewer notes

- The file name uses the local calendar date. `winston-daily-rotate-file` defaults to `utc: false`, and `pino-roll` has no UTC option. Each record keeps `time` in epoch milliseconds.
- The roll is the pino-roll mechanism, in-process on the `SonicBoom` of the cli. One `unref` timer fires at the next local midnight, the destination flushes and reopens, then the timer arms again. The change adds no work to the write path, no worker-thread transport, and no dependency.
- A typed cast gives the fd-built stream its `file` field, because sonic-boom refuses a reopen without it. The fd opens synchronously on purpose (`f60e1e66`), thus the exit hook never meets an unopened stream.
- An `error` listener sits on the destination, because a failed asynchronous reopen emits `error`, and an unheard `error` ends the process.
- `pino.destination` leaves `minLength` at 0, thus `flush` moves nothing here. The call stays as the correct contract of the seam, and the comment says why.
- The retention sweep reads the stamp on the local calendar. It runs at startup and at each roll, and an old UTC-stamped file still matches the pattern.
- The `--help` text of the log directory now states one file per local day, 7-day retention. The generated docs are untracked.
- No test opens a real sonic-boom destination, because a real fd in the shared test process is a hazard. A fake destination drives the roll.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). Eslint, prettier, `openspec validate --strict`, and the new test file pass. The full cli test suite did not run. The typecheck errors under `src/modules` and `src/types` are old, and they come from a stale linked harness.


vivere-dally:
@claude review

radutanasa (approved):

PR #524, opened 2026-09-08
fix(harness): report a spent retryable status as exhausted, and obey Retry-After (closes #522)

### LINKED ISSUE #522: apiFetch reports a 429 that outlives its retries as an expected absence, and it ignores Retry-After
## Summary

`apiFetch` retries a 429, but it does not read the `Retry-After` header. And a 429 that survives the last retry returns as `http_status`, which `isUnexpectedApiError` reads as an expected absence. Thus a throttled request becomes `found: false` or an empty list at some call sites. The fix is in one place. No client must change.

The review of #518 found this.

## The facts

- `src/tools/lib/api-utils.ts:37` puts 429 in `RETRYABLE_STATUSES`. Line 103 retries it with `retryDelayMs * 2 ** attempt`, thus four attempts over about 7 seconds by default.
- The sleep at line 104 ignores `Retry-After`. `src/literature/sources/http.ts:45-74` reads the header and caps it with `maxRetryDelayMs`. `src/providers/ai-sdk.ts` honors `retry-after-ms`. Thus `apiFetch` is the one HTTP layer of the harness that does not obey the server.
- On the last attempt, `attempt < maxRetries` is false at line 103. The 429 then falls to `if (!res.ok)` at line 109, and it returns `{ type: "http_status", status: 429 }`. The `exhausted` variant at line 128 is reached only through the catch path.
- `isUnexpectedApiError` at lines 155-156 reads each `http_status` in the 4xx range as expected. Its comment says that a 4xx means "the request itself was wrong". A throttle does not fit that rule.
- `src/tools/lib/api-utils.test.ts` has no test with a 429.

## Where it shows

A client that calls `isUnexpectedApiError` turns the exhausted 429 into an absence:

- `alphafold-client.ts:71` gives `null`, and the tool gives `found: false`. The tool description tells the agent not to retry a `found: false`.
- `bgee-client.ts:140`, `159`, and `171` give `null` or `[]`.
- `impc-client.ts:208`, `226`, `238`, and `249` give an empty result.
- `monarch-client.ts:93` gives an empty result.
- `chembl-client.ts:30` and `iuphar-client.ts:276` drop the log line, thus nothing records the throttle.

The clients that test `status === 404` directly throw on each other error, thus a 429 reaches the loop as an error there. Examples: `chembl-client.ts:273-274`, `pubchem-client.ts:92-93`, and `uniprot-client.ts:102-103`. These do not change.

## The proposed fix

One change in `runFetch`, and none in a client:

1. When a retryable status occurs on the last attempt, return `{ type: "exhausted", attempts, lastError: "HTTP 429" }` instead of `http_status`. `isUnexpectedApiError` already reads `exhausted` as unexpected, thus each caller above throws without a code change.
2. Read `Retry-After` in the sleep at line 104, capped by a `maxRetryDelayMs` option, the same as `http.ts`.
3. Add a 429 test to `api-utils.test.ts` with `retryDelayMs: 0`.

## Out of scope

22 call sites return `[]` or `null` on each `ApiError` with no classification at all, for example 7 in `ctx-ops.ts` and 5 in `iuphar-client.ts`. That is a different concern, and it swallows a 5xx as well. It is not part of this issue.

vivere-dally:
## What & why

A retryable status that survived the last retry of `apiFetch` returned as `http_status`, and `isUnexpectedApiError` reads each 4xx as an expected absence. Thus a throttled request became an absence at eleven call sites, and the wait ignored `Retry-After`. A spent retryable status now resolves to the `exhausted` variant, and the wait obeys `Retry-After` under a finite cap.

Closes #522

## Reviewer notes

- Under `harness/src` only `api-utils.ts` and its test changed. `isUnexpectedApiError` already returns true for `exhausted`, thus the eleven call sites that classify with it throw or log with no edit.
- The behavior at those sites changes on purpose. A throttled AlphaFold, Bgee, IMPC, or Monarch call gives an error result in place of an absence. A throttled ChEMBL or IUPHAR call writes an error log where it was silent. In the target-assessment collectors a throw maps to `coverage: "queried_no_data"`.
- One finite cap `maxRetryDelayMs`, 30 seconds by default, applies to every wait, the server value included. The `sleep` of this layer has no cancel path, and the calls run inside DBOS steps. An uncapped `Retry-After: 3600` would make a step wait for one hour.
- The header parser is a private copy of the one in `src/literature/sources/http.ts`, because that module has its own schedule, signal, and result type. The parser reads a numeric value first, because `Date.parse("-5")` gives a date in the past and thus a wait of zero.
- The new optional `sleep` on `ApiFetchOptions` is a test seam. `api-utils` is not an export of `harness/src/index.ts`, thus no embedder sees it.
- The ChEMBL requirement changes in place, because it is the one statement of the retry policy. The call sites that return an empty list on each `ApiError` with no classification stay out of scope, as the issue scoped them.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). The `api-utils` test file, the typecheck, eslint, prettier, and `openspec validate --strict` on this change pass. The full harness test suite did not run, because it needs Postgres. Two old spec validation failures, `docker-sandbox-provider` and `harness-sandbox-exec`, are not part of this branch.
- [x] Tests added or updated for the change. The cases are the spent 429, the recovered 429, the capped `Retry-After` wait, a date in the past, and a negative count.
- [x] Documentation updated if user-facing behavior changed. The specs and the doc comments of `apiFetch` and `isUnexpectedApiError` carry the new rule.


vivere-dally:
@claude review

krajla (approved):

PR #526, opened 2026-09-08
feat(harness): declare the packages of an ad-hoc step, and run the link pass on them

### LINKED ISSUE #521: Ad-hoc runs bypass the pre-launch package check: buildAdHocPlan writes a step with no packages, so the link pass has nothing to check
## Summary

An ad-hoc launch (`execute_analysis` with `mode: "adhoc"`) builds its plan with `buildAdHocPlan`, whose single step carries no `packages` field (`harness/src/tools/execute-analysis.ts:244-276`). The pre-launch link pass unions `step.packages ?? []` (`execute-analysis.ts:134`), finds nothing, and returns. So every ad-hoc run bypasses the package check that plan-mode runs get, in the CLI and in Cortex alike. In practice ad-hoc is the path the agent takes whenever the user says "run it" (`execute-analysis.ts:309`), which made it the path of every run in today's dev session.

## Proof

Harness 0.35.4, Cortex dev with a frozen-catalog realization of `extendAnalysisFarm` bound on the conversation agent (inflexa-ai/cortex#362). Two runs on 2026-09-07:

| run | plan | mode (step `track`) | steps with a `packages` key |
|-|-|-|-|
| `e430f946-…` 16:21 | `pln-50d7f0c1` "Write and run a Python script that trains a gradient-boosted classifier using…" | `ad-hoc` | 0 of 1 |
| `66075836-…` 16:25 | `pln-2ac49dcc` "Run TWO independent steps in parallel — one in R, one in Python — that both p…" | `ad-hoc` | 0 of 1 |

The stored step of `pln-2ac49dcc` has exactly the keys `buildAdHocPlan` writes: `id, name, agent, track, status, caveats, context, maxSteps, question, resources, step_type, depends_on, constraints, acceptance_criteria`. No `packages`.

That the plan-mode path could not have produced these rows, checked against the published schemas:

```
PlannerPlanSchema.safeParse(plan without packages) -> success: false
  steps.0.packages: Invalid input: expected nonoptional, received undefined
PlannerPlanSchema.safeParse(plan with packages)    -> success: true, kept: ["r:decoupleR","python:decoupler"]
after hydratePlanSteps                             -> ["r:decoupleR","python:decoupler"]
```

`PlanStepSchema` requires `packages` (`harness/src/schemas/plan-schemas.ts:24`), `submit_plan` validates with it, and `persistPlan` writes the hydrated plan verbatim. The persistence schema keeps the field optional for pre-packages history (`harness/src/schemas/workflow-state.ts:30-34`), and the ad-hoc builder relies on that optionality to write a step without it.

## Consequences

- With a seam bound, the launch refusal that names a missing or ambiguous spelling never fires for an ad-hoc request. The first run above trained with `xgboost`, a spelling that both tracks hold; had the plan carried it bare, the link pass would have refused with the prefixed remedy. It carried nothing, so the sandbox agent resolved it on its own at import time.
- The sandbox step briefing withholds `packages` on purpose because "the pre-launch link pass consumes the packages of a step" (`harness/src/prompts/briefing.ts`). For an ad-hoc step there is no such pass, so the step runs with no declared packages at all and every miss is discovered at import, mid-step, after spend.
- The CLI's own launch refusal (`run --plan`) has the same hole for its ad-hoc equivalent.

## Ask

Give ad-hoc steps a `packages` array like every other step. Two options, not exclusive:

1. **Derive it from the request.** The ad-hoc route already runs an LLM routing step to pick the agent (`AdHocRoute`); have that same step emit the packages the request names, in the query grammar, and put them on the step. The link pass then runs on ad-hoc launches unchanged.
2. **Make the absence explicit.** If ad-hoc steps are meant to skip the pass, say so in the launch log and in the step's `caveats` ("no package check: ad-hoc step declares no packages"), so an operator reading a mid-step import failure knows the check was not skipped by accident.

Option 1 is the one that closes the gap. Related: inflexa-ai/inflexa#519 (planner budget), inflexa-ai/cortex#362.

vivere-dally:
## What & why

An ad-hoc launch bypassed the pre-launch package check. The ad-hoc plan builder wrote its one step without a `packages` key, and the link pass found nothing to link. Thus the refusal that names a missing or ambiguous spelling never fired on the path that the agent takes when the user says "run it". The router now names the packages of the step, and the link pass runs on them with no change to the pass.

Closes #521

## Reviewer notes

- The router emits `packages` and validates them on their own axis. An entry that does not parse is dropped. A package failure never sets `fallbackClass`, because that class describes the whole route.
- One exported `packagesSection()` teaches the grammar to the planner and to the router. The planner prompt is byte-identical to before.
- A targeted resolution runs against the inventory of the embedder, not a census. A name that the pool does not hold is dropped with a caveat, because the list is a guess of a model. A plan step keeps the refusal, because a person approved the plan. A known spelling replaces the entry, and the list dedupes after the rewrite.
- A bare name that both tracks hold stays as written. The link pass refuses it with the two prefixed forms, the same as a plan step.
- The inventory read of `list_available_packages` is an exported function with a discriminated result, because the unavailable note of the tool keeps the pool reason. The answer of the tool is byte-identical.
- No CLI change, and no embedder change. `createExecuteAnalysisTool` is not a public export. Cortex binds the seam on the conversation agent only, thus a green CLI does not prove Cortex.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [ ] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`). Seven targeted test files, the typecheck, eslint, prettier, and `openspec validate --strict` pass. The full harness test suite did not run, because it needs Postgres. Two old spec validation failures, `docker-sandbox-provider` and `harness-sandbox-exec`, are not part of this branch.
- [x] Tests added or updated for the change. The cases cover the router packages, an unparsable entry, and the resolution with a present name, a known spelling, and an absent name. They also cover the both-track refusal, the empty list with its caveat, and the inventory read.
- [x] Documentation updated if user-facing behavior changed. The spec, the prompt section, and the briefing comment carry the new rule.


vivere-dally:
@claude review

krajla (approved):

PR #531, opened 2026-09-09
fix(harness): refuse to leak a Postgres testcontainer when no reaper …

### LINKED ISSUE #358: A local test run leaks a Postgres container for each database test file
## Goal

Stop a local run from leaking a Postgres container for each database test file.

## What happens today

A machine that runs the harness tests collects one standing
`pgvector/pgvector:pg18` container for each database test file. The containers
never end. A recent session found ten, and eight were over two hours old.

## Why it happens

Three facts multiply, and `harness/src/__tests__/setup/postgres.ts` documents
each one:

- **One container for each file, and not one for the run.** `getTestPool`
  memoizes a module-level promise (`postgres.ts:37`, `postgres.ts:97`). But Bun
  isolates the module state of each file. The header names the result: each of
  the roughly 130 files would start its own container (`postgres.ts:18-22`).
- **No teardown runs.** `stopTestContainer` exists (`postgres.ts:173`). Its doc
  states that the operation matters only when the Bun runner invokes it. The
  runner does not invoke it by default (`postgres.ts:168-172`).
- **The one reaper does not come up.** The fallback leaves the teardown to the
  ryuk sidecar (`postgres.ts:16`), and `harness/CLAUDE.md` records that ryuk does
  not come up under podman.

Thus a database test file that runs with no `CORTEX_TEST_PG_URL` leaves one
container, and nothing removes it.

## The two safe routes exist already

- `bun run test:full` starts one container and removes it on exit
  (`harness/scripts/test-full.sh:1-25`).
- An exported `CORTEX_TEST_PG_URL` points every file at one Postgres, and the
  helper then starts no container (`postgres.ts:41-44`).

The defect is that the unsafe route is the default.

## Scope

- Add a guard before a bare `bun test`. It refuses when no `CORTEX_TEST_PG_URL`
  is set and no reaper is reachable, and it names the two routes.
- Decide where the guard lives. A `pretest` script and a `bunfig.toml` preload
  are the candidates, and `cli` already runs a preload guard.
- Decide whether a unit file, which starts no container, must pass the guard.

## Out of scope

CI, which runs one `services:` container and leaks nothing. Also the ryuk
failure itself, which is upstream.

AniketR10:
## What & why

Every test file that talks to Postgres starts its own container if you
just run `bun test`. Nothing ever removes them. Ryuk (the cleanup
container from `testcontainers`) is supposed to do that, but it does
not work on podman, so containers pile up and never go away.

This PR adds a check before a container is started. It looks for a
real cleanup container running. If there isn't one, it stops and shows
you the two safe ways to run tests instead, so nothing gets left
running by accident.

Closes #358

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`) — one pre-existing, unrelated lint failure in `providers/ai-sdk.test.ts` remains; it is not touched by this PR.

Not applicable — this PR only changes test setup code (how a test
Postgres container is started), not any analysis method, the sandbox,
or provenance recording.


AniketR10:
team please review

vivere-dally:
@claude review

@AniketR10 I'll probably provide feedback / merge this tomorrow

AniketR10:
@vivere-dally sure 👍

harness/src/__tests__/setup/require-reaper.ts
```diff
@@ -0,0 +1,102 @@
...
+const ALLOW_LEAK_ENV = "CORTEX_TEST_ALLOW_LEAKED_PG";
+
+const ROUTES = `Use one of the two routes that start ONE container for the whole run:
+
+  bun run test:full [paths...]     starts one container, exports CORTEX_TEST_PG_URL, removes it on exit
+  CORTEX_TEST_PG_URL=postgres://cortex:dev@localhost:5433/cortex bun test [paths...]
+                                   points every file at a Postgres that already runs; starts nothing
+
+Set ${ALLOW_LEAK_ENV}=1 to start the container regardless, accepting one
+unreaped container for each database test file.`;
```
vivere-dally: This route does not work on a podman host, and the guard is not the reason. `assertContainerWillBeReaped` returns at line 71, then `GenericContainer.start()` calls `getReaper` itself (`generic-container.js:79`), and that call throws.

I did a test with `CORTEX_TEST_ALLOW_LEAKED_PG=1` and no other variable. The result is `Log stream ended and message "/.*Started.*/" was not received`, and the count of containers does not change.

To start the container, a developer must set `TESTCONTAINERS_RYUK_DISABLED=true` too. Name both variables in this message.

harness/CLAUDE.md
```diff
@@ -578,23 +578,30 @@ database state. Do not assert that method X was called N times with arguments Y,
...
+`bunfig.toml` preload (bun resolves bunfig from the cwd only), so a file that
+starts no container — any unit test — never meets it.
 
 There are two supported local routes. `bun run test:full` starts the one container
 itself and reaches podman through its docker-compat socket. As an alternative,
 export `CORTEX_TEST_PG_URL` at a Postgres that already runs.
-`TESTCONTAINERS_RYUK_DISABLED=true` forces the fallback through as a last resort,
-at the cost of one unreaped container for each DB test file.
+`CORTEX_TEST_ALLOW_LEAKED_PG=1` forces the fallback through as a last resort, at
+the cost of one unreaped container for each DB test file.
```
vivere-dally: This text has the same fault as the message in `require-reaper.ts`. `CORTEX_TEST_ALLOW_LEAKED_PG=1` alone does not force the fallback through on a podman host. `GenericContainer.start()` calls `getReaper` after the guard, and that call throws. Name both variables in this sentence.

harness/src/__tests__/setup/require-reaper.ts
```diff
@@ -0,0 +1,102 @@
...
+    try {
+        reaper = await getReaper(client);
+    } catch (err) {
+        throw new Error(refusalMessage(`the ryuk reaper could not be started (${describe(err)})`), { cause: err });
+    }
+
+    // A reaper with no container behind it is testcontainers' disabled stand-in.
+    if (!reaper.containerId) {
+        throw new Error(refusalMessage("the ryuk reaper is not running"));
+    }
```
vivere-dally: This branch is unreachable. Only `DisabledReaper` gives an empty `containerId` (`reaper.js:125-131`), and `DisabledReaper` comes only from `TESTCONTAINERS_RYUK_DISABLED === "true"` (`reaper.js:22-23`). Line 76 throws on that condition already.

Keep the check as a guard against a change upstream, but write that reason in the comment. The comment now names a condition that line 76 makes impossible.

harness/CLAUDE.md
```diff
@@ -578,23 +578,30 @@ database state. Do not assert that method X was called N times with arguments Y,
...
+`postgres.ts` is not shared, and the container fallback starts one container for
+each database test file rather than one for the run.
+
+Nothing but the ryuk sidecar removes those containers, and ryuk does not come up
+under podman. So the fallback refuses to start a container until it has confirmed
+a reaper is up to remove it again (`__tests__/setup/require-reaper.ts`), and the
+refusal names the routes below. The guard sits on the container-creation path
+rather than in a `pretest` script (`bun test` runs no lifecycle script) or a
+`bunfig.toml` preload (bun resolves bunfig from the cwd only), so a file that
+starts no container — any unit test — never meets it.
```
vivere-dally: These sentences break the STE rules of the root `CLAUDE.md`. Lines 587-590 make one sentence of 52 words, and lines 595-598 make one of 32. The limit is 25 words in a description. Line 593 uses `has confirmed`, but the rule says `Do not use the perfect tenses`. Line 593 also starts with `So`, and the approved connectors are `and`, `but`, `then`, `thus`, and `as a result`.

vivere-dally (changes requested):
The guard closes the leak. One documented route does not work, thus this review asks for a change.

I did a test of the guard on a podman host. `getReaper()` throws in 772 ms with `Log stream ended and message "/.*Started.*/" was not received`. `GenericContainer.start()` calls `getReaper` before it makes a container (`generic-container.js:79-82`). Thus a bare `bun test` on podman fails today, and it leaks nothing.

The leak comes from `TESTCONTAINERS_RYUK_DISABLED=true`. Three archived task files record that route, for example `harness/openspec/changes/archive/2026-07-16-harness-logger-seam/tasks.md:47`. Line 76 of `require-reaper.ts` refuses it. As a result, the change fixes issue #358.

The guard sits at the one choke point. `dbos.ts:180` reaches Postgres through `getTestPool()`, thus the guard covers the DBOS suite too. `tsconfig.json` excludes `src/**/__tests__`, thus the new module does not ship to npm. The change adds no dependency, and it adds no new privilege, because `getReaper` and the docker-socket mount of ryuk are already on this path. I found no security problem.

The inline comments name each change.

vivere-dally:
@AniketR10 one item sits outside the diff, thus it comes as a comment and not as an inline note.

`harness/openspec/specs/postgres-storage-backend/spec.md:152-154` says that each test runs against "a single `pgvector/pgvector:pg18` container reused across the run". Your change records the opposite in `harness/CLAUDE.md:587-590`: on the fallback path each database test file starts its own container. The refusal is also a new normative behavior of the test setup, and the spec does not name it.

Can you add an OpenSpec delta for `postgres-storage-backend` in this pull request? Go into `harness` first, then run the `openspec` CLI, so that the delta lands in the correct spec tree. If you prefer a follow-up change, say so, and I open it.

AniketR10:
@vivere-dally have made the requested changes and openspec delta as well, please review

vivere-dally (approved):

PR #538, opened 2026-09-11
feat(harness): report stale DBOS workflows and their recovery attempts

krajla:
## What & why

On 2026-08-23 Cortex entered an OOM loop at boot. DBOS recovered the same workflow at each boot, and its `recovery_attempts` increased toward the SDK limit of 100. No signal showed this. `bootHarness` now registers two gauges after the DBOS launch (checklist B8, signal O8, alert A6, decision D-14):

- `cortex.dbos.workflows.stale{status}`: the `PENDING` and `ENQUEUED` workflows that are older than 6 h.
- `cortex.dbos.workflows.recovery_attempts.max`: the highest `recovery_attempts` of those workflows, of all ages. A workflow that ran one time reads 1.

One query feeds the two gauges at each export (60 s):

```sql
SET LOCAL statement_timeout = 2000;
SELECT status,
       count(*) FILTER (WHERE created_at < (extract(epoch FROM now()) * 1000)::bigint - 21600000) AS stale,
       max(recovery_attempts) AS max_recovery_attempts
  FROM dbos.workflow_status
 WHERE status IN ('PENDING', 'ENQUEUED')
 GROUP BY status
```

Notes for the reviewer:

- **Cost.** DBOS 4.23.6 has the partial index `idx_workflow_status_in_flight` on these two statuses. The plan scans that index, thus the cost changes with the number of in-flight rows, not with the size of the table (1.6 ms on a local DBOS schema). The `SET LOCAL` ends with the message, thus the pooled connection keeps its setting. A client limit of 5 s includes the wait for a connection. A failed query logs a warning.
- **6 h.** The parent of an analysis run stays `PENDING` for the full run, and `cortex.run.duration` has its last finite bucket at 4 h. With the 30 m of A6, each run longer than 30 min shows as stale. The recovery gauge finds a crash loop in minutes.
- **Labels.** `status` only, never an id. The cumulative export sends the last value again for a label set that a collection does not observe. The gauge cannot observe each workflow name at zero. Thus, with a name label, a count stays high after the workflow ends.
- **Alert.** Each replica reads the same table. Use `max`, never `sum`: `max(cortex_dbos_workflows_recovery_attempts_max) > 3`, `max(cortex_dbos_workflows_stale) > 0`.
- The gauge sends no query when metric export is off, in the CLI, or while DBOS is not launched.
- **Tests:** the row mapping and the return to zero, a failed query, and no query.

The version does not change. Cortex gets the gauges when it pins the next harness release.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

https://claude.ai/code/session_01G47UrASYb15J6JmG3npn1b

### DIFF
--- a/harness/src/lib/otel.ts
+++ b/harness/src/lib/otel.ts
@@ -18,8 +18,10 @@
  *          outcome instruments (`cortex.run.*`, `cortex.step.*`) and the
  *          reconcile counters are defined in metrics.ts; the agent loop, the
  *          thread memory, and the workflows hold their own instruments next
- *          to their record sites. Every instrument binds lazily to the
- *          provider registered here.
+ *          to their record sites, and `runtime/dbos.ts` observes the DBOS
+ *          workflow table at each export. Every instrument binds to the
+ *          provider registered here: a record site at its first use, the
+ *          DBOS gauges at boot, after this init.
  *
  * Resource: the host's `serviceName` / `serviceVersion` merged with the SDK
  *           env detector (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`);
--- a/harness/src/runtime/boot.ts
+++ b/harness/src/runtime/boot.ts
@@ -25,6 +25,9 @@
  *      legacy-workflow reap, an agent-switch install). Runs after registration
  *      so it may close over the registered callables.
  *   7. `launchDbos(...)` — the last registration-dependent step.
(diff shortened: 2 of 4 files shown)


krajla:
Closing: metrics are not read from the database at runtime. Stuck or repeatedly recovered workflows come from what DBOS emits itself, or from instruments recorded where the harness acts.

PR #543, opened 2026-09-12
chore(images): add tximport to the package-store manifest

radutanasa:
## What & why

The package-store manifest gets one entry, `tximport`, beside the count-model packages. The knowledge templates import Salmon and kallisto quantifications with tximport, and the catalog farm does not hold the package. Each quantification test of the templates stops at the library call. With the entry, the store build resolves tximport into the two lock files, and the import path can run in the sandbox.

Notes for the reviewer:

- The entry comes from the knowledge plane branch, where it sat since the Salmon path was completed. The locks never saw it, because no store build ran from that branch.
- The manifest is the intent layer. The store pipeline resolves it and commits the lock files back, thus this pull request changes no lock file.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

### DIFF
--- a/images/package-store/manifest.yaml
+++ b/images/package-store/manifest.yaml
@@ -225,6 +225,11 @@ r:
         - limma
         - sva
         - variancePartition
+        - name: tximport
+          reason: >
+            The import path of tpl-deseq2-two-group 1.1.0. It summarizes the
+            Salmon and kallisto transcript quantifications to the gene level,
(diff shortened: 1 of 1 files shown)


vivere-dally (approved):
Can you maybe do a wider audit and try to do a bulk add of things? It takes ~3 hours for the builders to finish the package pipeline (maybe that can be improved in the future).
But yeah, it would be best to build it once and have it for a longer time

PR #511, opened 2026-09-04
The package identity: one rule for each track, and one query grammar

vivere-dally:
## What & why

A launch refused a plan because the pool holds the Python distribution `decoupler` and the R package `decoupleR`. The store build folded each R name with the PEP 503 rule, thus the two packages shared one lookup key. A plan could not name an ecosystem, thus the refusal had no remedy, and the agent dropped the method. A trace of the branch found the deeper cause: a package name crossed three subsystems as a bare string, and each site applied its own rule.

This pull request gives the harness one module, `package-identity`. The module holds a `PackageQuery`, an opaque `PackageIdentity`, one grammar `[python:|r:]<spelling>[==<version>]` with one parser, and one resolution ladder over a pool index. The cli and the provisioner consume the module.

Notes for the reviewer:

- Graph version 2 is redefined in place, because no release carries it. A node holds no `r_dir`, and the fold gate is gone. A development store from this branch rebuilds.
- The fold exists in one TypeScript function and in one Python function. One fixture, `harness/src/sandbox/__fixtures__/package-identity.json`, binds the two.
- Three user-visible changes. `store link seurat` refuses and names `Seurat`. The census answers a name the same way as the link. `store add scanpy==1.11` records the version.
- Migration 10 rebuilds the two request tables. The column `spelling` replaces `name` and `raw_name`, and each flight id comes from the spelling.
- `link_packages` takes the one grammar, and it loses its `ecosystem` field. The link pass absorbs nothing.
- The design decisions and their rejected alternatives are in the archived changes `2026-09-04-package-name-resolution` and `2026-09-04-package-identity` of the harness and of the cli.
- The harness suites that use Postgres did not run on the review machine. The touched suites, the typecheck, and the provisioner suite ran green.

## Type of change

- [ ] Bug fix
- [x] New feature / capability
- [ ] Documentation
- [x] Refactor / internal change
- [ ] Analytical method, sandbox, or provenance change (gets extra scrutiny — see below)

## Checklist

- [x] Lint, typecheck, and tests pass locally (`bun run lint`, `bun run typecheck`, `bun test`).
- [x] Tests added or updated for the change.
- [x] Documentation updated if user-facing behavior changed.
- [x] Commits follow [Conventional Commits](https://www.conventionalcommits.org/).
- [x] Commits are **signed off** for the DCO (`git commit -s`).
- [x] This PR is focused on a single logical change.

PR #548, opened 2026-09-15
feat(harness): resolve the packages of a plan against the pool at sub…

### LINKED ISSUE #512: Resolve the packages of a plan against the pool at submit time, and count the image packages as present
## Why

`submit_plan` validates each package entry with a parse only (`harness/src/schemas/validate-plan.ts:181-184`). A bare name that both tracks hold, or a name that the pool does not hold, passes and refuses at launch (`harness/src/tools/execute-analysis.ts:125-189`). The planner never sees that refusal in its loop, but it holds the census that the link pass uses (`harness/src/tools/research/generate-plan.ts:1081-1083`).

A base R package such as `stats` or `grid`, and a Python standard-library module such as `json`, are not in the pool. The ladder answers `unknown` (`harness/src/sandbox/package-identity.ts:316`), and the launch refuses a plan that names one.

An experiment of 2026-09-05 measured this on `claude-haiku-4-5`. The comment below holds the numbers. Half of the accepted plans refuse at launch. With the resolution inside the validator, 97 to 100% end clean.

## What to change

1. Give `validatePlan` an optional `PoolIndex`. Resolve each entry with `resolveQuery`. Refuse `ambiguous` with the two prefixed forms. Refuse `unknown` with the suggestion of the ladder. Use the words of `absentOutcome`, `cli/src/modules/libs/composition.ts:1587`.
2. Build the index in the planner from the census it reads already, and pass it to `fullyValidate` (`generate-plan.ts:452`). The link pass stays.
3. Count a base R package and a Python standard-library module as present, in the validator and in the link pass. The image record `/opt/inflexa/image-packages.json` is the truthful source. `sys.stdlib_module_names` gives the Python set.
4. Keep the string grammar. No tolerated list, no object shape.
5. Modify `planning-enhancements` and `package-identity` in `harness/openspec/specs`. Add tests for the two refusals and for a base package that passes.

## Acceptance criteria

- A bare both-track name is refused at `submit_plan`, and the message names the two prefixed forms.
- `sklearn` is refused at `submit_plan`, and `r:stats` or `json` is accepted and links.
- The link pass and the validator give one answer for each entry.

AniketR10:
## What & why

Pasted image (3)
Pasted image (2)

The `submit_plan` tool now resolves each package entry against the pool. Before, a bare name that both tracks hold, or a name that the pool does not hold, passed the submit. Then the plan failed at launch, after the planner loop had stopped. Now the planner sees the refusal in its loop and can correct the plan.

The image record now lists the base R packages (`r_base`) and the Python standard-library modules (`python_stdlib`). The validator and the CLI link pass count these names as present. Thus `r:stats` and `json` are accepted and link.

Notes for the reviewer:

- The change touches the harness, the CLI, and the image record script. The CLI part needs the harness release and the pin update first.
- A store that was packed before this change has no base-name fields. On such a store, `r:stats` is refused at the submit until the store is downloaded again.
- The `names` lookup of `list_available_packages` is not changed. It still reports `stats` as absent.
- Two CLI tests fail locally: a usage e2e timeout and a SIGKILL timing test. They fail the same way without this change.

Closes #512

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [ ] **Methods:** Not applicable. No analytical method changed.
- [x] **Sandbox:** The security model does not change. The image record script reads two lists of names from the R and Python runtimes during the build. It adds no privilege, no network access, and no write path.
- [ ] **Provenance:** Not applicable. No provenance record changed.


AniketR10:
team please review

vivere-dally:
@claude review

AniketR10:
team please review

cli/src/modules/libs/composition.ts
```diff
@@ -1625,12 +1661,88 @@ function absentOutcome(query: PackageQuery, failure: RequestResolutionError): Pa
...
+    const pool = poolIndexOf(graph);
+    const graphAnswer = (resolution: QueryResolution): LinkResolution =>
+        answerResolution(graph, query, resolution).match(
+            (answer): LinkResolution => ({ kind: "pool", answer }),
+            (failure): LinkResolution => ({ kind: "refused", failure }),
+        );
+    if (record === undefined) return graphAnswer(resolveQuery(query, pool));
+
+    const image = imagePoolIndex(record);
+    const resolution = resolveQuery(query, joinPoolIndexes(pool, image));
```
vivere-dally: **Must change.** The join makes a bare `optparse` refuse.

The manifest of the package store holds the R package `optparse` (`images/package-store/manifest.yaml:59`). Python 3.12 lists `optparse` in `sys.stdlib_module_names`. Thus the joined index holds `r:optparse` and `python:optparse`.

I ran `resolveQuery` over the two indexes:

- over the graph index alone: `resolved`, with the identity `r:optparse`
- over the joined index: `ambiguous`

Thus a step that names a bare `optparse` passes today, and `submit_plan` refuses it after this change. The link pass refuses it too. `optparse` is the one such name in the manifest of today. A new Python version, or a new R package, can add another one.

The cause is the rank of the two indexes. The ambiguity arm treats the image as a peer of the pool.

The remedy is two steps, in one harness function that both readers call:

1. Resolve the query over the pool index. If the answer is `resolved`, take it.
2. Otherwise resolve the query over the joined index.

The design document rejects a lookup of the image BEFORE the ladder. This rule is the opposite order, thus that objection does not apply. The two readers stay in agreement, because one function holds the rule.

The rule against the acceptance criteria of #512:

- `r:stats` — step 2 resolves. Accepted.
- `json` — step 2 resolves. Accepted.
- `optparse` — step 1 resolves as `r:optparse`. No refusal.
- `igraph` — ambiguous at both steps. Refused.
- `sklearn` — unknown at both steps. Refused.

Step 2 also states which source answered. Thus line 1707 needs no `!pool.has(...)` guard.

cli/src/modules/libs/composition.ts
```diff
@@ -1625,12 +1661,88 @@ function absentOutcome(query: PackageQuery, failure: RequestResolutionError): Pa
...
+    | { readonly kind: "image"; readonly version: string }
+    | { readonly kind: "refused"; readonly failure: RequestResolutionError };
+
+/**
+ * The valid image record at the root of the store, or `undefined`. A store
+ * from before the record, and a record that does not parse, both give nothing:
+ * the base sets are an enrichment, and the graph still answers without them.
+ */
+function readStoreImageRecord(storeRoot: string): ImagePackages | undefined {
+    return readImagePackagesFile(join(storeRoot, IMAGE_PACKAGES_FILE)).unwrapOr(undefined);
```
vivere-dally: **Must change.** `unwrapOr(undefined)` discards the one distinction that matters.

`readImagePackagesFile` gives two errors (`harness/src/sandbox/image-packages.ts:97`):

- `record_unreadable` — a store from before the record. A normal state.
- `record_invalid` — a damaged record. A structural fault.

This function maps both to `undefined`. Thus a damaged record answers `absent` with `acquisitionPossible: true` for `r:stats`, and no message names the damage. The agent then asks for an acquisition of a base R package, and no repository holds one.

`linkPackagesIntoFarm` already holds the correct rule for the sibling read: an unreadable graph answers `unavailable` with the reason (line 1756). The comment there states why a false absence is worse than a refusal.

Keep the `Result`, and split the two errors:

- `record_unreadable` gives an index that holds nothing.
- `record_invalid` gives `unavailable` for each query, with the reason.

The state `undefined` then leaves the signatures. `imagePoolIndex` answers nothing for a record with neither field, thus an absent record is an empty `PoolIndex` and not a second state. Line 1703 loses its branch, the tail at line 1735 covers both paths, and `claimOf` loses two of its four parameters.

One more point for the same change. `resolveLinkRequest` rebuilds both indexes for each query (lines 1697 and 1705), and `imagePoolIndex` makes a map of about 314 entries at each call. The batch reads the graph one time and the record one time (lines 1756 and 1767). Build the two indexes there.

cli/src/modules/libs/composition.ts
```diff
@@ -1625,12 +1661,88 @@ function absentOutcome(query: PackageQuery, failure: RequestResolutionError): Pa
...
+        return { kind: "image", version };
+    }
+    if (resolution.kind === "ambiguous") {
+        // One side of the ambiguity can be a base package of the image, and its
+        // track holds no store directory. `answerResolution` reads the two
+        // shelves of the graph alone, thus it would report the pair as unknown
+        // and lose the one remedy that a caller can act on: name the track.
+        const python = claimOf(graph, record, image, resolution.python);
+        const r = claimOf(graph, record, image, resolution.r);
+        if (python !== undefined && r !== undefined) {
```
vivere-dally: **Must change.** This guard cannot be false, and its fallback restores the defect that `90448b90` fixed.

- An `ambiguous` answer states that the joined index holds the two identities (`harness/src/sandbox/package-identity.ts:352`).
- The joined `has` is the disjunction of the two indexes (`harness/src/sandbox/package-identity.ts:287`).
- `claimOf` answers the head of the graph, or the runtime of the image (line 1666).

Thus each claim is defined, and the guard always passes.

If a later change of `poolIndexOf` or of `claimOf` makes it false, the code falls to `graphAnswer` at line 1735. That call answers `unknown_distribution` for the pair, which is the exact defect that this commit fixed. The fallback is silent.

Make the claim total. `claimOf` returns a string, and it throws for an identity that neither source holds. This file already throws for an unreachable state at line 1658. The caller then needs no guard.

harness/src/schemas/validate-plan.ts
```diff
@@ -176,11 +222,19 @@ export function validatePlan(plan: AnalysisPlan, options?: ValidatePlanOptions):
...
     for (const step of plan.steps) {
         for (const entry of step.packages ?? []) {
             const parsed = parseQuery(entry);
-            if (parsed.isErr()) errors.push(describeParseError(step.id, entry, parsed.error));
+            if (parsed.isErr()) {
+                errors.push(describeParseError(step.id, entry, parsed.error));
+                continue;
+            }
+            if (!pool) continue;
+            const issue = describeResolution(step.id, entry, parsed.value, resolveQuery(parsed.value, pool));
```
vivere-dally: **Must change.** A pinned base package passes the submit, and it refuses at the launch.

The version takes no part here, thus `r:stats==3.0.0` passes. The link pass refuses it (`cli/src/modules/libs/composition.ts:1712`). That is the class of defect that #512 names.

The design document names the version as a non-goal, because the census holds only the newest pin of each package. That reason is correct for a package of the pool. It is not correct for a package of the image: the record holds the exact version of each runtime, and `imagePoolIndex` reads that same record.

When only the image resolves the identity, compare `query.version` with `record.runtimes[track]` here. The two-step rule of my comment on `composition.ts:1706` already states which source answered.

cli/src/modules/libs/composition.ts
```diff
@@ -1652,28 +1764,27 @@ export async function linkPackagesIntoFarm(storeRoot: string, analysisId: string
...
 
-    return resolutions.map(({ query, resolved }) =>
-        resolved.match(
-            (answer): PackageRequestOutcome => {
+    return resolutions.map(({ query, resolved }): PackageRequestOutcome => {
+        switch (resolved.kind) {
+            case "image":
+                // The image holds the package, thus nothing links, and the
+                // extension of the batch cannot refuse it.
+                return { kind: "present", spelling: query.spelling, version: resolved.version };
```
vivere-dally, line 1771: **Must change.** The harness defines `present` as "the farm linked it already" (`harness/src/sandbox/types.ts:87`). This line answers `present` for a package that the farm never links.

The harness owns the contract of the seam. Thus the doc comment of `PackageRequestOutcome` and the harness spec must carry the second meaning. The root `CLAUDE.md` gives the rule: design the capability in the harness first, then connect it from the embedder.

cli/src/modules/libs/composition.ts
```diff
@@ -1625,12 +1661,88 @@ function absentOutcome(query: PackageQuery, failure: RequestResolutionError): Pa
     }
 }
 
+/** What claims one track of a spelling: a store directory of the pool, or the runtime of the image. */
+function claimOf(graph: DepsGraph, record: ImagePackages, image: PoolIndex, identity: PackageIdentity): string | undefined {
+    const head = graph.byName[identity.track].get(identity.name)?.[0];
+    if (head !== undefined) return head;
+    return image.has(identity) ? `the ${identity.track} runtime of the image (${record.runtimes[identity.track]})` : undefined;
```
vivere-dally: **Must change.** `storeDirs` is a pair of store directories (`harness/src/sandbox/types.ts:110`). This line puts a phrase of the runtime in that pair. `harness/src/tools/execute-analysis.ts:190` then renders this text:

```
optparse — two store directories claim it, the python runtime of the image (3.12.3) and r-optparse-1.7.5
```

Widen the definition in the harness, and change the sentence at `execute-analysis.ts:190`, thus it promises no two directories.

vivere-dally (commented):
I read the change against #512, against its parent #511, and against the code. The pull request fixes the issue, and the root cause is correct: one package entry had three readers, and only the link pass asked the pool.

The bugbot comment is answered. `90448b90` refuses a wrong pin, and it builds the refusal of a two-source ambiguity. Each of the two carries a test. My comment on `composition.ts:1724` is a remainder of that fix.

Six items must change before the merge, and five nits are optional. Each one is an inline comment.

Against the head `97b2074d`: the four touched harness suites pass (106 tests), `composition.test.ts` of the cli passes (69 tests), and the typecheck and the lint of both subsystems are clean under `harness:local`.

One finding is new. A comparison of `images/package-store/manifest.yaml` against `sys.stdlib_module_names` of Python 3.12 gives one collision: `optparse`. A bare `optparse` resolves today, and it refuses after this change. The comment on `composition.ts:1706` holds the detail and the remedy.

Two notes for the merge:

- The change is inert until three builds run. The image build writes the two fields, the store build copies the record (`.github/workflows/package-store-build.yml:491`), and the user downloads the store again.
- The harness gains new public exports, and the cli pin stays at `^0.35.7`. `version-guard` blocks a version change from a non-maintainer, thus the maintainer must publish the harness and bump the pin before the next cli release.

I also read the change for a security fault, and I found none. No workflow, no dependency, and no lockfile changed. An `image` outcome contributes no farm root, thus the change adds no mount and no privilege.

cli/openspec/changes/resolve-plan-packages-at-submit/specs/farm-composition/spec.md
```diff
@@ -0,0 +1,53 @@
+## ADDED Requirements
+
+### Requirement: The seam route counts an image package as present
```
vivere-dally: **Must change.** This delta must modify, and not add.

`cli/openspec/specs/farm-composition/spec.md:37` holds the requirement "Extension walks the graph and refuses ambiguity". It states two things that this change contradicts:

- "A farm MUST extend only through the graph"
- "The host supplies a `PoolIndex` over the graph"

Two requirements of one spec tree then disagree. `openspec validate --strict` reads the structure only, thus it stays green.

Write the delta as `## MODIFIED Requirements` for "Extension walks the graph and refuses ambiguity", with the full text of the new requirement.

harness/src/tools/sandbox/list-available-packages.ts
```diff
@@ -481,7 +504,14 @@ export type ListAvailablePackagesDeps = Pick<EnvironmentStorePaths, "farmLockFil
  * distinguishable from an inventory that holds nothing, and a caller that
  * cannot tell them apart reports a present package as absent.
  */
-export type InventoryRead = { readonly kind: "sections"; readonly sections: readonly Section[] } | { readonly kind: "unavailable"; readonly reason?: string };
+export type InventoryRead =
+    | {
+          readonly kind: "sections";
+          readonly sections: readonly Section[];
+          /** The valid image record that merged into the sections, where the host could read one. */
+          readonly record?: ImagePackages;
```
vivere-dally: Nit. The record rides through the public read shape so that one caller can rebuild an index. A read that carries its own `PoolIndex` needs no new public field. It also prevents a call of `inventoryPoolIndex` that forgets the record, and such a call drops the base sets with no signal.

harness/src/tools/sandbox/list-available-packages.ts
```diff
@@ -522,7 +552,30 @@ export async function readInventorySections(deps: ListAvailablePackagesDeps): Pr
...
+ * conversation surface has no runtime to probe.
+ *
+ * The tool answers through here, and so does the planner, which reads the
+ * inventory one time for its seed and for its submit index.
+ *
+ * @param read One inventory read.
+ * @param input The query of the call.
+ * @param poolScope Whether the read came from the pool-scope reader.
+ */
+export function answerPackagesQuery(read: InventoryRead, input: PackagesQuery, poolScope: boolean): PackagesResult {
```
vivere-dally: Nit. `poolScope` is a positional boolean, and each of the two call sites computes `readPoolInventory !== undefined` again. The read can carry its own scope, thus no caller can pass the wrong value.

harness/src/tools/sandbox/list-available-packages.ts
```diff
@@ -340,6 +341,28 @@ function poolIndexOf(held: readonly Held[]): PoolIndex {
...
+ * tool, a node package) is not a package of a track, and it resolves nothing.
+ *
+ * @param sections The sections of one inventory read.
+ * @param record The valid image record of that read, where there was one.
+ */
+export function inventoryPoolIndex(sections: readonly Section[], record?: ImagePackages): PoolIndex {
+    const held: Held[] = [];
+    for (const section of sections) {
+        if (section.track === undefined) continue;
+        for (const pkg of section.packages) held.push({ entry: pkg, section: section.title, identity: identityOf(section.track, pkg.name) });
```
vivere-dally: Nit. This builds a `Held` row with `entry` and `section`, and `poolIndexOf` reads neither (line 327). The `names` path needs those two fields, and this path does not. Give `poolIndexOf` the identities.

harness/src/tools/research/generate-plan.ts
```diff
@@ -1070,17 +1077,33 @@ export function createGeneratePlanTool(deps: GeneratePlanDeps): Tool {
...
+                    ...(deps.readPoolInventory === undefined ? {} : { readPoolInventory: deps.readPoolInventory }),
+                }),
+            ]);
             const refsBlock = inventoryContent("Available Reference Data", refsResult, logger);
-            const packagesBlock = inventoryContent("Available Packages", packagesResult, logger);
+            // A listing with no `names` always answers with rendered content, and
+            // an unreadable inventory answers with its UNKNOWN note. The read never
+            // fails as a call, thus the block carries no lookup-failed arm.
+            const packagesAnswer = answerPackagesQuery(packagesRead, {}, deps.readPoolInventory !== undefined);
+            const packagesBlock = `## Available Packages\n\n${"content" in packagesAnswer ? packagesAnswer.content : JSON.stringify(packagesAnswer)}`;
```
vivere-dally: Nit. This writes the heading by hand, and line 1091 renders the sibling block with `inventoryContent`. Two renderers serve one seed. `inventoryContent` can take the answer, or this block can route through it.

images/sandbox-base/scripts/image-record.py
```diff
@@ -49,6 +53,36 @@ def runtime_version(command, prefix):
...
+    return names
+
+
+def python_stdlib_modules():
+    """Return the names of the standard-library modules of this interpreter, sorted.
+
+    The script runs under the system python3 of the runtime stage, which is the
+    interpreter that a sandbox runs. Thus the set is the set of the image.
+    """
+    names = sorted(sys.stdlib_module_names)
```
vivere-dally: Nit. `sys.stdlib_module_names` holds 88 private names of 300 on Python 3.12, such as `_abc` and `_ast`. The fold turns `_abc` into the address `-abc`. No plan can name one of them. A filter on the leading underscore keeps the record smaller.

AniketR10:
@vivere-dally have addresssed all the comments please review

AniketR10:
@vivere-dally lemme know if any modifications needed in this PR or if it is good to merge, thanks!

vivere-dally:
Hi @AniketR10, I looked at the PR, and the agents did the reviews. But I'm not on board with the updates made to the sandbox, and that we're reporting the stdlibs of Python and R. I don't like how the fix turned out, and it requires quite a lot of time to test it, and create a new sandbox build, and I didn't have the time the past few days.
If you want to work on it a bit more, you're welcome to do so. Make sure on your side that every change is warranted and that there is nothing extra. Otherwise, we will leave the PR open for a while, until I get some time to seriously look over it and test it

PR #562, opened 2026-09-20
fix(cli): make the compose stack work on a Windows host

### LINKED ISSUE #560: Windows: inflexa setup generates docker-compose.yml with unparseable double-quoted paths
**Environment**
- OS: Windows 11
- Compose provider: podman compose (external provider: `docker-compose.exe`)

**What happened**

Running `inflexa` (or `inflexa setup`) fails while pulling container images:

```
go-yaml load error in scanner (while scanning a quoted scalar) at L14.C9-C14: did not find expected hexadecimal number
Error: executing C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\docker-compose.exe -f C:\Users\<user>\AppData\Local\inflexa\docker-compose.yml pull: exit status 1
```

**Root cause**

The generated `docker-compose.yml` wraps Windows paths in double quotes, e.g.:

```yaml
volumes:
  - "C:\Users\<user>\AppData\Local\inflexa\cliproxy\config.yaml:/CLIProxyAPI/config.yaml"
  - "C:\Users\<user>\AppData\Local\inflexa\cliproxy\auth:/root/.cli-proxy-api"
```

In YAML, double-quoted scalars treat backslash as an escape character. `\U` (from `\Users`) is parsed as the start of a `\U########` Unicode escape sequence, which fails to parse since the following characters aren't hex digits — hence "did not find expected hexadecimal number".

This will affect every Windows user, since `%LOCALAPPDATA%\inflexa\...` paths always contain backslashes.

**Workaround**

Manually editing the generated file to use single-quoted YAML strings instead (single quotes are literal, no escape processing) fixes the immediate pull:

```yaml
volumes:
  - 'C:\Users\<user>\AppData\Local\inflexa\cliproxy\config.yaml:/CLIProxyAPI/config.yaml'
  - 'C:\Users\<user>\AppData\Local\inflexa\cliproxy\auth:/root/.cli-proxy-api'
```

This is only a local workaround — the file header states it is regenerated on every `inflexa setup` run, so the fix does not survive a re-run.

**Suggested fix**

In whatever template/serializer generates `docker-compose.yml`, either:
- single-quote path strings instead of double-quoting them, or
- escape backslashes (`\`) when double-quoting, or
- normalize Windows paths to forward slashes before writing them into the compose file.

**Expected behavior**

`inflexa setup` / `inflexa` should generate a valid `docker-compose.yml` on Windows and successfully pull images without manual editing.

### LINKED ISSUE #561: Windows: inflexa-postgres crash-loops (Windows bind mount can't chmod), inflexa hangs with no error
**Environment**
- OS: Windows 11
- Compose provider: podman compose (external provider: `docker-compose.exe`)
- Podman version: 5.8.5

**What happened**

After working around the YAML quoting bug (#560), `inflexa` starts containers successfully, but `inflexa-postgres` immediately crash-loops. `podman ps -a` shows it repeatedly exiting with code 1. From the outside, since `inflexa` doesn't surface this — it just waits on a container that will never become healthy — the whole `inflexa` command appears to hang indefinitely with no error output.

**Root cause**

`docker-compose.yml` bind-mounts a Windows host directory directly into the Postgres container for its data directory:

```yaml
volumes:
  - "C:\Users\<user>\AppData\Local\inflexa\postgres:/var/lib/postgresql"
```

`podman logs inflexa-postgres` shows:

```
initdb: error: could not change permissions of directory "/var/lib/postgresql/18/docker": Operation not permitted
fixing permissions on existing directory /var/lib/postgresql/18/docker ... chmod: changing permissions of '/var/lib/postgresql/18/docker': Operation not permitted
```

Podman on Windows shares Windows host directories into its Linux VM via a 9p/virtiofs-style file-sharing layer that does not support `chmod`/`chown` on those paths. Postgres's `initdb` requires the data directory to have specific ownership/permission bits (0700, owned by the `postgres` user), which it cannot set on a Windows-backed bind mount, so `initdb` fails every time the container starts, `restart: unless-stopped` keeps retrying, and the container never becomes healthy.

**Workaround**

Using a Podman-managed named volume instead of a Windows bind mount for Postgres's data directory avoids the cross-OS permission-translation issue entirely (data lives inside the Linux VM's own filesystem, which is also the standard way to run Postgres-in-a-container on Windows):

```yaml
services:
  inflexa-postgres:
    volumes:
      - inflexa-postgres-data:/var/lib/postgresql
    ...

volumes:
  inflexa-postgres-data:
```

With this change, `inflexa-postgres` starts and stays up normally.

This is a local workaround only — the file is regenerated (with the Windows bind mount) on every `inflexa`/`inflexa setup`/`inflexa up` invocation, so it doesn't survive a re-run.

**Suggested fix**

On Windows specifically, generate the Postgres volume entry as a named/managed volume rather than a host bind mount to the Windows filesystem. (The CLIProxyAPI config/auth bind mounts appear fine since that container doesn't need to chmod/chown its mounted paths.)

**Expected behavior**

`inflexa-postgres` should start and stay healthy on Windows without manual intervention, and if it does crash-loop, `inflexa` should surface that failure clearly instead of appearing to hang indefinitely.

**Related**

Likely worth fixing alongside #560, since both are in the same generated `docker-compose.yml` and both are Windows-specific.

radutanasa:
## What & why

On a Windows host, `inflexa setup` and `inflexa` failed for two reasons. The compose template wrote each host path raw into a double-quoted YAML scalar, thus the parser read `\U` in `C:\Users` as a Unicode escape. And the Postgres service bind-mounted a host directory, but the file share of a Windows host refuses `chmod`. Thus `initdb` failed at each start, and the container restarted again and again. The readiness wait reported only a timeout.

Now the template escapes each host path. A Windows host keeps the Postgres data in a named volume, and each other host keeps its bind mount. The readiness wait reads the restart count of the container. The restart-loop error and the timeout error carry the last lines of the log.

Notes for the reviewer:

- **Nothing ran on a Windows host or on Podman.** Unit tests cover each Windows branch through the new `ComposeHost` parameter.
- **The key is the platform, not the engine.** Thus Docker on Windows gets the named volume too.
- **`down --delete-data` calls `volume rm` on each platform.** An absent volume is `ok`. `compose down -v` reads a compose file that `inflexa down` never writes.
- **The wait uses a rise of 2 in `RestartCount`, not the status.** A past restart of a healthy container must not give the error.
- **One difference from the design.** The code calls `container inspect`. Design D6 shows plain `inspect`.
- **One open question is in the design.** `capture` has no time limit for one call, and this change adds none.
- **One test fails on the development machine, outside this change.** `local-provider.test.ts` expects `err` from a loopback embed behind a poisoned proxy, and it gets `ok`. The test fails alone too. No file of the embedding module differs from `main`. The other 3124 tests pass, thus the first box stays open.

Closes #560
Closes #561

## Type of change

- [x] Bug fix
- [ ] Documentation

## Checklist


radutanasa:
@claude review

cli/src/modules/infra/postgres_types.ts
```diff
@@ -71,6 +71,10 @@ export type PostgresError =
     | { type: "container_stop_failed"; message: string }
     | { type: "compose_file_write_failed"; message: string }
     | { type: "ready_timeout"; message: string }
+    // The engine restarted the Postgres container again and again while the readiness wait ran, so the
+    // wait stopped before its timeout. The message carries the last lines of the container log, which
+    // hold the real cause (for example an `initdb` failure), so consumers render it verbatim.
```
vivere-dally: No need for the comment. The LLM added it because it had more context. The other branches are just as important and are not commented

vivere-dally (dismissed):

krajla (approved):

PR #566, opened 2026-09-21
feat(harness)!: replace the billing seams with host hooks and a generic suspension

krajla:
## What & why

The harness read the values of its billing seam, and two spawn paths stamped no attribution labels. A workflow paused
only when the text of an error matched `budget exceeded`. Now a request headers hook replaces the billing seam, and
each host hook is a gate or a notice with its kind in its type. The sandbox client stamps the host labels at each
spawn, on the Docker backend and on the K8s backend. A suspension starts only from a status in the suspend map of a
provider, or from a gate refusal with the suspend flag. A `Result` also survives a DBOS checkpoint, thus a step can
give a failure as a value.

## Notes for the reviewer

- This change breaks the API of each embedder. `ResolveBilling` and `createNoopBillingResolver` are gone, each seam
  method gives a `ResultAsync`, and `createSandbox` takes the session. Cortex must adopt the hooks in its own pull
  request.
- The CLI changes in this pull request, because the `cli` jobs of CI link the working-copy harness.
- The DB and DBOS suites did not run locally, because Postgres was not available.
- The pull request also archives `drop-analysis-billing-context`, makes two older sandbox specs valid, and rebuilds
  the stale delta of `fix-lineage-attestation-severity`.
- Some hook calls in `data-profile`, `extract-values`, and `sandbox-step` stay inside a `try` of the body. The spec
  forbids such a `try` only in `executeAnalysis`.
- The model calls of the step summary and the file metadata still discard a `suspend` error, as before this change.

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Checklist

- [x] **Sandbox:** the security model is preserved (unprivileged uid-1000 workload, all capabilities dropped, `no-new-privileges`, no network egress in poll mode — the in-container firewall on Docker, a NetworkPolicy on K8s, signature-authenticated exec endpoints, read-only analysis tree with writes confined to the step's output directory, resource limits, and no host credentials in spawned commands). Any loosening is called out and justified.
- [x] **Provenance:** prior analyses remain reproducible (or the expected variance is documented).


vivere-dally:
@claude review

vivere-dally (dismissed):

vivere-dally (approved):

